PR #20, opened 2020-04-08
Added Kubernetes scripts

barankyle:
Added Kubernetes scripts. Currently only the base/local ones have been updated from the templates they were copied from and work.

Updated sequelize files to conditionally construct connectionString
from environment variables if being run on Kubernetes (also indicated
by an environment variable).

Added a (temporary?) dev Dockerfile so it can be run locally not in prod mode.

Added section to README about running this in minikube.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,4 +5,9 @@
 node_modules
 lib/
 
+kubernetes/**
 .vscode
+config/local.json
+config/dev.json
+config/prod.json
+config/production.json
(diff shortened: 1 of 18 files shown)


shawticus:
Great. I will accept the PR, but we will want to keep all our devops in the xrchat-devops. Please coordinate with @mfouad and @MustafaSaber, our friends in Egypt who are working on the deploy and devops stuff as well.
Sorry that separation of concern and what everyone is doing have generally been a mess, I've gotten Github Projects + Issues set up so we should be able to start peeling cards off that and taking ownership tomorrow.

PR #21, opened 2020-04-08
Mailer

green9016:


src/services/email/email.service.ts
```diff
@@ -0,0 +1,32 @@
...
+declare module '../../declarations' {
+  interface ServiceTypes {
+    'email': Email & ServiceAddons<any>;
+  }
+}
+
+export default function (app: Application) {
+  // Initialize our service with any options it requires
+  app.use('/email', Mailer(smtpTransport({
+    host: 'email-smtp.us-west-2.amazonaws.com',
```
dmitrizagidulin, line 20: Can we pull this from config as well? I'd like to be able to test this locally using Mailtrap.io, for example.

src/services/auth-management/notifier.ts
```diff
@@ -0,0 +1,154 @@
+// Initializes the `authmanagement` service on path `/authmanagement`
+import { Application } from '../../declarations';
+import * as path from 'path';
+
+const pug = require('pug');
+
+export default function (app: Application) {
+  function getLink(type: string, hash: string) {
```
dmitrizagidulin, line 8: Do you mind extracting `getLink` and `sendEmail` to their own standalone functions, just above? I'd like to avoid embedded functions if at all possible.

dmitrizagidulin:
@green9016 looks good at first glance (just a couple of minor comments, inline.)
Let's make sure to pull the SMTP host from config as well -- I'd like to put together a testing script, for QA to test all of these steps using something like Mailtrap.io.

dmitrizagidulin:
@green9016 great, thanks!
Only thing remains is - can you extract the `getLink` and `sendEmail` functions to not be embedded?

green9016:
No problem.

PR #24, opened 2020-04-10
Changed names, fixed syntax, enforced stricer typescript...

shawticus:
Removed XR from object type names, renamed some things to have a cleaner, more readable paradigm. Multiple small figures related to typescript were fixed/added.


src/channels.ts
```diff
@@ -51,17 +51,17 @@ export default function (app: Application): void {
 
     console.log(hook)
 
-    console.log('Publishing all events to all authenticated users. See `channels.js` and https://docs.feathersjs.com/api/channels.html for more information.') // eslint-disable-line
+    console.log('Publishing all events to all authenticated user. See `channels.js` and https://docs.feathersjs.com/api/channels.html for more information.') // eslint-disable-line
 
-    // e.g. to publish all service events to all authenticated users use
+    // e.g. to publish all service events to all authenticated user use
```
dmitrizagidulin, line 56: Heh, so, this is fine to leave as is, but a good example of why an unattended/non-interactive search and replace leads to grammar weirdness. :)

src/services/auth-management/utils.ts
```diff
@@ -1,16 +1,12 @@
...
-  const host = process.env.APP_HOST ?? ''
-  const url = host + 'magicLink' + `?type=${type}&token=${hash}`
-  return url
-}
+export const getLink = (type: string, hash: string): string =>
+  (process.env.APP_HOST ?? '') + 'magicLink' + `?type=${type}&token=${hash}`
 
-export async function sendEmail (app: Application, email: any): Promise<void> {
-  return await app.service('email').create(email).then(() => {
+export const sendEmail = async (app: Application, email: any): Promise<void> =>
```
dmitrizagidulin, line 7: Why change from a named function to an assignment here?

src/services/object/object.class.ts
```diff
@@ -1,7 +1,7 @@
 import { Service, SequelizeServiceOptions } from 'feathers-sequelize'
 import { Application } from '../../declarations'
 
-export class XrObjects extends Service {
+export class Objects extends Service {
```
dmitrizagidulin, line 4: Since all the other services are non-plural names now, maybe it makes sense to make this `XrObject`, to match? ("Object" is too general-purpose anyway, this is specifically an in-xr object).

PR #31, opened 2020-04-12
fix issue in dockerignore

mfouad:
Fixes a docker issue ommitting env vars decalred in production.json in dockerignore
I think @barankyle did it to protect production.json secrets. but I added a comment to avoid people putting secrets there

it would be nice to pull production.json from a k8s configMap, but it can't happen becae configMaps map to volumes
also we can eventually just use NODE_CONFIG="{config json content}" if config grows

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,7 +7,4 @@ lib/
 
 kubernetes/**
 .vscode
-config/local.json
-config/dev.json
 config/prod.json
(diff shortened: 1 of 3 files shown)


dmitrizagidulin:
@mfouad 
> just write env variable names in the values like below

Does that mean that server now pulls _all_ of its config from env variables?

PR #32, opened 2020-04-13
fix authorize

green9016:


dmitrizagidulin:
Reviewing. (goes with https://github.com/xrchat/xrchat-client/pull/28 on the client side).

dmitrizagidulin:
Tested, works. Merging.

PR #41, opened 2020-04-15
Added additional fields to public_video schema.

barankyle:
Updated description field to be 1000 characters max.


barankyle:
We'll need to make sure to update the public_video table with the column additions/changes in this.

PR #46, opened 2020-04-17
fix can not read toString of undefined

green9016:


dmitrizagidulin:
Adresses issue https://github.com/xrchat/xrchat-client/issues/31#issuecomment-614765024.
Tested, working.

Issue #33, opened 2020-04-13
Add port and 'secure' to SMTP config


dmitrizagidulin:
Currently our SMTP config looks like this:

https://github.com/xrchat/xrchat-server/blob/master/src/services/email/email.service.ts

```
{
    host: process.env.SMTP_HOST,
    secure: true,
    auth: {
      user: process.env.SMTP_USER,
      pass: process.env.SMTP_PASS
    }
  }
```

Need to:

**1)** Add a PORT config (default to `2525`).

**2)** Load the `secure` argument from config (right now it's hardcoded to `true`. Which is the correct default for deployment, but makes it harder to do Mailtrap.io local testing).

dmitrizagidulin:
Assigning to @Vickyogesh.

dmitrizagidulin:
Implemented by @Vickyogesh in PR #47, closing.

PR #47, opened 2020-04-17
issue #33

Vickyogesh:


src/services/email/email.service.ts
```diff
@@ -16,7 +16,8 @@ declare module '../../declarations' {
 export default (app: Application): void => {
   app.use('/email', Mailer(smtpTransport({
     host: process.env.SMTP_HOST,
-    secure: true,
+    port: (process.env.SMTP_PORT) ? +process.env.SMTP_PORT : 2525,
```
dmitrizagidulin, line 19: Is the `+` in `+process.env` a typo?

Vickyogesh: No, it is to convert string to number.

dmitrizagidulin: Thanks.

PR #67, opened 2020-04-21
Linted to pass tests

shawticus:


dmitrizagidulin:
@shawticus I fixed the `collection-entity` test, and skipped the `uploads` test temporarily, to unblock merging. (I also opened a tracking issue to fix the uploads test, https://github.com/xrchat/xrchat-server/issues/73, so take a look at that when you have a chance.)

PR #68, opened 2020-04-21
Fixed issues with convert-video hook.

barankyle:
Asking for format 'best' usually only gets 720p, so now yt-dl is
executed directly to get the best video-only file @1080p and the best
audio-only, which are then automatically combined.

Executing yt-dl directly fixed issues with Vimeo DASH downloads
triggering 'end' when only the .mpd file had been downloaded.


dmitrizagidulin:
Merging into dev.

PR #78, opened 2020-04-24
Sequelize db fixes

shawticus:
Fixes:
- Race condition from seeding in last push (db not created yet) -- seed is now promised.
- All of the objects and types are seeded and/or faked, just need to fix some foreign key tables
- 'npm run dev-reinit-db' will force the DB to refresh and resync, which should fix most dev user migration issues (although proper migration is also welcome for prod!)


dmitrizagidulin:
@shawticus This looks promising. Can we fix the compile errors? (otherwise, can't deploy this to dev)

PR #86, opened 2020-04-26
Fixed resource relationships (many, to many), added composite keys fo…

shawticus:
Fixed resource relationships (many, to many), added composite keys for user and resource relationships


dmitrizagidulin:
👍

PR #108, opened 2020-04-30
Feature: GraphQL Mutations + Subscriptions for all Services

shawticus:
To test: run this branch, go to localhost:3030/graphql, try an example query and pull out the shelf on the right to view introspected types. Currently everything is open and unauthorized, but we can dev with it.


shawticus:
@dmitrizagidulin please hold on merging this -- need to add connected account types, enums are not well handled. Tests are failing due to some obscure syntax crap inside a dependency, we might need to fork and update graphql-compose ourselves?

shawticus:
PR is ready but dependency is still failing tests :-/

PR #112, opened 2020-05-04
Added UUID generation hook. Updated how remote resource paths are generated to prevent collisions.

barankyle:
Upload service now generates UUID and uses it as an identifier for all
related resources. So, if a file is uploaded with a thumbnail, the
static-resource for the file will have that UUID, the stored path will
start with the UUID, and both the file and its thumbnail will be stored
under that UUID path, e.g. abcd1234/video/cool-video.mp4 and
abcd1234/image/cool-video-thumbnail.jpg

/video endpoint is still generating "public" videos, so they are all
grouped under the "UUID" of "public". Under that, it's now <service>/
<service-id>/<type>/<name>, e.g.
public/youtube/abcd1234_m/video/manifest.mpd.
This may change in the future once we implement a better way to indicate
public vs. private videos.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "@types/lodash": "^4.14.150",
     "@types/multer": "^1.4.3",
     "@types/nodemailer": "^6.4.0",
+    "@types/uuid": "^7.0.3",
     "acorn": "^7.1.1",
     "apollo-server": "^2.12.0",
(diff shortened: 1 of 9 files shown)


barankyle:
Did a lot more work on top of this, easier to just merge that PR that contains all of this code.

Issue #142, opened 2020-05-11
Fix Travis CI MySQL connection so tests pass


shawticus:
To test: Uncomment any of the tests (Access Control is as good as any) and push. CI should connect to MySQL, but MariaDB isn't running and server isn't connected so it fails.

PR #153, opened 2020-05-12
Refactor DB config, enable Travis.ci db for integration tests.

dmitrizagidulin:
Addresses issue #142.

### DIFF
--- a/test/models/collection.test.ts
+++ b/test/models/collection.test.ts
@@ -1,14 +1,17 @@
 // // TODO: Add collection type association
 
 import app from '../../src/app'
+import { sequelize } from '../../src/models'
 
 describe('CRUD operation on \'Collection\' model', () => {
   const model = app.service('collection').Model
 
   before(async () => {
-    setTimeout(() => {
-      console.log('Waited for thirty seconds before test started.')
-    }, 30000)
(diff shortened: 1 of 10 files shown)


dmitrizagidulin:
WIP. Database is starting up, integration test is connecting. Now just need to add sequelize migration to the Travis script, so that the table gets initialized.

test/models/collection.test.ts
```diff
@@ -1,51 +1,54 @@
...
-//   })
-// })
+import app from '../../src/app'
+import { sequelize } from '../../src/models'
+
+describe('CRUD operation on \'Collection\' model', () => {
+  const model = app.service('collection').Model
+
+  before(async () => {
+    await sequelize.sync()
```
dmitrizagidulin, line 10: We will not want to do this operation before every test file. For the moment, we can just run it once, before the very first set of integration tests. (In this case, it's the Collection tests, since that's the only one I uncommented.)

dmitrizagidulin:
@shawticus ready for review.

shawticus:
@dmitrizagidulin Unfortunately, it's the same but different issue -- tests are failing because tables don't exist now -- we need to create the XRChat table and populate it before trying the tests.

PR #155, opened 2020-05-12
updated existing test cases according to TODOs

vinayqueppelin:
I have resolved the CRUD operation test cases issues and also added test cases for new models also.Please verify it by running  npm test.


shawticus:
@dmitrizagidulin please review. All tests pass locally but fail because the tables don't exist in CI

shawticus:
tests passed!

PR #156, opened 2020-05-12
Moved user thumbnail reference to user-settings.

barankyle:
Original fix was backwards, and correcting it on user would have resulted in a circular
dependency (static-resource has a user which has a static-resource).


barankyle:
Not the direction we're going for this.

PR #166, opened 2020-05-17
friend search

green9016:
https://github.com/xrchat/xrchat-client/issues/137


barankyle:
I made some updates to what green9016 did. One, I felt it was better to have create, remove, and update/patch logic in separate services instead of funneling everything through create. Two, I added a fourth userRelationShipType, 'blocking'. Now, if user A blocks user B, A->B relationship is 'blocking' and B->A relationship is 'blocked'.

I also got rid of userId as an input to the services and instead am pulling it from the authenticated user.

PR #169, opened 2020-05-19
Video conversion fixes

barankyle:
Added support for stereoscopic videos. Passing in metadata.stereoscopic: true will make ffmpeg crop out the right eye and rotate + stretch the left eye to match how mono videos look.

Fixed omission of parentResourceId on videos' thumbnails.

Thumbnail will be grabbed from source if metadata.thumbnail_url is null or a string of 0 length.

Fixed some bugs in mapping of urls to correct db entry caused by shared references to context.

Changed staticResource.type references to staticResource.staticResourceType.

Made staticResource.description field 1024 characters long instead of default 256.

Fixed staticResourceType seeder path to point to correct table.

### DIFF
--- a/src/hooks/convert-video.ts
+++ b/src/hooks/convert-video.ts
@@ -65,10 +65,10 @@ export default async (context: any): Promise<void> => {
       fileId = match[1]
       if (/youtube.com/.test(url) || /youtu.be/.test(url)) {
         context.params.videoSource = 'youtube'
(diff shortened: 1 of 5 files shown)


mrhegemon:
Please use this template going forward:

1. What is the purpose of this PR?
2. What does issue #s are solved by this PR?
3. What unit tests have been put in place?
4. How does our QA team test this PR?
5. What files or features are including in this PR that are not related to the main feature?

PR #173, opened 2020-05-22
Admin videos

barankyle:
**What is the purpose of this PR?**
Updating the backend to work with the admin console on the client.
Preparing content to be gated behind one or more paid tiers.
Better implement staticResource patching so that changing the thumbnail_url on
a video will cache the new thumbnail and remove the old one.

**What does issue #s are solved by this PR?**
No specific issues

**What unit tests have been put in place?**
None

**What files or features are including in this PR that are not related to the main feature?**
New script script/make-user-admin.js
Updated field 'staticResource.mime_type' to 'staticResource.mimeType'.
Added table subscriptionLevel and added it as a field to staticResource.

**How does our QA team test this PR?**
This is most easily tested in conjunction with the branch 'admin-video' of client:

Sign up as a user in the client.
Find out what the user's ID is. If testing locally, you can run `mysql -h 127.0.0.1 -u server -p` and enter the local password; then do `use xrchat`, then `select * from user` and note the ID.
From the top level of the package, run `node scripts/make-user-admin.js --userId=<user_id>`. It should return OK and, if you inspect the user in the SQL database, its userRole field should be 'admin'.

Back in the client, refresh the page and open the Nav menu in the upper right. There should be a new option 'Admin Console'. Click on that to be taken to the admin console.
From here, you can make new videos, edit existing ones (other than the URL of the video, didn't feel like making patching handle that), and delete existing ones.
You can make a video, patch its thumbnail_url and refresh the page after a few seconds to see that it's changed. Deleting a video should remove it from the admin console after a refresh (didn't implement subscriptions to listen for that), and if you check out the static_resource table after it's removed there should be nothing remaining (assuming you deleted all of the videos locally).

Note that, as videos are still all being uploaded to S3 into a shared public folder, if you delete a video, all other references to it on other people's local setups, as well as on dev, staging, and beta, will be broken. You can upload it again to put it back and fix all those references. Once the media server setup is in place, I will probably get rid of the shared public sourcing so that we don't have this issue.

### DIFF
--- a/scripts/make-user-admin.js
+++ b/scripts/make-user-admin.js
@@ -1,5 +1,4 @@
 const dotenv = require('dotenv')
-
 const cli = require('cli')
 const Sequelize = require('sequelize')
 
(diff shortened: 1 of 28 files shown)


scripts/make-user-admin.js
```diff
@@ -0,0 +1,72 @@
+const dotenv = require('dotenv')
+
+const cli = require('cli')
+const Sequelize = require('sequelize')
+
+dotenv.config()
+
+const db = {
```
dmitrizagidulin, line 8: This needs to be:

```
import { db } from '../src/db-config'
```
(to avoid duplicating configuring the db).

dmitrizagidulin: Fixed.

scripts/make-user-admin.js
```diff
@@ -0,0 +1,60 @@
...
+
+cli.enable('status')
+
+cli.parse({
+	userId: [false, 'UserID to make admin', 'string']
+});
+
+cli.main(async function(args, options) {
+	try {
+		const sequelizeClient = new Sequelize({
```
dmitrizagidulin, line 17: The indentation here is a bit wonky :)

dmitrizagidulin: Fixed.

PR #192, opened 2020-05-28
Pugemails

syedsimanta03:
Please check now


dmitrizagidulin:
Tested, email template is working,
 here is the login link email:
Screen Shot 2020-05-29 at 12 24 42 PM

So, +1 to this PR, I propose we do this (I'll take on the action item):

* merge it into a local `pugemails` branch
* Update/fix the config values
* Change the email wording from 'Sign In' to 'Log In' (to match the current Log In link in the client).

Issue #204, opened 2020-05-29
Refactor config system, pull everything from ENV variables


dmitrizagidulin:
* Deprecate the current system, which is a mixture of `dotenv` and  [`node config`](https://www.npmjs.com/package/config) / [`@feathersjs/configuration`](https://docs.feathersjs.com/api/configuration.html) system
* Pull in all configs via ENV variables
* Optionally, use [dotenv flow](https://www.npmjs.com/package/dotenv-flow) package to help manage the ENV vars.

barankyle:
Implemented as part of #205

PR #205, opened 2020-05-29
Refactor config system

dmitrizagidulin:
Implements issue #204.

Currently, we are using a mix of configuration tech:
1. [`@feathers/configuration`](https://docs.feathersjs.com/api/configuration.html) (which is a wrapper around [`node-config`](https://www.npmjs.com/package/config)). This is what layers `config/default.json` and `config/NODE_ENV.json`.
2. [`dotenv`](https://www.npmjs.com/package/dotenv) package, which reads in ENV vars and (inconsistently) loads them into the configuration.
3. Using `process.env.WHATEVER` directly in the code.

This PR is an attempt to unify the config system, by doing the following:
1. All variables are loaded from ENV variables, using `process.env.WHATEVER`.
2. This is done in `src/config.ts`, which is _also_ where all defaults live.
3. [`dotenv-flow`](https://www.npmjs.com/package/dotenv-flow) is OPTIONALLY used to help manage a developer's local environment variables.


.gitignore
```diff
@@ -74,7 +74,7 @@ typings/
 
 # dotenv environment variables file
 .env
```
mfouad, line 76: I think .env should be commited to git, to be used as the default
this way the code will run even if a developer does not have any env values

src/config.ts
```diff
@@ -0,0 +1,148 @@
...
+ * Server / backend
+ */
+const server: any = {
+  mode: process.env.SERVER_MODE ?? 'media',
+  hostname: process.env.SERVER_HOSTNAME ?? 'localhost',
+  port: process.env.SERVER_PORT ?? 3030,
+  // Public directory (used for favicon.ico, logo, etc)
+  publicDir: process.env.SERVER_PUBLIC_DIR ?? path.resolve(__dirname, '..', 'public'),
+  // Used for CI/tests to force Sequelize init an empty database
+  performDryRun: process.env.PERFORM_DRY_RUN ?? false,
```
mfouad, line 34: a major pitfall in env vars is that they are always strings, 
so if 
`export PERFORM_DRY_RUN=false`
it is loaded as a string "false"
`if ("false") true; else false; // returns true`

so if we define this default value in .env and we want it to be false
it muse be like this:
`export PERFORM_DRY_RUN=`
empty string is false in JS

another option is to rewrite this code to be 
`performDryRun: process.env.PERFORM_DRY_RUN=="true" ?? false,`

Note: same issue will happen with SERVER_PORT but JS automatic coersion will fix that automatically.

src/config.ts
```diff
@@ -0,0 +1,148 @@
...
+import url from 'url'
+import { inspect } from 'util'
+// Load all the ENV variables from `.env`, then `.env.local`, into process.env
+import dotenv from 'dotenv-flow'
+dotenv.config()
+
+/**
+ * Database
+ */
+const db: any = {
```
mfouad, line 12: consider removing `any` to allow the compiler to warn a developer if a config member changes or get renamed in this file and he forgets to fix it in the file that consumes the config.

it will also get rid of most of the `config.db.something as string`

.gitignore
dmitrizagidulin: yeah, agreed, I'll add that shortly.

src/config.ts
dmitrizagidulin: Ahh right thanks! Will update.

dmitrizagidulin: Ok, check, will fix.

.gitignore
dmitrizagidulin: Fixed.

src/config.ts
dmitrizagidulin: Fixed.

src/config.ts
dmitrizagidulin: Removed `any` from config.

PR #208, opened 2020-05-30
Integrated Spoke as an npm package

shawticus:
This pulls in xrchat-spoke from npm. When you start the server in dev, it'll automatically launch spoke.

@dmitrizagidulin The thing we need to think about this is devops -- all we're doing here is copying a few files back into the spoke node_modules folder, so we might be able to pull those files from this git into our spoke CI/CD at build time... I'm open.

The main thing here is that we can modify our routes and stay on vanilla spoke. Robert wants to do this with vanilla Spoke--make it an npm package-- so we will push this feature back to them and toss the xrchat-spoke repo when appropriate.


dmitrizagidulin:
@shawticus Very cool. Yeah, let's chat about this on a call sometime soon, see what our options are in terms of devops. You know I love me some monorepos, but I wonder if there's other options, etc.

dmitrizagidulin:
Closing (per conversation with @shawticus) - this is moving to the `-client` repo.

PR #212, opened 2020-06-01
Updated make-user-admin script

barankyle:
Made make-user-admin script into .ts file. Updated it to take in the user's email instead of ID; it now looks up a magiclink identity-provider with that email and gets the userId from there.

New run command is npx ts-node make-user-admin.ts --email=<email>


scripts/make-user-admin.ts
```diff
@@ -0,0 +1,82 @@
...
+dotenv.config()
+import { db } from '../src/db-config'
+
+cli.enable('status')
+
+cli.parse({
+  email: [false, 'Email of user to make admin', 'string']
+});
+
+cli.main(async function(this: any, args: any, options: any) {
```
dmitrizagidulin, line 14:
```suggestion
cli.main(async (this: any, args: any, options: any) => {
```

scripts/make-user-admin.ts
```diff
@@ -0,0 +1,82 @@
...
+const Sequelize = require('sequelize')
+
+dotenv.config()
+import { db } from '../src/db-config'
+
+cli.enable('status')
+
+cli.parse({
+  email: [false, 'Email of user to make admin', 'string']
+});
```
dmitrizagidulin, line 12:
```suggestion
})
```

PR #217, opened 2020-06-05
Chat group party

shawticus:
Making sure this is done.


dmitrizagidulin:
@shawticus how does one test this? (Is there a corresponding front-end branch?)

PR #218, opened 2020-06-05
add invite friend

shawticus:
add get my friends


dmitrizagidulin:
@shawticus how does one test this? (Is there a corresponding front-end branch?)

PR #305, opened 2020-06-25
Moved kubernetes folder into ops folder

shawticus:
The monolith now contains our docs and ops. Instead of having /ops and /kubernetes, we should combine the two. This moves the kubernetes folder into ops. We might need to fix paths and set up ops with this repo appropriately.


barankyle:
This stuff can actually all be deleted. It was what I tried to throw together a few months ago when Kai wanted to see something running in a production environment, but it never worked.

barankyle:
Whoops, accidentally merged it in when I meant to merge in a different PR, but since it's all unused K8s scripts that shouldn't cause any problems.

PR #406, opened 2020-08-20
 Fix post process shaders, add webpack to server

shawticus:


shawticus:
@barankyle @SaadAhmedSiddiqui 
This is an attempt to integrate the work from the open branches, as they were not pulled in before the switch to monorepo.
If you could just review and make sure that nothing important was lost in the merge. Thanks

PR #442, opened 2020-08-31
fix: clearEventQueues breaks some event listeners of queries

XiaNi:
function clearEventQueues assignes new array refs, and breaks some event listeners that are using that references


XiaNi (commented):
it was breaking this eventList reference
```ts
                  const eventList = (this.queryResults[queryName][eventName] = []);
                  query.eventDispatcher.addEventListener(QUERY_COMPONENT_CHANGED, entity => {
                    // Avoid duplicates
                    if (!eventList.includes(entity)) {
                      eventList.push(entity);
                    }
                  });
```

PR #555, opened 2020-10-22
fixes

HydraFire:
I refactored the code a bit in the character init, moved the assigned values into the component character.

I slightly refactored the raycastbox selection code, fixed the error due to which the boundingbox was at the 0 coordinate, now it is updated in the transform component. Transferred the constants of the distance and the width and height of the selection to the follow camera component.

Fixed offsets for vehicle colliders and moved to the VechicleBody component

### DIFF
--- a/packages/engine/src/interaction/behaviors/interactBoxRaycast.ts
+++ b/packages/engine/src/interaction/behaviors/interactBoxRaycast.ts
@@ -1,19 +1,14 @@
+import { Object3D, Ray, Raycaster, Vector3, Vector2, Mesh, Frustum, Matrix4, Box3 } from "three";
 import { Behavior } from "../../common/interfaces/Behavior";
 import { Entity } from "../../ecs/classes/Entity";
 import { InteractBehaviorArguments } from "../types";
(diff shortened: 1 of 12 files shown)


XiaNi (approved):

packages/engine/src/interaction/behaviors/interactBoxRaycast.ts
```diff
@@ -1,20 +1,15 @@
+import { Object3D, Ray, Raycaster, Vector3, Vector2, Mesh, Frustum, Matrix4, Box3 } from "three";
 import { Behavior } from "../../common/interfaces/Behavior";
 import { Entity } from "../../ecs/classes/Entity";
 import { InteractBehaviorArguments } from "../types";
-import { getComponent, getMutableComponent, hasComponent } from "../../ecs/functions/EntityFunctions";
+import { getComponent, getMutableComponent, hasComponent } from "../../ecs/functions/EntityFunctions";\
```
XiaNi, line 5: remove "\" from end of line

XiaNi (changes requested):

PR #631, opened 2020-11-10
Changed all imagePullPolicies from Always to IfNotPresent.

barankyle:
Gameservers spinning up and down may hit Docker rate limits.
It's also just silly to always get the image if it's present,
since we never get :latest, and instead get one tagged with
the commit number, which will never be overwritten.

Rolled back networking version to v1beta1 since full release v1 is
only in K8s 1.19+, and EKS only goes up to 1.18 right now.

Changed some hook return types to 'any' since something was
causing TS to not like the actual types.


Ryuno-Ki:
Note, that `unknown` is preferred over `any` these days.

PR #733, opened 2020-11-21
VRM import with anim

bodamat:
Created a VRM import for models and avatars. You can choose a VRM avatar from the avatars list in the hamburger menu. Animation works, but not fine, because I don't have good knowledge about a skeleton and can't retarget.


shawticus (commented):
Removed a couple of non-necessary infrastructure changes. Will work on animation set before we merge PR

PR #770, opened 2020-11-27
668 interactable component in editor

tanya-vykliuk:


XiaNi:
i decided to add client side part

PR #853, opened 2020-12-06
initializeNetworkObject with transform

XiaNi:
added override to createNetworkPrefab to be able to override prefab components data on creation, like transform, or avatar id.
may be it's too messy...


shawticus:
Looks good to me. I can see how it would be much easier to set those values on initialization then trying to initialize all this stuff, wait, then set avatar ID, etc.

PR #941, opened 2020-12-23
merge dev to resolve conflicts

XiaNi:


XiaNi:
not sure it works, will merge tomorrow morning after all checks (tests, load/run client)

shawticus:
Seems to work correctly! But let's push it to production and see how it does on the server :)

PR #1048, opened 2021-01-13
1010 resolution scaling

bodamat:
Resolution scale and post-process change when fps slow or backwards. Check to please that works ok

I see flashing at the sharp change of post-process. I'm not sure it's ok

### DIFF
--- a/packages/engine/src/renderer/systems/WebGLRendererSystem.ts
+++ b/packages/engine/src/renderer/systems/WebGLRendererSystem.ts
@@ -261,8 +261,10 @@ export class WebGLRendererSystem extends System {
           break;
       }
 
-      Engine.renderer?.setPixelRatio(window.devicePixelRatio * this.scaleFactor);
-      this.prevQualityLevel = this.qualityLevel;
+      if (Engine.renderer) {
+        Engine.renderer.setPixelRatio(window.devicePixelRatio * this.scaleFactor);
+        this.prevQualityLevel = this.qualityLevel;
+      }
     }
   }
 
--- a/packages/engine/src/renderer/systems/WebGLRendererSystem.ts
+++ b/packages/engine/src/renderer/systems/WebGLRendererSystem.ts
@@ -42,7 +42,8 @@ export class WebGLRendererSystem extends System {
   scaleFactor: number = 1
   downGradeTimer: number = 0
   upGradeTimer: number = 0
-  qualityLevel: number = 0
+  maxQualityLevel: number = 4
+  qualityLevel: number = this.maxQualityLevel
   prevQualityLevel: number = this.qualityLevel
 
   // fps counting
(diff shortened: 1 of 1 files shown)


packages/engine/src/renderer/systems/WebGLRendererSystem.ts
```diff
@@ -159,16 +182,76 @@ export class WebGLRendererSystem extends System {
...
+  changeQualityLevel(delta: number) {
+    // start timer when fps changes
+    if (this.fps <= 45) {
+      this.downGradeTimer += delta;
+      this.upGradeTimer = 0;
+    }
+    else if (this.fps >= 55) {
+      this.upGradeTimer += delta;
+      this.downGradeTimer = 0;
+    }
```
XiaNi, line 224: reset both timers if fps is not lower 45 and not higher then 55
```
    else {
      this.upGradeTimer = 0;
      this.downGradeTimer = 0;
    }
```

packages/engine/src/renderer/systems/WebGLRendererSystem.ts
```diff
@@ -159,16 +182,76 @@ export class WebGLRendererSystem extends System {
...
         resize(entity);
-        getComponent<RendererComponent>(entity, RendererComponent).composer.render(delta);
+
+        if (this.qualityLevel >= 2) {
+          getComponent<RendererComponent>(entity, RendererComponent).composer.render(delta);
+          Engine.renderer.outputEncoding = LinearEncoding; // need this if postprocessing is used
+        }
+        else {
+          Engine.renderer?.render(Engine.scene, Engine.camera);
+          Engine.renderer.outputEncoding = sRGBEncoding; // need this if postprocessing is not used
```
XiaNi, line 206: if in line 205 you are not sure that Engine.renderer is set, then add same optional chaining in line 206.
Engine.renderer?.outputEncoding = sRGBEncoding;

bodamat: unfortunately, it throws an error in line 206

bodamat: The left-hand side of an assignment expression may not be an optional property access.

XiaNi: oh, sure, than
```js
if (Engine.renderer) {
          Engine.renderer.render(Engine.scene, Engine.camera);
          Engine.renderer.outputEncoding = sRGBEncoding; 
}
```

bodamat: Thanks. This is better solution to the problem

packages/engine/src/renderer/systems/WebGLRendererSystem.ts
```diff
@@ -159,16 +183,89 @@ export class WebGLRendererSystem extends System {
...
+        case 3:
+          this.scaleFactor = 1;
+          break;
+        default:
+          this.scaleFactor = 1;
+          break;
+      }
+
+      Engine.renderer?.setPixelRatio(window.devicePixelRatio * this.scaleFactor);
+      this.prevQualityLevel = this.qualityLevel;
```
XiaNi, line 265: change it to
```js
if (Engine.renderer) {
      Engine.renderer.setPixelRatio(window.devicePixelRatio * this.scaleFactor);
      this.prevQualityLevel = this.qualityLevel;
}
```

XiaNi (approved):

PR #1083, opened 2021-01-15
1010 resolution scaling

bodamat:
So,
- I did a more stable change resolution. 
- Switch from fps to ms for render. It's a more stable counter. 
- Add a few more levels to smooth change (+2 levels).

Can you check please if all works well? Would be nice if we can test with a real slow device.


bodamat:
I see the image in the dev is dark. We can change to sRGBEncoding in renderer.outputEncoding and image become light.

PR #1091, opened 2021-01-18
1054Throwing warnings if Docker isn't running or DB isn't connected

ANDFEWD:


XiaNi (changes requested):
```ts
  const processDocker = processList.find(
    c => c.cmd.match(/docker-compose/)
  );
  const processMysql = processList.find(
    c => c.cmd.match(/docker-proxy/)
  );
  const processAgones = processList.find(
    c => c.cmd.match(/start-agones/)
  );
```
add check if there are no docker and dockerMysql - check local mysql, or check 3036 port.

second is not required - just optional improvement, show all checks result:
```
one or more of required services are not started:
docker [present]
docker-proxy [missing!]
agones [missing!]
```

shawticus (approved):
Looks great

XiaNi (approved):

PR #1227, opened 2021-02-11
refreshed login/sign in/forger pass; -duplicate some auth components …

tanya-vykliuk:
…for app and change input fields; added pass mask hiding


packages/client-core/components/social/Login/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+  loginUserByOAuth?: typeof loginUserByOAuth;
+  logo: string;
+  isAddConnection?: boolean;
+}
+
+const FlatSignIn = (props: Props) => {
+  const [view, setView] =  useState('login');
+  let enableUserPassword = config?.auth ? config.auth.enableUserPassword : false;
+  let enableGoogleSocial = config?.auth ? config.auth.enableGoogleSocial : false;
+  let enableFacebookSocial = config?.auth ? config.auth.enableFacebookSocial : false;
```
XiaNi, line 43: replace "let" with "const"

XiaNi (changes requested):

PR #1240, opened 2021-02-16
canvas cam fixes

HexaField:
Solves:
-- Prevent scrolling and hide scrollbars
-- Camera can still go below view plane, clamp to 1-3 degrees


HexaField:
@shawticus canvas.ondragstart may also solve `-- When leaving and returning, input on mouse turning doesn't work any more`. I have not seen it since this fix, but had trouble replicating it reliably before.

mrhegemon (approved):
looks good

PR #1271, opened 2021-02-24
Jsdoc

shawticus:


shawticus:
Still some issues getting thrown. Try running "yarn build" to see if anything comes up, or review the fail by clicking on the "x".

I think the problem is that "ts-ignore" entries are getting removed and this is causing Typescript linter errors.

mrhegemon (approved):
Great to have docs.

PR #1302, opened 2021-03-05
Offscreen canvas worker

HexaField:
fix stuff & resolve docker build errors


mrhegemon (approved):
loving the abstraction and events

PR #1328, opened 2021-03-10
Arcapp 147 crud for bookmark feed

tanya-vykliuk:


packages/server/src/services/feed-fires/feed-fires.class.ts
```diff
@@ -1,12 +1,102 @@
...
+   * @function find it is used to find specific users
+   *
+   * @param params user id
+   * @returns {@Array} of found users
+   */
+
+     async find (params: Params): Promise<any> {
+      const loggedInUser = extractLoggedInUserFromParams(params);
+      if (!loggedInUser.userId) {
+        return Promise.reject(new BadRequest('Could not add fire. Users isn\'t logged in! '));
```
XiaNi, line 31: i don't think we need this rejection here, or at least change it to "Could not get fires"

XiaNi (changes requested):

XiaNi (approved):

PR #1419, opened 2021-03-25
will try merge dev

HydraFire:


packages/engine/src/assets/components/AssetLoader.ts
```diff
@@ -16,7 +16,7 @@ export class AssetLoader extends Component<AssetLoader> {
   /** Class of the asset. Supported Asset classes can be found in {@link assets/enums/AssetClass.AssetClass | AssetClass}. */
   assetClass: AssetClassAlias = null
   /** the object will parse to Colliders when loading location */
-  parseColliders = true
+  dontParseModel = true
```
HexaField, line 19: let's not have negatives, can we replace this with `parseModel = false` as default?

packages/engine/src/scene/behaviors/createCollidersFromScaneData.ts
```diff
@@ -0,0 +1,71 @@
...
+import { AssetLoader } from '../../assets/components/AssetLoader';
+import { TransformComponent } from "@xr3ngine/engine/src/transform/components/TransformComponent";
+import { ColliderComponent } from '../../physics/components/ColliderComponent';
+import { RigidBody } from '../../physics/components/RigidBody';
+import { addColliderWithoutEntity } from '../../physics/behaviors/addColliderWithoutEntity';
+import { Network } from '../../networking/classes/Network';
+import { createNetworkRigidBody } from '@xr3ngine/engine/src/templates/interactive/prefabs/NetworkRigidBody';
+import { addCollidersToNetworkVehicle } from '@xr3ngine/engine/src/templates/vehicle/prefabs/NetworkVehicle';
+
+export const createCollidersFromScaneData: Behavior = ( entity: Entity, args: any ) => { //{ data:string, type: string,}
```
HexaField, line 13: mispelling in createCollidersFromScaneData. should be createCollidersFromSc**e**neData as well as in file name

HexaField (changes requested):

packages/engine/src/scene/behaviors/createCollidersFromScaneData.ts
HydraFire: sorry

packages/engine/src/assets/components/AssetLoader.ts
HydraFire: wait, let's be careful, I may already be confused, the fact is that this is the, default its to parse the model. If we don't know anything about the model yet, we parse it, and if we have already saved everything, then we don't need it anymore.

HydraFire: correctly then named, alreadyParsed

PR #1438, opened 2021-03-28
Fix:Skybox resetting after saving and reloading for equirectangle/cubemap

Abhishek-Pathak-Here:


mrhegemon (approved):
Thanks for the fix.

PR #1445, opened 2021-03-29
removed association of feed with static_resource

shawticus:


XiaNi:
very outdated branch, i'll remove it

Issue #1370, opened 2021-03-17
Remove next.js dependency from client-core


shawticus:
Since some of our partners use React but not next, we need to remove this dependency from client-core to extend the flexibility

PR #1475, opened 2021-04-02
Refactor remove next

NPatel10:
Fixed bugs in Client and Social package occured by removal of the 'next' library from the client-core.


NPatel10:
#1370 should fix this

Issue #1309, opened 2021-03-06
i18n


shawticus:
We should isolate all of the strings in our app and internationalize them.

In doing so, we want to also make all of these things customizable to the user.

PR #1476, opened 2021-04-02
Feature add i18n

NPatel10:
Added i18n support in Client package


NPatel10:
#1309 Added support in client package.

PR #1503, opened 2021-04-09
Post Processing Node

Abhishek-Pathak-Here:


packages/client-core/components/editor/inputs/ColorInput.tsx
```diff
@@ -65,16 +65,17 @@ const ColorInputPopover = (styled as any).div`
...
+      ({ hex }) => {
+        onChange(isValueAsInteger?(new Color(hex)).getHex():(new Color(hex)));
+      },
+      [onChange]
+    );
 
-  //initializing hexColor by getting hexString
-  const hexColor = "#" + value.getHexString();
+    //initializing hexColor by getting hexString
+    const hexColor = "#" + (isValueAsInteger?value.toString(16):value.getHexString());
```
shawticus, line 78: Code is right but please be aware of formatting -- if you use VSCode just run the formatter before pushing. Otherwise please format these things like:
```
isValueAsInteger ? value.toString(16) : value.getHexString()
```

packages/client-core/components/editor/Editor.tsx
```diff
@@ -2223,6 +2224,36 @@ export default class Editor extends EventEmitter {
...
     return object;
   }
+/**
+ * [setObjectProperty used to set propery to selected object]
+ * @param {[type]}  propertyName
+ * @param {[type]}  value
+ * @param {Boolean} [useHistory=true]
+ */
+
+  setObjectProperty(propertyName, value,useHistory=true) {
```
shawticus, line 2234: Formatting
, useHistory = true

packages/engine/src/scene/classes/PostProcessing.ts
```diff
@@ -0,0 +1,149 @@
...
+import { HueSaturationEffect } from "../../renderer/postprocessing/HueSaturationEffect";
+import { LinearTosRGBEffect } from "../../renderer/postprocessing/LinearTosRGBEffect";
+import { OutlineEffect } from "../../renderer/postprocessing/OutlineEffect";
+import { SSAOEffect } from "../../renderer/postprocessing/SSAOEffect";
+import { ToneMappingEffect } from "../../renderer/postprocessing/ToneMappingEffect";
+
+/**
+ * @author Abhishek Pathak <abhi.pathak401@gmail.com>
+ */
+export default class PostProcessing extends Object3D{
```
shawticus, line 17: This should be a component that extends Component from engine/ecs.
However, we haven't unified our game engine and editor (the editor is Mozilla Spoke).
The editor doesn't use the engine, so there is no way to attach a component.
For now, let's keep this class, but we will basically need to duplicate it in order to make a component for the engine side. Later, when we unify the editor and engine, we can delete this class and only use entities + components.

shawticus (changes requested):
Please use a formatter to fix the readability. Instead of 'object=prop' it should be 'object = prop' and so on. We have eslint rules for this, just need to run the linter or fix in VSCode.
Otherwise this PR looks good. Please see comments that I wrote on the class. TL;DR, we will need a component version of the class as well for our engine, since it doesn't know anything about Object3D really, it's more about data on components attached to entities.

PR #1575, opened 2021-04-11
Vusr

shawticus:


packages/engine/package.json
```diff
@@ -58,6 +58,7 @@
     "three-mesh-bvh": "^0.3.7",
     "three-stdlib": "^1.2.1",
     "three-to-cannon": "^3.0.2",
+    "@tweenjs/tween.js": "18.6.4",
```
HexaField, line 61: as we're not using tween, we should not include it from our deps

packages/engine/src/initialize.ts
```diff
@@ -119,6 +121,14 @@ export const initializeEngine = async (options): Promise<void> => {
   }
 
   if (!useOffscreen || options.editor) {
+    await AnimationManager.instance.getDefaultModel()
+    registerSystem(RaycastSystem);
+    registerSystem(StateSystem);
+    registerSystem(CharacterControllerSystem);
+    registerSystem(PhysicsSystem);
```
HexaField, line 128: seems to be a merge error here doubly registering systems.

packages/engine/src/raycast/systems/RaycastSystem.ts
```diff
@@ -0,0 +1,31 @@
...
+export class RaycastSystem extends System {
+  constructor(attributes?: SystemAttributes) {
+    super(attributes);
+  }
+  execute(): void {
+    this.queryResults.raycasts?.all?.forEach((entity: Entity) => {
+      const raycastComponent = getMutableComponent(entity, RaycastComponent);
+      raycastComponent.hitResults = [];
+      const hitResults = []
+      // use bitmask, then do appropriate raycastScene() raycastPhysics() raycastCamera() or something
```
HexaField, line 16: should be using this system instead of a raycast class on the component

packages/engine/src/ui/systems/UIPanelSystem.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  panelContainer: Group;
+
+  constructor(attributes?: SystemAttributes) {
+    super(attributes);
+    this.panelContainer = new Group();
+    Engine.scene.add(this.panelContainer);
+    
+    
+    createPanelComponent({ panel: new UIPanelComponent(), raycast: new RaycastComponent() });
```
HexaField, line 22: rather than creating this in any context, we should be calling this function the client page

packages/engine/src/raycast/classes/Raycast.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    init(){
+        this.rayCaster = new Raycaster();
+        this.selectState = false;
+
+        this.mouse = new Vector2();
+        this.mouse.x = this.mouse.y = null;
+
+        window.addEventListener( 'pointermove', ( event )=>{
```
HexaField, line 27: window events are problematic for if we need to use the raycasting system on the server. this should be integrating with the RaycastSystem, ClientInputSystem and ActionSystem instead of a class that exists on components.

HexaField (changes requested):

packages/engine/src/initialize.ts
Realitian: is this means to remove line 
`    registerSystem(PhysicsSystem);`
 `    registerSystem(TransformSystem, { priority: 900 });` 
 ?

HexaField: i've solved this in the latest commits

packages/engine/src/raycast/systems/RaycastSystem.ts
HexaField: this has been solved in the latest commits

packages/engine/src/ui/systems/UIPanelSystem.ts
HexaField: this has been solved in the latest commit

packages/engine/src/raycast/classes/Raycast.ts
HexaField: this also doesn't work cross-platform

PR #1625, opened 2021-04-15
spit engine init into editor & engine init functions

HexaField:


shawticus (approved):

shawticus:
@HexaField just need to resolve conflicts

PR #1648, opened 2021-04-19
Vite

shawticus:


shawticus:
@barankyle this is done and merged with dev -- however there are some weird errors happening with the env-config.js, if you could just review this and pull it in if it's good for you

PR #1666, opened 2021-04-20
Remove webpack and next.js, replace with ts-node, vite, react-router

shawticus:


shawticus:
@barankyle This is *almost* done but I am getting an error
```Error: secretOrPrivateKey must have a value``` coming from the attempt to access the identity-provider service from the client. This happens on the social package and on the client package, and is blocking us from landing this. It's all connected to doLoginAuto.

PR #1683, opened 2021-04-22
Don't enable postprocessing from schema

Abhishek-Pathak-Here:


shawticus (changes requested):
@Abhishek-Pathak-Here I can't tell what's different because mostly it seems like we've just removed semicolons. I actually think we want to keep semicolons everywhere. Please review linter settings-- is it our linter settings? Or is it yours? But if we could just commit the changes that are actual code changes, it would be easier to track. Thanks.

PR #1696, opened 2021-04-23
Don't set PostProcessing from schema

Abhishek-Pathak-Here:


packages/engine/src/scene/behaviors/setPostProcessing.ts
```diff
@@ -5,9 +5,8 @@ export const setPostProcessing: Behavior = (entity, args: {}) => {
     const options=args["objArgs"]["options"];
     if(WebGLRendererSystem.instance!==undefined){
         const usePostProcessing=(options!==undefined && Object.keys(options).length>0);
-        if(usePostProcessing){
+        if(usePostProcessing && WebGLRendererSystem.usePostProcessing){
```
HexaField, line 8: we should set the schema no matter what, so get rid of `&& WebGLRendererSystem.usePostProcessing`

Abhishek-Pathak-Here: Done

PR #1717, opened 2021-04-25
Remove unnecessary dependencies from project

shawticus:


shawticus:
@AbhinavShah421 We actually don't want to keep any of the auto-generated docs in our github history, because it will be to be like millions of changes with every build commit. So we should git ignore and remove all these conflict files from cache.

AbhinavShah421:
> ignore

Okay I'll take care of this, thanks for you guidelines.

PR #1744, opened 2021-04-28
Game Manager System

HydraFire:
all main work done, want adding to dev for sync


packages/engine/src/templates/character/CharacterInputSchema.ts
```diff
@@ -45,10 +45,9 @@ const interact: Behavior = (entity: Entity, args: any = { }, delta): void => {
     unequipEntity(entity)
     return;
   }
-  if (isServer) {
-    interactOnServer(entity);
-    return;
-  }
+
```
HexaField, line 48: Should rename back to `interactOnServer`

PR #1795, opened 2021-05-05
Environment variable changes.

NPatel10:
- Added Variables in .env file which were missing.
- Structured the .env file.
- Scripted a function to add environment variables in config json file for client which will be replaced by the value of environment variables.
- Some Clean up


shawticus:
@barankyle please review this since it affects your deployment. And we need to make sure we're not in conflict hell with xr3ngine / xrengine

PR #1800, opened 2021-05-05
Key buttons stay continued bug fixes , some character fixes

HydraFire:
Key buttons stay continued bug fixes , some character fixes
Please, review part with Inputs and Buttons. As its now, its resolve bug, but don't know about joist or mobile, how its will be.


shawticus:
@HexaField Rebased over dev. Please review, since this interacts mostly with your code. Pull in when you feel confident.

PR #1824, opened 2021-05-07
Fix server w/ ClientStorage.ts

speigg:
Importing ImmortalDB dynamically to avoid errors on the server.

### DIFF
--- a/packages/engine/src/common/classes/ClientStorage.ts
+++ b/packages/engine/src/common/classes/ClientStorage.ts
@@ -1,23 +1,33 @@
-import { ImmortalDB } from 'immortal-db'
+import { isClient } from "../../common/functions/isClient";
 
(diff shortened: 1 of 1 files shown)


shawticus (commented):
I patched this with a dynamic import on ClientStorage itself-- if we import this it will still work with no conflicts. However-- is there a use case where we want client storage on a client that doesn't have a window object? (Could a Node.js instance ever be the client?)

speigg:
So we should probably just check for the existence of the `window` object?

speigg:
We do this kind of check elsewhere throughout the codebase. This should be sufficient: ```js
const isBrowserCheck = typeof window === 'object' && typeof document !== 'undefined'
```

shawticus:
@

PR #1867, opened 2021-05-18
Some Memory Leak bug fixes

NPatel10:


packages/engine/src/physics/systems/PhysicsSystem.ts
```diff
@@ -96,7 +96,7 @@ export class PhysicsSystem extends System {
       }
       */
       collider.body.addEventListener(CollisionEvents.COLLISION_START, (ev: ColliderHitEvent) => {
-        ev.bodyOther.shapes[0].config.collisionLayer == ev.bodySelf.shapes[0].config.collisionMask ? collider.collisions.push(ev):'';
+        (ev.bodyOther as any).shapes[0].config.collisionLayer == (ev.bodySelf as any).shapes[0].config.collisionMask ? collider.collisions.push(ev):'';
```
HexaField, line 99: this can be removed as the next update to three-physx will solve this type error

HexaField (changes requested):

PR #1875, opened 2021-05-19
Removed Scene Object loading schema

NPatel10:


HexaField (approved):
fixed a few gameobject related issues, looks good!

PR #1891, opened 2021-05-21
add interact and some console.warn to debug why interact on spawn obj…

HydraFire:
i try to debug why HaveBeenInteracted don't have effect, not found why, add interactive component, but addForce don't start on press E. And even isNamed dont run, dont understand what changed.
add interact and some console.warn to debug why interact on spawn object don't have effect


HexaField:
Why are we using addForce on the golf ball? If we are spawning it programatically, there's no reason to add the InteractiveComponent as we can put the GameObject tag directly on.

PR #1950, opened 2021-05-31
change how physics collisions events work

HexaField:
- physics collisions are now stored on the physics bodies internally by three-physx
- on colliders as `ColliderComponent.body.collisionEvents`
- on controllers as `ControllerColliderComponent.controller.controllerCollisionEvents`
- behaviors can be added to GameObject components under collisionBehaviors as `[role]: behavior` (example: `gameObject.collisionBehaviors['GolfBall'] = onCollideWithBall`)
- GameManagerSystem now queries for [ColliderComponent, GameObject], and runs behaviors


packages/engine/src/game/functions/functionsActions.ts
```diff
@@ -17,7 +17,7 @@ import { getEntityFromRoleUuid, getGame, getGameEntityFromName, getRole, getUuid
 // TODO: create enum actions
 const gameActionComponents = {
   'HaveBeenInteracted': HaveBeenInteracted,
-  'HasHadCollision': HasHadCollision
+  'HasHadCollision': GameObjectCollisionTag
```
HexaField, line 20: whoops, i must have missed this

HydraFire (approved):

packages/engine/src/game/functions/functionsActions.ts
HydraFire: i re-work this stuff, you see it in new PR

PR #1954, opened 2021-06-01
BPCEM runtime

Abhishek-Pathak-Here:


packages/engine/src/scene/behaviors/setReflectionProbe.ts
```diff
@@ -39,25 +36,32 @@ export const setReflectionProbe: Behavior = (entity, args: {}) => {
...
-              );
-              shader.fragmentShader = shader.fragmentShader.replace(
-                  '#include <envmap_physical_pars_fragment>',
-                  envmapPhysicalParsReplace
-              );
-        };
-      }
-    });
-  
+      Engine.entities.forEach(entity=>{
```
HexaField, line 39: This would be quite expensive on big scenes, we should push this to an added query for Object3DComponent in a system.

HexaField (changes requested):

packages/engine/src/scene/behaviors/setReflectionProbe.ts
Abhishek-Pathak-Here: ok on it

PR #1966, opened 2021-06-04
 Initialize engine function moved to Engine package and some linting issue fix

NPatel10:


packages/engine/src/renderer/functions/canvas.ts
```diff
@@ -0,0 +1,23 @@
...
+  return canvas;
+}
+
+export const createCanvas = () => {
+  if(isWebWorker || isNode) {
+    return;
+  }
+  const canvas = document.createElement("canvas");
+  document.body.appendChild(canvas);
+  configCanvasElement(canvas);
```
HexaField, line 21: we're passing a canvas element even though the function requires a canvas identifier? please double check this

PR #2096, opened 2021-06-15
Integration of bot dashboard 

kimenyikevin:


barankyle (commented):
* Seeing this error in the browser console. It doesn't appear to be blocking functionality, but it should be addressed: `Warning: Failed prop type: Invalid prop `options` of type `object` supplied to `ForwardRef(Autocomplete2)`, expected `array`.
    at Autocomplete2 (https://localhost:3000/node_modules/.vite/chunk-65GLPEEW.js?v=c257b422:1054:35)
    at WithStyles2 (https://localhost:3000/node_modules/.vite/chunk-CSRE2QZ6.js?v=c257b422:50:31)
    at form
    at div
    at CardContent2 (https://localhost:3000/node_modules/.vite/chunk-QBDS7DNT.js?v=c257b422:35:23)
    at WithStyles2 (https://localhost:3000/node_modules/.vite/chunk-CSRE2QZ6.js?v=c257b422:50:31)
    at div
    at Paper2 (https://localhost:3000/node_modules/.vite/chunk-KAJVER6G.js?v=c257b422:51:23)
    at WithStyles2 (https://localhost:3000/node_modules/.vite/chunk-CSRE2QZ6.js?v=c257b422:50:31)
    at Card2 (https://localhost:3000/node_modules/.vite/chunk-PVUHB6HY.js?v=c257b422:35:23)
    at WithStyles2 (https://localhost:3000/node_modules/.vite/chunk-CSRE2QZ6.js?v=c257b422:50:31)
    at CreateBot (https://localhost:3000/@fs/home/kyle/Development/Forks/xrengine/packages/client-core/src/admin/components/Bots/CreateBot.tsx:50:5)
    at ConnectFunction (https://localhost:3000/node_modules/.vite/react-redux.js?v=c257b422:332:48)`
* Make the command description in the Create Bot window non-mandatory
* Rename 'Explanation' to 'Description'
* Under the Add Command button, the div with class 'alterContainer' is still visible since it has a border, but it just looks like a black line since it has no contents. It shouldn't be visible if there are no children.
* The Instance and Location selectors aren't getting cleared when a new bot is created. This also appears to be creating problems when I try to make another one - despite the Instance selector having an instance ID selected, when I go to create a second bot without refreshing the page, the actual value being sent to the API is an empty string, which fails. I've tried clicking the 'X' to clear that selector, but when I click away, the instance IP is still visible in the div, but the value is still an empty string. I think the same thing is happening with the Location selector, too. (If you make sure that you can create multiple bots successfully without refreshing the page, it should address all of these issues).
* Client should validate that the selected instance/location is valid before sending the API request to create/patch a bot and display an error message similar to the command name being invalid.
* Empty string should not be a valid bot name.
* There should be a little space between the bottom of the 'XREngine Bots' column header and the first bot accordion selector.

kimenyikevin:
I have implement all requested changes, here is [PR](https://github.com/XRFoundation/XREngine/pull/2111)

PR #2150, opened 2021-06-19
# Integration of update bot on admin dashboard 

kimenyikevin:


barankyle (changes requested):
Overall it looks very good. There's some changes I'd like to see, many of them related to fields that should not be required or even editable at all.

For the bots, instances are not required. I think we have an idea that some bots may be assigned to a location in general, and will be spun up on each instance of that location that gets created, e.g. for a product showroom, there may be a greeter bot that isn't tied to a specific instance of that location, but rather gets made in every room. So, instance should not be required for creating or updating a bot.

I'd also like a little refresh button next to the location and instance selectors that re-fetches the list of locations and instances of the selected location, respectively. Right now, whatever instances of a location that are available when I go to the page are the only ones I can select, and I'd have to refresh the whole page to get an updated list.

With users, there are several fields that should be changed:
* Avatar should be a select with a limited choice. Those choices are 'static-resource's that have staticResourceType 'avatar'. As of now, the only one that should be there is 'Sonny'; you can get the name from the 'name' field on the static-resource.
* Instance and Party should not be editable at all. There should be no Select for either of them, and the "Don't see <X>? Create One" for each should be removed as well. These aren't fields we want admins changing, and they're not required.
* Invite Codes should not be editable by an admin. They're just randomized codes that should be made automatically by the backend for non-guests when they're created. There's currently an after patch hook on user that adds an invite code if a non-guest user doesn't have one; this should probably be copied over to the after create hook on users as well.

On the user View drawer, when I change someone's userRole from the little pencil chip under their name, the displayed role there doesn't change (but it does on the main list).

When I click Edit from the user View drawer, the Location is titled 'User role' by mistake. I think the "Don't see Location? Create One" should be removed from here - I don't see admins ever creating a Location for a specific user. 
All of the above comments about Instance/Party/Invite Codes not being editable at all and Avatar being selected from a discrete list apply to the Edit User drawer as well. 
The 'Clear' button on User Edit should be renamed to 'Cancel'. 'Clear' suggests that we're wiping the user's data or something similar.

On the Instance page, some of that functionality should be removed. Instances are pretty autonomous; we can't really be creating Instances manually, so the Create New Instance button should be removed, and the Edit button should probably also be deleted as well. I would have said remove the Delete button, but that would likely be useful if there's an Instance record that should have been removed when the corresponding gameserver finished but wasn't, so keep that.

kimenyikevin:
> 
> I'd also like a little refresh button next to the location and instance selectors that re-fetches the list of locations and instances of the selected location, respectively. Right now, whatever instances of a location that are available when I go to the page are the only ones I can select, and I'd have to refresh the whole page to get an updated list.
> 

This refresh is done in background at the moment, When I create new instance or location I can be able to get  it on the list during creation of bot.  Do we still need to add button ?

barankyle:
I'm not seeing it refresh the instance list on my local environment. I've got one tab on /admin/bots, with the new bot selector set to 'Test', and then in another tab I go to /location/test. When the location loads, there's an instance in the database, but the select for instance isn't clickable. Even if I have another location to select, selecting the other one and then selecting Test again still doesn't let me select an instance.

kimenyikevin:
### I have created [PR](https://github.com/XRFoundation/XREngine/pull/2195) for all requested changes.

PR #2274, opened 2021-07-01
# Refactor Admin dashboard for location 

kimenyikevin:
### Navicate on /admin/location I have changed UI view and I made some refactor as well as adding validation


barankyle:
It looks good, I'm just going to hold off on merging it in until location editing is done.

barankyle:
A couple last-minute improvements to the location page:

There's a mismatch on a couple of the columns. The one with the header 'Type' seems to contain the Scene name, and should be titled 'Scene'. I don't see a column containing the Type, and there should be one.

The column 'Video Enabled' appears to be using the wrong value. When I have a location where that value is true, it's displayed as false.

The editing modal is missing the toggles for Video Enabled, Global Media Enabled, Make Lobby, and Is Featured.

If I try to make a scene with the same name as one that already exists, there should be an error message but leave the create modal up. Currently, the modal is just closing without any indication that there was a problem.

kimenyikevin:
@barankyle I have pushed requested changes

mrhegemon (approved):

PR #2320, opened 2021-07-06
Remove Upload and share on Avatar 

kimenyikevin:


shawwwwwwww:
@kimenyikevin We don't necessarily want to remove this functionality though from our demo, etc. We want to make these things configurable so they are hidden in the event UI and shown in our /location/ UI, otherwise the features will break.

shawwwwwwww:
@kimenyikevin we should do this by prop injection
We have a custom Layout file called EventLayout.tsx -- this can pass props down to the menus. The menus should expose generic functionality like "enableSharing" as boolean props, and the layout will pass these props in

mrhegemon (approved):

PR #2337, opened 2021-07-07
envmap in Scene node

Abhishek-Pathak-Here:
Currently supports texture/color.


packages/engine/src/scene/functions/SceneLoading.ts
```diff
@@ -159,22 +160,7 @@ export class WorldScene {
               clearFromColliders(res);
             } else {
               parseModelColliders(entity, { asset: res, uniqueId: component.data.sceneEntityId });
-            }
-            if (isClient && component.data.textureOverride) {
```
HexaField, line 163: this must be kept for basscoast event

packages/engine/src/scene/functions/SceneLoading.ts
```diff
@@ -192,24 +178,19 @@ export class WorldScene {
...
-          createVideo(entity, component.data);
-          // }
-          // addComponent(entity, LivestreamComponent)
-          // } else if(component.data.isLivestream) {
-          // @todo
-          // addComponent(entity, LivestreamProxyComponent, { src: component.data.src })
+          if(!component.data.isLivestream) {
+            createVideo(entity, component.data);
+          }
+          addComponent(entity, LivestreamComponent)
```
HexaField, line 191: livestream things can be ignored for now, they are not fully implemented and can cause problems

HexaField (changes requested):

PR #2339, opened 2021-07-07
DJ animation Sync

NPatel10:


HexaField (changes requested):
instead of adding CharacterComponent to DJ, move CharacterComponent.mixer to AnimationComponent. keeps things cleaner and will make life easier after the basscoast event. thanks.

HexaField (approved):

PR #2372, opened 2021-07-08
Editor > Move "viewport" row buttons into top row and style publish button

kimenyikevin:


shawwwwwwww:
@kimenyikevin could you please resolve the deepcode error, and post a screenshot of what it looks like on your machine?

shawwwwwwww (approved):

PR #2421, opened 2021-07-12
Move creating PhysX worker to be contained in Physics System

HexaField:
Due to vite and esbuild not support `import.meta.url`, we have had to create the webworker seperate for node and the browser. This PR adds tools to our compilation that solve this issue.

- [x] vite
- [ ] esbuild (contingent on https://github.com/mhart/ts-eager/issues/8)


HexaField:
Looking into other options. https://github.com/a7ul/esbuild-node-tsc/issues/23

HexaField:
https://github.com/folke/esbuild-runner/issues/32

PR #2449, opened 2021-07-15
Golf System uses ECS directly

HexaField:


HexaField:
closing in favour of different branch #2465

PR #2504, opened 2021-07-21
Fixes

Abhishek-Pathak-Here:


packages/server-core/src/hooks/add-thumbnail-file-id.ts
```diff
@@ -4,12 +4,12 @@ export default () => {
   return async (context: HookContext): Promise<HookContext> => {
     const fileIdentifier = context.params.body.fileIdentifier
     if (context.params.body.projectId && fileIdentifier) {
-      const { thumbnailOwnedFileId } = await (context.app.service('collection') as any).Model.findOne({
+      const { ownedFileIds } = await (context.app.service('collection') as any).Model.findOne({
```
HexaField, line 7: can we change this file to be `add-file-id.ts` instead of `add-thumbnail-file-id.ts` please

HexaField (changes requested):
just need to rename the file to make sense with it's new function, and test that the thumbnail now loads from the correct source

PR #2506, opened 2021-07-21
Fix Probe Editor

Abhishek-Pathak-Here:


packages/engine/src/editor/nodes/ReflectionProbeNode.ts
```diff
@@ -110,21 +110,18 @@ export default class ReflectionProbeNode extends EditorNodeMixin(Object3D) {
   }
 
   injectShader() {
-    const position = this.reflectionProbeSettings.probePosition.clone()
-    position.add(this.reflectionProbeSettings.probePositionOffset)
-    const scale = this.reflectionProbeSettings.probeScale
     this.editor.scene.traverse((child) => {
       if (child.material) {
         child.material.onBeforeCompile = function (shader) {
```
HexaField, line 115: let's use an arrow function instead of bind(this)

HexaField (approved):

HexaField (approved):

PR #2524, opened 2021-07-23
Refactor Engine Initialisation

HexaField:
In preparation of moving golf and mappa code into their own repositories.


speigg:
Awesome

PR #2537, opened 2021-07-24
Portal animation changes

NPatel10:


packages/engine/src/scene/behaviors/createPortal.ts
```diff
@@ -27,7 +27,7 @@ export const createPortal: Behavior = (entity, args) => {
   const { location, displayText, spawnPosition } = args
 
   const spawnRotation = new Quaternion().setFromEuler(
-    new Euler().setFromVector3(new Vector3(args.spawnRotation.x, args.spawnRotation.y, args.spawnRotation.z), 'XYZ')
+    new Euler(args.spawnRotation._x, args.spawnRotation._y, args.spawnRotation._z, 'XYZ')
```
HexaField, line 30: In initialiseEngine, we override Quaternion.prototype.toJSON to return x, y, z, w without the underscore, we should add the same override for Euler to keep things clean.

HexaField (changes requested):

NPatel10:
Portals are being transitioned smoothly. Still there are some glitches and I will work on them after other important tasks will be completed.

packages/engine/src/camera/systems/CameraSystem.ts
```diff
@@ -30,36 +30,58 @@ import { Object3DComponent } from '../../scene/components/Object3DComponent'
...
 
-const followCameraBehavior = (entity: Entity) => {
+/**
+ * Calculates and returns view vector for give angle. View vector will be at the given angle after the calculation
+ * @param viewVector Current view vector
+ * @param angle angle to which view vector will be rotated
+ * @param isDegree Whether the angle is in degree or radian
+ * @returns View vector having given angle in the world space
+ */
+export const getViewVectorFromAngle = (viewVector: Vector3, angle: number, isDegree?: boolean): Vector3 => {
```
HexaField, line 45: we don't want this to create a new Vector3 each frame, can we instead make this a function that changes the viewVector passed in directly?

packages/engine/src/editor/nodes/PortalNode.ts
```diff
@@ -29,9 +29,9 @@ export default class PortalNode extends EditorNodeMixin(Model) {
       node.spawnRotation = new Euler()
       if (portalComponent.props.spawnRotation)
         node.spawnRotation.set(
-          portalComponent.props.spawnRotation.x,
-          portalComponent.props.spawnRotation.y,
-          portalComponent.props.spawnRotation.z
+          portalComponent.props.spawnRotation._x,
```
HexaField, line 32: these need to have _ removed still

packages/engine/src/scene/functions/teleportToScene.ts
```diff
@@ -76,51 +86,39 @@ export const teleportToScene = async (portalComponent: PortalProps, handleNewSce
...
-  transform.rotation.copy(portalComponent.spawnRotation)
-
-  resetFollowCamera()
-
-  await delay(100)
 
-  hyperspaceEffect.position.copy(playerObj.value.position)
-  hyperspaceEffect.quaternion.copy(playerObj.value.quaternion)
+  const actor = getComponent(Network.instance.localClientEntity, CharacterComponent)
+  actor.viewVector.copy(getViewVectorFromAngle(actor.viewVector, (portalComponent.spawnEuler as any)._y))
```
HexaField, line 96: remove _ from ._y

HexaField (approved):

HexaField (changes requested):
must resolve conflict issue

packages/engine/src/camera/systems/CameraSystem.ts
NPatel10: Implemented the changed suggested

HexaField (approved):

PR #2576, opened 2021-07-28
Autopilot system

XiaNi:
Base for "Tap to move" with Navigation Mesh


packages/engine/src/map/parseGeometry.ts
```diff
@@ -0,0 +1,67 @@
+import { Polygon, Vector3 } from "yuka";
```
patreeceeo, line 1: Though this is really nice code, if I'm not mistaken, we're going to need to generate the navmesh dynamically from map data. Anyway, how is this different from the NavMeshLoader from Yuka?

XiaNi: this function is from that loader. i know that we need dynamic navmesh from map, but i needed graph to test.
it should be replaced

PR #2596, opened 2021-07-29
Adds static clouds mesh and corresponding editor nodes

HexaField:


packages/engine/src/scene/functions/SceneLoading.ts
```diff
@@ -275,6 +276,10 @@ export class WorldScene {
         createParticleEmitterObject(entity, component.data)
         break
 
+      case 'cloud':
+        addObject3DComponent(entity, { obj3d: Clouds, objArgs: component.data })
```
HexaField, line 280: this is outdated, please merge from dev. we no longer apply parameters in this way, it is all explicit. this way, we ensure typings are correct and there is no ambiguity about where values come from.

packages/engine/src/scene/classes/Clouds.ts
```diff
@@ -0,0 +1,210 @@
...
+  worldScale: Vector3
+  dimensions: Vector3 // Number of particles (x,y,z)
+  noiseZoom: Vector3
+  noiseOffset: Vector3
+  spriteScaleRange: Vector2
+  fogRange: Vector2
+  fogColor: Color
+  noise: SimplexNoise
+
+  constructor(texture: Texture, configs: any) {
```
HexaField, line 71: can this please be refactored, so arguments are not applied retrospectively? instead, i suggest create the material and geometry in the constructor, but then updating the uniforms in a `setTexture` function, that can then be called each time the texture needs to be updated, rather than reaching into the uniforms directly.

packages/engine/src/editor/nodes/CloudNode.ts
```diff
@@ -0,0 +1,143 @@
...
+    loadAsync(
+      (async () => {
+        await node.load(src, onError)
+      })()
+    )
+    node.updateParticles()
+
+    return node
+  }
+
```
HexaField, line 40: please add typings and defaults for the parameters

packages/engine/src/scene/classes/Clouds.ts
```diff
@@ -0,0 +1,210 @@
...
+    this.worldScale = new Vector3(1000, 150, 1000)
+    this.dimensions = new Vector3(8, 4, 8)
+    this.noiseZoom = new Vector3(7, 11, 7)
+    this.noiseOffset = new Vector3(0, 4000, 3137)
+    this.spriteScaleRange = new Vector2(50, 100) // Min/Max
+    this.fogColor = new Color(0x4584b4)
+    this.fogRange = new Vector2(-100, 3000)
+    this.noise = new SimplexNoise('seed')
+
+    if (configs) {
```
HexaField, line 107: this can be removed, all parameters should be set explicitly

packages/engine/src/scene/classes/Clouds.ts
```diff
@@ -0,0 +1,210 @@
...
+    this.noiseZoom.copy(source.noiseZoom)
+    this.noiseOffset.copy(source.noiseOffset)
+    this.spriteScaleRange.copy(source.spriteScaleRange)
+    this.fogColor.copy(source.fogColor)
+    this.fogRange.copy(source.fogRange)
+
+    return this
+  }
+
+  static fromArgs(args): Promise<Clouds> {
```
HexaField, line 200: please use the AssetLoader class for loading textures

mohsenheydari: Removed the function since it was not used

packages/engine/src/scene/classes/Clouds.ts
mohsenheydari: Moved the defaults to CloudNode.ts

PR #2617, opened 2021-08-02
Map 19: Add roads

patreeceeo:
Despite the name of the branch, this merely draws roads with THREE.ShapeGeometry using map vector data. Will add raster tile support in another PR.

image


patreeceeo:
also at some point the code for buildings/roads should probably be combined, with the minor differences being fully dictated by the style data

PR #2625, opened 2021-08-03
Add isometric camera

shawwwwwwww:


shawwwwwwww:
@hamzzam can you pull dev in and make sure this is up to date if it's finished?

packages/engine/src/camera/systems/CameraSystem.ts
```diff
@@ -51,6 +51,17 @@ const followCameraBehavior = (entity: Entity) => {
     FollowCameraComponent
   ) as FollowCameraComponent
 
+  if (CameraSystem.instance.updateCameraMode) {
+    let index = 0
+    for (const key of Object.keys(CameraModes)) {
+      if (index === CameraSystem.instance.cameraModeIndex && CameraSystem.instance.updateCameraMode) {
```
NPatel10, line 57: Why are we checking `CameraSystem.instance.updateCameraMode` here? We have already checked on the if condition above.

HexaField:
We should really change how we are handling the camera modes in the engine. I think instead of one follow camera component, we should have a separate component for each of the modes, and then also the FollowCameraComponent. This way, input behaviors just add and remove components, and all camera logic can be handled in the camera system. This also allows more modularity with how the camera system operates.

shawwwwwwww:
> We should really change how we are handling the camera modes in the engine. I think instead of one follow camera component, we should have a separate component for each of the modes, and then also the FollowCameraComponent. This way, input behaviors just add and remove components, and all camera logic can be handled in the camera system. This also allows more modularity with how the camera system operates.

I agreed, but we need to get this in for Mappa

PR #2675, opened 2021-08-06
isMobile replaced with appropriate code

NPatel10:


HexaField (changes requested):
I think instead of checking if the screen is larger or smaller than 768, we should use the device pixel ratio and the screen size relative to a standard size, let's say 1920px is 1, a screen with half the width would need a multiplier of 2 to make up the difference. The pixel ratio would  also be a multiplier.

packages/engine/src/character/CharacterInputSchema.ts
```diff
@@ -202,7 +201,7 @@ const changeCameraDistanceByDelta: InputBehaviorType = (
   const inputPrevValue = (inputComponent.prevData.get(inputKey)?.value as number) ?? 0
   const value = inputValue.value as number
 
-  const scrollDelta = Math.min(1, Math.max(-1, value - inputPrevValue)) * (isMobile ? 0.25 : 1)
+  const scrollDelta = Math.min(1, Math.max(-1, value - inputPrevValue)) * (window?.innerWidth <= 768 ? 0.25 : 1)
```
speigg, line 204: Scroll delta has nothing to do with window size. The feature we should be detecting here is likely related to smooth scrolling capability, or input source (e.g., scroll vs pinch gesture).

packages/engine/src/camera/systems/CameraSystem.ts
```diff
@@ -89,10 +86,10 @@ const followCameraBehavior = (entity: Entity) => {
   }
 
   if (followCamera.mode !== CameraModes.Isometric) {
-    followCamera.theta -= inputValue[0] * (isMobile ? 60 : 100)
+    followCamera.theta -= inputValue[0]
```
speigg, line 89: Is it safe to remove these multipliers?

packages/client/src/components/InstanceChat/index.tsx
```diff
@@ -151,6 +154,21 @@ const InstanceChat = (props: Props): any => {
...
+    }
+  }, [])
+
+  const handleWindowResize = () => {
+    setDimensions({
+      height: window.innerHeight,
+      width: window.innerWidth
+    })
+  }
+
```
speigg, line 171: Can we centralize this logic into a feature detection store, rather than repeating it multiple times? Look for where I am detecting whether or not touch events are available.

packages/engine/src/camera/systems/CameraSystem.ts
```diff
@@ -224,8 +221,8 @@ export class CameraSystem extends System {
       }
       addComponent(Engine.activeCameraEntity, DesiredTransformComponent, {
         lockRotationAxis: [false, true, false],
-        rotationRate: isMobile ? 5 : 3.5,
-        positionRate: isMobile ? 3.5 : 2
+        rotationRate: window?.innerWidth <= 768 ? 3.5 : 5,
+        positionRate: window?.innerWidth <= 768 ? 2 : 3.5
```
speigg, line 225: I think you incorrectly flipped the multipliers here; but also, we should be sure we understand the rationale for these multipliers; what is the basis for the different multipliers? Is it because of screen size, or it is because of differences in touch vs mouse input? This ambiguity (with “isMobile”) is one of the things we should really try to resolve here.

speigg (changes requested):

speigg:
> I think instead of checking if the screen is larger or smaller than 768, we should use the device pixel ratio and the screen size relative to a standard size, let's say 1920px is 1, a screen with half the width would need a multiplier of 2 to make up the difference. The pixel ratio would also be a multiplier.



> I think instead of checking if the screen is larger or smaller than 768, we should use the device pixel ratio and the screen size relative to a standard size, let's say 1920px is 1, a screen with half the width would need a multiplier of 2 to make up the difference. The pixel ratio would also be a multiplier.

I don’t think we want to consider pixel density here; CSS “pixel” units are effectively defined as visual angles based on typical viewing distances.

PR #2680, opened 2021-08-06
isMobile removed and suggestion implemented

NPatel10:


speigg:
Great! Thanks for making the fixes.

PR #2681, opened 2021-08-06
Refactor Golf Prefabs

HexaField:


speigg:
I love how much more typed everything is. The spawnPrefab function receives three different uuids. Do we need all three? If so, might be better to pass in named parameters.

PR #2709, opened 2021-08-09
Trigger volume node can effect volumetric and gltf animation

shawwwwwwww:
Still a work in progress. Rolling well on the editor side, needs work and testing on the front end.


shawwwwwwww:
@hamzzam how we doing with this? Could you make a PR with the code you have? Happy to help land it.

hamzzam:
Trigger functions of trigger volume component are now being called, but isTrigger of body of trigger volume component is set to false for now because avatar collision events were not working with trigger type bodies.

Triggering of model/video part is also remaining.

packages/engine/src/avatar/AvatarControllerSystem.ts
```diff
@@ -68,7 +69,34 @@ export const AvatarControllerSystem = async (): Promise<System> => {
...
 
       // iterate on all collisions since the last update
-      controller.controller.controllerCollisionEvents?.forEach((event: ControllerHitEvent) => {})
+      controller.controller.controllerCollisionEvents?.forEach((event: ControllerHitEvent) => {
+        let entity = event.body.userData.entity
+        let triggerComponent = getComponent(entity, TriggerVolumeComponent)
+        if (triggerComponent) {
+          if (!triggerComponent.active) {
+            triggerComponent.active = true
+            triggerComponent.onTriggerEnter()
```
HexaField, line 78: These triggers should be component add and remove queries. Integrating with the existing InteractiveSystem would be great.

HexaField (changes requested):

packages/engine/src/avatar/AvatarControllerSystem.ts
hamzzam: ok will do, can you point me to an example where it's being followed already. thanks

shawwwwwwww:
@hamzzam ready to pull?

HexaField:
> @hamzzam ready to pull?

Still waiting for the behaviors on the trigger components to be moved to a system instead.

hamzzam:
> > @hamzzam ready to pull?
> 
> Still waiting for the behaviors on the trigger components to be moved to a system instead.

Hey @HexaField , I have started working on those changes. But it will take me a little time to get that in. I am looking into how that flow works and has been implemented at other places and will start updating my code accordingly soon.

hamzzam:
@HexaField I have made all of the changes you asked for including the option to show hide helper mesh from the editor node.
Just the issue with physics isTrigger config, resulting in controller collision events not being registered needs looking into.

HexaField:
Looks like you merged dev incorrectly :( it is pulling in a bunch of previous changes. Can you please fix this?

PR #2735, opened 2021-08-10
Location.tsx refactor

NPatel10:


HexaField (commented):
Looks good so far! Just that top level await that may break on node that should be addressed.

PR #2746, opened 2021-08-11
change chat ui

kinjalravel:


shawwwwwwww:
@kinjalravel 
Can we hide this option list:
image

Give this a little more left margin:
image

Get rid of blue entirely?
image

shawwwwwwww:
@kinjalravel also can we make it start in open instead of closed position

PR #2763, opened 2021-08-12
Ocean shader

shawwwwwwww:


packages/engine/src/scene/systems/OceanSystem.ts
```diff
@@ -0,0 +1,34 @@
+import { getComponent } from '../../ecs/functions/EntityFunctions'
+import { ECSWorld } from '../../ecs/classes/World'
+import { defineQuery, defineSystem, System } from '../../ecs/bitecs'
+import { OceanComponent } from '../components/OceanComponent'
+import { isClient } from '../../common/functions/isClient'
+import { TransformComponent } from '../../transform/components/TransformComponent'
+
+export const applyTransform = (entity, ocean): void => {
```
HexaField, line 8: No need to apply transform, it is done via TransformSystem

packages/engine/src/scene/systems/OceanSystem.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!mesh) return
+
+  const transform = getComponent(entity, TransformComponent)
+  mesh.position.copy(transform.position)
+  mesh.quaternion.set(transform.rotation.x, transform.rotation.y, transform.rotation.z, transform.rotation.w)
+  mesh.scale.copy(transform.scale)
+  mesh.updateMatrix()
+}
+
+export const OceanSystem = async (): Promise<System> => {
```
HexaField, line 20: Rather than an ocean system by itself, we could make this generic by creating an Object3DUpdateTag and querying for it to then run the update function from the SceneObjectSystem.

HexaField (changes requested):
Fantastic work, just a few ECS changes and this is good to pull in.

PR #2764, opened 2021-08-12
File browser

shawwwwwwww:


shawwwwwwww:
@Abhishek-Pathak-Here ready for review?

Abhishek-Pathak-Here:
@shawwwwwwww Currently Still Working.
Some Database stuff is required as of now.

packages/client-core/src/world/components/editor/assets/FileBrowserContentPanel.tsx
```diff
@@ -0,0 +1,195 @@
...
+import { EditorContext } from '../contexts/EditorContext'
+import AssetDropZone from './AssetDropZone'
+// @ts-ignore
+import styles from './styles.module.scss'
+import { useAssetSearch } from './useAssetSearch'
+import { AssetPanelContentContainer } from './AssetsPanel'
+import AssetGrid from './AssetGrid'
+import SelectInput from '../inputs/SelectInput'
+import InputGroup from '../inputs/InputGroup'
+import ModelNode from '../../../../../../engine/src/editor/nodes/ModelNode'
```
HexaField, line 14: please use @xrengine instead of '../../../../../../'

packages/engine/src/editor/nodes/ReflectionProbeNode.ts
```diff
@@ -14,6 +10,8 @@ import {
   Vector3,
   WebGLCubeRenderTarget
 } from 'three'
+import { FilesToUpload } from '../../../../client-core/src/world/components/editor/Api'
```
HexaField, line 13: we must keep the package hierarchy in fact, the engine cannot depend on client-core, only client-core on the engine, so i think these interfaces should be moved to the engine. we will be overhauling the editor next month anyway, so this will all change again, but for now we must preserve this so npm distribution works.

HexaField (changes requested):

packages/server-core/src/hooks/create-owned-file.ts
```diff
@@ -16,6 +16,7 @@ export default () => {
         : config.server.localStorageProvider
 
     let savedFile
+    console.log('ProjectID is:' + body.projectId + ' previousFileId is:' + context.params.previousFileId)
```
NPatel10, line 19: I this comment neccessary

packages/server-core/src/hooks/add-file-id.ts
```diff
@@ -3,6 +3,7 @@ import { HookContext } from '@feathersjs/feathers'
 export default () => {
   return async (context: HookContext): Promise<HookContext> => {
     const fileIdentifier = context.params.body.fileIdentifier
+    console.log('FileIdentifier is:' + fileIdentifier)
```
NPatel10, line 6: Is this log is necessary

Abhishek-Pathak-Here: No,will remove it

Issue #1392, opened 2021-03-19
Avatars materialize and dematerialize on load/unload


shawwwwwwww:
We want a materialization / dematerialization shader here. Typically this would be cutout dissolve effect with a noise pattern.

The creator of the ECS we borrowed ours from has a system for dissolving already:
https://github.com/fernandojsg/beatpads/blob/791034b0669a5c7a2cf9952c3f7e033d39855fab/src/Systems/DissolveSystem.js

https://forum.unity.com/threads/adding-edge-glow-feature-to-dissolve-shader.210153/

![image](https://user-images.githubusercontent.com/18633264/128136464-34900d54-a292-4419-8b73-5baf911e729e.png)

PR #2768, opened 2021-08-12
Dissolve Effect (#1392)

sorincinculei:


packages/engine/src/avatar/functions/avatarFunctions.ts
```diff
@@ -67,9 +79,121 @@ const loadAvatarFromURL = (entity: Entity, avatarURL: string) => {
...
       avatar.modelContainer.children.forEach((child) => child.removeFromParent())
 
+      const textureNoise = await AssetLoader.loadAsync({ url: '/noise.jpg' })
+
+      let materialList = []
+
       model.traverse((object) => {
         if (typeof object.material !== 'undefined') {
-          object.material = object.material.clone()
+          // object.material = object.material.clone()
```
HexaField, line 88: this should be moved to a separate function in another file

packages/engine/src/avatar/functions/avatarFunctions.ts
```diff
@@ -82,6 +206,47 @@ const loadAvatarFromURL = (entity: Entity, avatarURL: string) => {
 
       // advance animation for a frame to eliminate potential t-pose
       animationComponent.mixer.update(1 / 60)
+
+      const textureLight = await AssetLoader.loadAsync({ url: '/itemLight.png' })
```
HexaField, line 210: this can be moved to it's own function also

HexaField (changes requested):
Just needs bulky functions moved to their own files :) Looks very good! Great work.

HexaField (approved):

PR #2769, opened 2021-08-12
XRE 296 Integration of create new scope 

kimenyikevin:


barankyle (changes requested):
scope-type, and some of the front- and back-end that deals with it, needs some changes to match the intended vision. My wording on the Jira ticket was unintentionally confusing and vague on this subject.

The scope-type table should only have one column, 'type', not the several different columns as has been implemented currently. Each row, which needs to be seeded (and thus requires creation of scope-type.seed.ts), will be one of the allowable scopes listed in the Jira ticket (location:read, scene:write, etc.) user-relationship-type is an example of how scope-type should be implemented.

Each scope that's created will be for exactly one scope-type. For example, you'd create a scope so that user 'abc123' has static_resource:write access. If you wanted to give them bot:read access, that would be done as a separate action, creating a separate scope.

In terms of ease of use, I think the scope page might be better structured as showing every user or every group at once (with buttons to change whether you're viewing users or groups), and you click on a user/group to see what scopes they have and add or remove scopes. It would probably be clearest for this scope select modal to use Material-UI's multiple select - all scope-types are options in the multi-select, and the front-end sends the selected ones to the API as an array when the user clicks save. This should probably be a PATCH, and the back-end handler for scope.patch removes existing scopes for that user/group that are not in the array and makes scopes that don't currently exist for that user/group.

So, for example, a user's scopes are being updated to be location:read and bot:write. Currently, their scopes are location:read, location:write, and globalAvatars:read. scope.patch first fetches all the user's current scopes, then iterates through the new list. It deletes the user's location:write and globalAvatars:read scopes since they're no longer being authorized for those, leaves their location:read scope as-is since it exists and is not being revoked, and creates a bot:write scope since it doesn't currently exist.

shawwwwwwww:
@barankyle @kimenyikevin how we doing with this? Ready for review again?

kimenyikevin:
> @barankyle @kimenyikevin how we doing with this? Ready for review again?

@shawwwwwwww  Yes, it is ready for review. I am waiting for @barankyle review

barankyle (approved):

PR #2788, opened 2021-08-13
Revert location and avatar works again

shawwwwwwww:
This reverts a bunch of changes made in the last couple of days to merge different bits together. This gets the avatars working again, but ideally we can look at the diff and bring the important stuff back.

### DIFF
--- a/packages/client/src/components/World/index.tsx
+++ b/packages/client/src/components/World/index.tsx
@@ -32,29 +32,27 @@ import { SocketWebRTCClientTransport } from '../../transports/SocketWebRTCClient
 
 const engineRendererCanvasId = 'engine-renderer-canvas'
 
-const getDefaulEngineInitializeOptions = (): InitializeOptions => {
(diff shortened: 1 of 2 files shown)


packages/client/src/components/World/index.tsx
```diff
@@ -32,29 +32,27 @@ import { SocketWebRTCClientTransport } from '../../transports/SocketWebRTCClient
...
-    },
-    systems: [
-      {
-        type: SystemUpdateType.Fixed,
-        system: GolfSystem,
-        after: GameManagerSystem
-      }
-    ]
-  }
+const defaulEngineInitializeOptions: InitializeOptions = {
```
NPatel10, line 35: I removed this because most of the places some change in the initialization option is required, like in golf or in map. We are also setting default options in the `initializeEngine` as well. It is your call now.

NPatel10 (approved):

PR #2796, opened 2021-08-13
Removed disabled/delete options from static-resource.seed.ts

barankyle:
When local configuration of disabled/delete was removed from .seed files in favor of top-level
one, static-resource.seed.ts erroneously had its left in, which resulted in the seeding when no
user table was detected not running for this since config.db.forceRefresh wasn't true.


speigg:
Thanks!

PR #2832, opened 2021-08-17
Implemented the Metadata Node

alextitonis:


shawwwwwwww (approved):

packages/client-core/src/world/components/editor/properties/MetadataNodeEditor.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { Component } from 'react'
+import { Link } from '@styled-icons/fa-solid/Link'
```
shawwwwwwww, line 2: Replace with a different styled icon from 
https://styled-icons.js.org/

shawwwwwwww (changes requested):

shawwwwwwww (approved):

PR #2843, opened 2021-08-17
MAP-update-tiles/mapSystem

aleksmaz:


patreeceeo:
So what does this do? Just tested it, this is what I'm seeing (with useCurrentLocation turned on):

Peek 2021-08-03 07-52

Is that what you expected?

shawwwwwwww:
For me (Chromium Linux) crashes every time I walk to edge of boundary.
Tiles should load so you're always in the center tile of a 9x9
[  ][  ][  ]
[  ][x][  ]
[  ][  ][  ]

aleksmaz:
@shawwwwwwww 
Problems that remain
-long update caused by generateNavMesh
- during the start of the update, the autopilot continues to work

If you have time to look at the code - I will describe some points
-in the mapUpdateSystem, we translate the coordinates of our character into longitude and latitude
- having these coordinates, we determine which tile we are in
-Then we compare the center tile and the tile we are in. If they match, we are in the center
-If not, we update the map. We assign the current coordinates to the center. We build again 9 tiles. And we give this whole mesh our avatar position.
-In general, the idea is that we are always in the central tile
-Also in the update function - I tried updating the component values. But for some reason it adds all the same tiles

packages/engine/src/scene/functions/createMap.ts
```diff
@@ -19,4 +21,20 @@ export async function createMap(entity: Entity, args: MapProps): Promise<void> {
...
+
+export async function updateMap(args: MapProps, longtitude, latitude, position): Promise<void> {
+  const remobj = Engine.scene.getObjectByName('MapObject')
+  const { mapMesh, navMesh, groundMesh } = await update(Engine.renderer, args, longtitude, latitude, position)
+
+  remobj.removeFromParent()
+  Engine.scene.add(mapMesh)
+
+  // getComponent(currentEnt, Object3DComponent).value.clear()
+  // getComponent(currentEnt, Object3DComponent).value.add(mapMesh)
```
patreeceeo, line 35: Tried reassigning to `component.value` i.e. `getComponent(currentEnt, Object3DComponent).value = mapMesh`?

aleksmaz: tried it. not works

PR #2855, opened 2021-08-18
Profile Screen UI Design

kinjalravel:


shawwwwwwww:
@kinjalravel looks good, but conflicts with dev, please just merge really quickly when you can

PR #2858, opened 2021-08-18
handle issue when autopilot.path.current() is undefined

XiaNi:
show warning, remove autopilot and skip to next if autopilot.path.current() is undefined


patreeceeo (commented):
Not sure I agree with this. If we're not sure why it gets in this state I think we should allow the error to happen until we understand it better, as long as it doesn't effect end-user experience.

patreeceeo:
If the error is something that players would notice, perhaps let's just change the `console.warn` to `console.error`?

PR #2879, opened 2021-08-19
Fixed-animation state

thanhhovn2016:


shawwwwwwww:
@thanhhovn2016 please review. I believe that this fixes the judder issue on mobile. But it breaks movement on PC completely, with WASD. So I think what you want is to disable the movement in the other place it is happening on mobile and undo some of these changes so that WASD still works.

thanhhovn2016:
Yes,I'm fixing.19.08.2021, 14:07, "shaw" ***@***.***>:
@thanhhovn2016 please review. I believe that this fixes the judder issue on mobile. But it breaks movement on PC completely, with WASD. So I think what you want is to disable the movement in the other place it is happening on mobile and undo some of these changes so that WASD still works.

PR #2888, opened 2021-08-20
Asset Loader Improvements

HexaField:
- [x] Asset loader is functional
- [x] Asset loader asset processing only happens on first load
- [x] Make more asset calls re-cached and synchronous
- [ ] Cache assets to browser storage

Solves https://github.com/XRFoundation/XREngine/issues/2833


HexaField:
Browser caching likely can be solved for with Cloudfront options, so I will leave the issue open but merge this PR

PR #2914, opened 2021-08-23
Improve Portals

HexaField:
- [x] preview mesh, frame and trigger can be controlled independently
- [x] improve stability and feel


NPatel10 (approved):
Every thing looks good to me

PR #2939, opened 2021-08-24
Golf test 2players

HydraFire:


packages/client/src/pages/golf/functions/golfBotHookFunctions.ts
```diff
@@ -5,26 +5,28 @@ import { Network } from '@xrengine/engine/src/networking/classes/Network'
...
   [GolfBotHooks.GetIsPlayerTurn]: getIsPlayerTurn,
-  [GolfBotHooks.GetIsGoal]: getIsGoal,
+  [GolfBotHooks.GetPlayerPosition]: getPlayerPosition,
+  [GolfBotHooks.GetPlayerScore]: getPlayerScore,
   [GolfBotHooks.GetIsBallStopped]: getIsBallStopped,
   [GolfBotHooks.GetIsOutOfCourse]: getIsOutOfCourse,
   [GolfBotHooks.SwingClub]: swingClub
 }
 
 export function getIsPlayerTurn() {
```
HexaField, line 25: this should be `getCurrentPlayerNumber(): number`

packages/client/src/pages/golf/functions/golfBotHookFunctions.ts
```diff
@@ -83,3 +92,8 @@ export function getBallPosition() {
   const ballTransform = getComponent(ballEntity, TransformComponent)
   return ballTransform.position
 }
+
+export function getPlayerScore() {
```
HexaField, line 96: this should take an player number and return the score

HexaField (changes requested):

PR #3023, opened 2021-08-28
Add scope in before hooks

kimenyikevin:


barankyle (changes requested):
The 'Scene' admin page selector in the sidebar needs to be not rendered if the user doesn't have the 'scene:read' scope.  The top-level 'Scene' accordion should not be rendered if none of its sub-selectors are selectable.

The top-level 'Location' accordion selector also needs to be turned off if none of its child selectors are rendered.

The various admin pages need to not be hittable without the corresponding read scope. Right now I can go to /admin/groups and see groups even without 'group:read'; not having it be selectable from the sidebar is not enough.

All /editor pages need to not be hittable if the user doesn't have 'editor:write' scope.

contentPacks, group, instance, invite, and party need write scopes as well as read scopes, and those write scopes need to be protecting the create/update/patch/remove functions on their respective services

barankyle:
We want to have default scopes that are granted to users based on their userRole. This will require some minor changes to user creation/patching.

There will be a couple of new environment variables in .env.local.default, `DEFAULT_GUEST_SCOPES` and `DEFAULT_USER_SCOPES`. They'll be comma-separated strings, e.g. `user:read,user:write,editor:write`. 

appconfig.ts will have a new sub-field on its export, `scopes`, that will specify which scopes certain userRoles will get when they're created/patched. They should be sourced from the respective environment variable and split on the commas to produce an array, or default to an empty array if the environment variable is null or an empty string. Here's an example of what it should look like, with guest roughly how it'll be pulled from the ENV_VAR and user an example of what the array should look like:

```
scopes: {
    guest: DEFAULT_GUEST_SCOPES?.split(',') || [],
    user: [
        'editor:write',
        'location:read',
        'location:write'
    ]
}
```

When a new user is created in identity-provider.class.ts:create(), a new scope for all of the scopes in config.scopes.guest should be made for that user. user.class.ts:patch needs to be added and, if their userRole is being patched to `user`, all of the scopes in config.scopes.user should be made for them. If they're being patched to an admin, give them every scope that's in the scopeType table. If they're being downgraded from an admin to a user, remove all of the scopes they have except the ones in config.scopes.user.

barankyle:
I can't see or select the location list from /editor/projects.

Need to add scopes `globalAvatars:write`, `groups:write`, `instance:write`, `party:write`, `user:write`, and `contentPacks:write`.

barankyle (approved):

PR #3049, opened 2021-08-31
Update: changing spawn-point models

OmarVector:
changing the default spawn point model to new model


OmarVector:
I closed, I did a mistake by pushing using gitkraken without waiting for the formatter

PR #3096, opened 2021-09-03
command handlers, proximity system, follow system

alextitonis:


packages/client-core/src/social/reducers/chat/actions.ts
```diff
@@ -22,6 +22,12 @@ import { handleCommand, isCommand } from '../../../../../common/src/utils/comman
 import { Network } from '../../../../../engine/src/networking/classes/Network'
```
HexaField, line 22: cross package imports should use @xrengine/package-name instead of relative path

packages/common/src/utils/commandHandler.ts
```diff
@@ -1,11 +1,26 @@
-import { getComponent, addComponent } from '../../../engine/src/ecs/functions/EntityFunctions'
+import { getComponent, addComponent, hasComponent } from '../../../engine/src/ecs/functions/EntityFunctions'
```
HexaField, line 1: common folder cannot import from engine, as engine depends on common. please move all of these to the engine.

packages/common/src/utils/commandHandler.ts
```diff
@@ -84,46 +100,370 @@ export function handleCommand(cmd: string, eid: any, isServer: boolean): boolean
...
-    console.log('scene_metadata|' + WorldScene.sceneMetadata)
+    console.log('scene_metadata|' + World.sceneMetadata)
   } else {
     const position = getComponent(eid, TransformComponent).position
     const maxDistance: number = parseFloat(params[1])
-    var _data
-    var x,
-      y,
-      z: number = 0
     var vector: Vector3
```
HexaField, line 253: please use let instead of var

packages/engine/src/networking/utils/chatSystem.ts
```diff
@@ -0,0 +1,97 @@
+import { sendChatMessage } from '../../../../client-core/src/social/reducers/chat/service'
```
HexaField, line 1: engine cannot depend on client-core, these must be moved to client-core

packages/engine/src/proximityChecker/components/ProximityCheckerComponent.ts
```diff
@@ -0,0 +1,11 @@
+import { createMappedComponent, hasComponent, addComponent, getComponent } from '../../ecs/functions/EntityFunctions'
+
+export type ProximityCheckerComponentType = {}
+
+export const createProximityCheckerComponent = (eid: number) => {
```
HexaField, line 5: checking for a component should be done where this function is called, instead of in a function like this to improve readability of ecs logic

packages/engine/src/proximityChecker/systems/ProximityCheckerSystem.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Network } from '../../networking/classes/Network'
+import { TransformComponent } from '../../transform/components/TransformComponent'
+import { ProximityCheckerComponent } from '../components/ProximityCheckerComponent'
+import { forwardVector3, multiplyQuaternion, normalize, subVector } from '../../../../common/src/utils/mathUtils'
+import { isEntityLocal } from '../../networking/utils/isPlayerLocal'
+import { getUserId } from '../../networking/utils/getUser'
+
+const maxDistance: number = 10
+
+export const ProximityCheckerSystem = async (): Promise<System> => {
```
HexaField, line 13: should be named `ProximitySystem`

HexaField (changes requested):

packages/engine/src/navigation/systems/FollowSystem.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const distanceToPlayer: number = 1
+const step: number = 10
+
+export const FollowSystem = async (): Promise<System> => {
+  const followQuery = defineQuery([FollowComponent])
+
+  return defineSystem((world: ECSWorld) => {
+    for (const eid of followQuery(world)) {
+      var { targetEid, cStep, prevTarget } = getComponent(eid, FollowComponent)
```
HexaField, line 18: let is preferred over var

packages/common/src/utils/mathUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export const positionBehind = (position: Vector3, rotation: Quaternion, distance: number): Vector3 => {
+  return subVector(position, multiplyVector(getForward(rotation), distance))
+}
+
+export const lookAt = (position: Vector3, rotation: Quaternion, targetPos: Vector3): number => {
+  const x = targetPos.x - position.x
+  const y = targetPos.z - position.z
+  var angle = Math.atan2(x, y)
```
HexaField, line 59: let is preferred over var

HexaField (changes requested):

PR #3226, opened 2021-09-10
Revert "Update dependency idb-keyval to v6"

HexaField:
Reverts XRFoundation/XREngine#3203


HexaField:
This update appears to break jest

PR #3244, opened 2021-09-11
ts-pattern changes

AbhinavShah421:


speigg:
@AbhinavShah421 

1) We should match on the entire `action`, not just `action.type`
2) Reducers need to return the next state, so we should be returning the results from match in those cases

AbhinavShah421:
ok I'll change it

PR #3251, opened 2021-09-13
initial commit for velocity sync of objects with client authority

alextitonis:


packages/common/src/utils/mathUtils.ts
```diff
@@ -68,3 +68,7 @@ export const rotate = (rot: Quaternion, x: number, y: number, z: number) => {
 export const eulerToQuaternion = (x, y, z, order = 'XYZ') => {
   return new Quaternion().setFromEuler(new Euler(x, y, z, order))
 }
+
+export const isZero = (v: Vector3): boolean => {
```
HexaField, line 72: replace with Vector3.isEqual against an empty vector3

packages/engine/src/networking/schema/networkSchema.ts
```diff
@@ -43,6 +52,10 @@ export interface WorldStateInterface {
     leftPose: Pose
     rightPose: Pose
   }[]
+  velocities: {
```
HexaField, line 55: velocities should be in the same object as the pose for the same networkId, no need to send networkIds twice and is clearer to have all of an object's information together

packages/engine/src/networking/systems/IncomingNetworkSystem.ts
```diff
@@ -117,8 +131,18 @@ export default async function IncomingNetworkSystem(world: World): Promise<Syste
...
           xrInputSourceComponent.controllerRight.quaternion.fromArray(rightPose, 3)
         }
+
+        for (const vel of newWorldState.velocities) {
+          if (!Network.instance.networkObjects[vel.networkId]) continue
+          const entity = Network.instance.networkObjects[vel.networkId].entity
+
+          if (hasComponent(entity, VelocityComponent)) {
+            const velC = getComponent(entity, VelocityComponent)
+            velC.velocity.fromArray(vel.velocity === 0 ? [0, 0, 0] : vel.velocity)
```
HexaField, line 141: why would velocity ever be 0 here? best to not create an empty object here too, and instead use setScalar(0) to avoid unnecessary GC

HexaField (changes requested):

packages/common/src/utils/mathUtils.ts
speigg: @HexaField wasn't this better the way it was before? now a new vector3 is being made every time this is called.

packages/common/src/utils/mathUtils.ts
```diff
@@ -68,3 +68,7 @@ export const rotate = (rot: Quaternion, x: number, y: number, z: number) => {
...
+  return v.equals(new Vector3())
+}
```
NateTheGreatt, line 74:
```suggestion
export const isZero = (v: Vector3): boolean => {
  return v.x === 0 && v.y === 0 && v.z === 0
}
```

HexaField (changes requested):
one last change :)

PR #3277, opened 2021-09-14
Gallery-add-func

aleksmaz:
-add remove post
-add all/featured lists
-add single View
-clean-up  unnecessary stuff


shawwwwwwww:
@aleksmaz looks good but can you lint it so it passes tests? Then I can pull in

HexaField:
The check that causes CI to fail here has since been removed so this should be considered good to merge.

PR #3285, opened 2021-09-15
Implement serve configs from server and make editable in admin panel 

kimenyikevin:
Related [issue](https://github.com/XRFoundation/XREngine/issues/2704)


barankyle:
Couple things I noticed after reviewing:

server.url seed needs to use the logic from appconfig.ts lines 84-85 (can condense it into one line by putting the `obj` assignment into url.format)

authentication.facebook.callback is not displayed on the settings page (but it does appear to be saved in the database)

Anything labeled 'secret key' should be a password display that can be toggled to plaintext, like what's currently in place for email.auth.password.

All of the configuration models are associated with the user table, but this should be removed since no user owns the configuration.


The Authentication Strategies section needs to be reworked a bit; I hadn't paid enough attention to this when I looked at it previously. 

Feathers authentication works like this: you specify the names of strategies you want to use in authentication.authStrategies. When feathers starts, it tries to register everything named in authentication.authStrategies, using authentication.callback.<strategy_name> and authentication.oauth.<strategy_name> (or authentication.local for local auth), as well as a custom strategy definition we register in server-core/src/user/authentication.ts. 

The current display, where there's a single `url` field for each strategy, doesn't really work, and doesn't convey the right information. I think it would make more sense for config.authStrategies to be the baseline set of strategies that can be used, and each one can be toggled on or off under its respective place on the Authentication page (apart from 'jwt', which will always be enabled). 

The way authStrategies is saved in the authentication table would be that, instead of just an array of string, it's an array of objects where each one is { <name>: <is_enabled> }, e.g. [{ 'jwt': true}, { 'local': false }, { 'facebook': true }, { 'github': true }]. The seed would set all of them to true by default. If an admin goes to auth settings and turns a toggle in the Facebook section off, then 'facebook' would be set to false, e.g. [{ 'jwt': true}, { 'local': false }, { 'facebook': false }, { 'github': true }]. A strategy's presence  in the authStrategies array determines whether or not to render an element containing its on/off toggle, inputs for keys, and callback URL; the true/false value is just for determining if the API server uses that strategy.

shawwwwwwww:
@kimenyikevin just wanted to make sure you saw this comment

PR #3288, opened 2021-09-15
Camera improvements

mohsenheydari:
- Added zoom transition when the line of sight is broken
- Replaced physics raycasting with Raycaster object
- All objects explicitly set to a layer
- Avatar mesh layer assigned on load


packages/engine/src/common/classes/GenerateMeshBVHWorker.ts
```diff
@@ -0,0 +1,94 @@
...
+      this.worker = new Worker('/workers/generateBVHAsync.worker.js')
+    } else {
+      this._serverWorker()
+    }
+  }
+
+  async _serverWorker() {
+    const path = await require('path')
+    // TODO: Is there a better way to resolve this path?
+    const workerPath = path.resolve(__dirname, '../../../../client/public/workers/generateBVHAsync.worker.js')
```
HexaField, line 24: rather than referring to the client dist, we should move this worker file into the engine, as hopefully soon our build chain will allow vite to import workers from within the engine too. so for now, node should reference the worker file in the same folder as this file, and the browser is fine as is for now.

HexaField (changes requested):

PR #3294, opened 2021-09-15
Admin dashboard responsiveness 

kimenyikevin:


barankyle:
The analytics dashboard has some issues, at least when I go to the page with Chrome's developer tools open and simulating a mobile view. Some elements aren't sized/flexed right, leading to scrollable whitespace. If you scroll to the bottom of the page, there's a white gap at the bottom, possibly caused by some parts of the graph.

Screenshot from 2021-09-20 15-54-16

The cards showing Active Parties/Active Users/etc. are being rendered on top of each other instead of being shrunk down, leading to text being blocked. This is also causing those cards to extend off the right side of the page, leaving you able to scroll into a bunch of whitespace. The font size of the statistics numbers could be shrunk a lot to make things fit better, possibly other font sizes, too.

Screenshot from 2021-09-20 15-56-54

Additionally, almost all of the other admin pages, except for `bot`, don't appear to be responsively adapting to screen size. The tables are much wider than the width of the screen and require a lot of scrolling.

Screenshot from 2021-09-20 16-05-13

PR #3300, opened 2021-09-16
optimized the quaternions to use 3 values and not 4 on sync

alextitonis:


packages/engine/src/networking/schema/networkSchema.ts
```diff
@@ -42,16 +45,19 @@ export interface WorldStateInterface {
   pose: {
     networkId: number
     position: number[]
-    rotation: number[]
+    rotation: Pose
```
HexaField, line 48: Pose here should be number[]

HexaField (changes requested):

PR #3358, opened 2021-09-20
Camera update

mohsenheydari:
- Added cone tracing to better predict obstacles
- Fixed a BVH generation bug that resulted in no ray hit detection
- BVH generation should now perform better as no copy operation is involved (see Transferable objects on MDN)


packages/engine/src/common/classes/GenerateMeshBVHWorker.ts
```diff
@@ -41,6 +41,11 @@ export class GenerateMeshBVHWorker {
         if (error) {
           reject(new Error(error))
         } else {
+          // we need to replace the arrays because they're neutered entirely by the
+          // webworker transfer.
+          geometry.setAttribute('position', new BufferAttribute(position, 3))
```
HexaField, line 46: i dont think we want to replace the geometry data when it comes back, this can cause problems with physics and other things that rely on the geometry. what is the reason it is needed to be returned back?

packages/engine/src/common/classes/GenerateMeshBVHWorker.ts
```diff
@@ -62,33 +62,30 @@ export class GenerateMeshBVHWorker {
...
-          'GenerateMeshBVHWorker: InterleavedBufferAttribute are not supported for the geometry attributes.',
-          geometry
+        throw new Error(
+          'GenerateMeshBVHWorker: InterleavedBufferAttribute are not supported for the geometry attributes.'
         )
-        return
       }
 
-      const transferrables = [position.buffer, index.buffer]
+      const index = geometry.index ? geometry.index.array : null
```
HexaField, line 74: transferring the original copy and having this async process will lead to the geometry data being undefined for a short period of time, which will cause errors in other areas.  This is the reason i did a copy of them.

HexaField (changes requested):

packages/engine/src/common/classes/GenerateMeshBVHWorker.ts
mohsenheydari: Yes, I'm aware of this problem. But the previous implementation had two issues:

- The raycasting did not work after the BVH generation is finished
- The buffer copy operation made generation slower and took more memory because GC does not kick in instantly after reference invalidation and when it does, it introduces jitter.

PR #3428, opened 2021-09-24
Nitin emote respawn tasks

Developmentexpert121:
Style emote menu and implement respawn button.


HexaField (approved):
Code looks good! Will need a QA.

PR #3465, opened 2021-09-27
implemented Hookstate in Instance Location Party RealityPack Scene

kinjalravel:


barankyle:
There are some more files that need to be converted, at the very least `packages/editor/src/components/EditorContainer.tsx` and `packages/editor/src/components/toolbar/PublishModel.tsx`. There may be others; those are just ones that threw errors when I initially tried to run the client.

Also, last week I'd restored several uses of getState() in service functions because trying to use use<Service>State in a .ts file was throwing errors about hooks only being usable in function components. Unless something has changed about the implementation in this PR, that will still be a problem, and getState() should be used again

kinjalravel:
> There are some more files that need to be converted, at the very least `packages/editor/src/components/EditorContainer.tsx` and `packages/editor/src/components/toolbar/PublishModel.tsx`. There may be others; those are just ones that threw errors when I initially tried to run the client.
> 
> Also, last week I'd restored several uses of getState() in service functions because trying to use useState in a .ts file was throwing errors about hooks only being usable in function components. Unless something has changed about the implementation in this PR, that will still be a problem, and getState() should be used again

@barankyle Thanks for suggestions , well actually we are going module by module because if we try to change whole thing at once then everything will break and it will be very messy to fix and all , also other devs are working on multiple parts so it will be very hard to merge all those things if we do all at once 

Eventually everything will be converted but it will be happen module by module

PR #3493, opened 2021-09-29
refactor analytics on admin system 

kimenyikevin:


barankyle:
I fixed some issues with data not rendering properly. Couple final things:

1) The activity graph isn't rendering any usable data the first time around, and it's not updating as new data come in. If I click the Users tab, then back to the Activity tab, it does render. I think the graph just needs to be forced to re-render whenever the data changes.

2) On mobile, the graph needs to scale better. I think the height is fixed at all times, and it needs to change based on screen size.
Screenshot from 2021-09-29 13-19-49

kimenyikevin:
@barankyle check this PR again

barankyle:
I still often see the activities graph only render Active Parties on its initial render. And the graph isn't updating on each new fetch of data. Even when I click to Users and then back to Activities, it renders all the data sets when it refetches right then, but if I, for instance, go to the test location in a separate tab, I don't see Active Instances or Active Locations tick up.

It looks like the axis scaling I'd written for the graphs got lost. See [this commit ](https://github.com/XRFoundation/XREngine/commit/162990f3a0cadbf1fc864e5e20b8beab70d71e6d), particularly at the top of both where I'd calculated maxY and the changes I'd made to xaxis.min and xaxis.max.

PR #3516, opened 2021-09-30
Network Refactors/Optimizations & Node Inspector

NateTheGreatt:
`packages/engine/src/networking/systems/IncomingNetworkSystem.ts`
`packages/engine/src/networking/systems/OutgoingNetworkSystem.ts`
- refactored into pipeline for readability, easier dependency tracking, and easier unit & integration testing
- GC optimizations

`packages/common/src/utils/bitConverter.ts`
- GC optimizations

`packages/gameserver/package.json`
- added `--inspect` flag to `npm run dev` script to run the node inspector

`packages/ops/docker-compose-local.yml`
- exposed port `9229` on the `gameserver` docker instance for node inspector

the node inspector can be accessed from chrome dev tools by clicking the icon shown here:
image


packages/engine/src/physics/systems/PhysicsSystem.ts
```diff
@@ -102,8 +102,6 @@ export default async function PhysicsSystem(
       const transform = getComponent(entity, TransformComponent)
       const network = getComponent(entity, NetworkObjectComponent)
 
-      if ((!isClient && network.userId === Engine.userId) || hasComponent(entity, AvatarComponent)) continue
```
NateTheGreatt, line 105: @speigg before we merge i want to make sure this wasn't doing anything specific i'm unaware of, but removing this line fixed the networked physics sync issue

speigg: I don't remember why this check is here to be honest. Feel free to refactor as much as you think is necessary for proper state sync implementation!

PR #3525, opened 2021-10-01
Removed reducers from gallery, using from social instead.

Sergey-Zaharov:


HexaField (changes requested):
We do not want to be importing across packages with relative urls. The gallery acts as it's own client, as do the client and social packages. It would be best to move as many things from these packages into client-core so they can be reused.

Related issues:
https://github.com/XRFoundation/XREngine/issues/3533
https://github.com/XRFoundation/XREngine/issues/3362

shawwwwwwww:
@Sergey-Zaharov please review, we need gallery to be self-contained. We'll be abstracting to a reality pack soon also

HexaField:
Actually on second thought, if the intention is to keep the social package as a template or set of utils, we should either move everything from social to client-core, or make clearer the boundaries of these packages. There's a few ways we could go about this.

PR #3548, opened 2021-10-03
Add GLTFInstancingExtension to GLTF Loader to support gltf based inst…

hamzzam:
…anced meshes.


hamzzam:
To export supported glbs we can simply use instanced meshes in blender and then run `https://gltf-transform.donmccurdy.com/functions.html#instance` to convert the glb to supported format.

HexaField:
> To export supported glbs we can simply use instanced meshes in blender and then run `https://gltf-transform.donmccurdy.com/functions.html#instance` to convert the glb to supported format.

Could we somehow get what this website does into the editor?

HexaField:
> To export supported glbs we can simply use instanced meshes in blender and then run `https://gltf-transform.donmccurdy.com/functions.html#instance` to convert the glb to supported format.

Can we get this into our editor? We should try to eliminate any middlemen in the asset pipeline

PR #3869, opened 2021-10-27
set isUserBanned default to false

XiaNi:
It's a quick fix, i didn't research side effects of this. isUserBanned is passing to next components deeper.

by default isUserBanned is set to true, that leads to flashing this message if you are not banned while we retrieving actual state.
in my opinion we should or set it to false or set it to "unknown" and show "please wait"/"loading" while we checking it out.


shawwwwwwww:
@barankyle what do you think?

barankyle:
Fine with me.

PR #3876, opened 2021-10-27
User Inventory changes

swpnit82:
Please review and merge user inventory changes


kinjalravel:
@swpnit82  For user and inventory item association it should need to return actual items details not it's association ids, I will send updated API

HexaField:
Let's get some basic tests with mock data before we pull this in.

swpnit82:
We are working on testcases

packages/server-core/src/user/inventory-item/inventory-item.model.ts
```diff
@@ -76,6 +76,26 @@ export default (app: Application): any => {
       through: models.user_inventory,
       foreignKey: 'inventoryItemId'
     })
+    ;(inventoryItem as any).hasMany(models.user_trade, { foreignKey: 'fromInventoryItemId1', required: true })
```
HexaField, line 79: this is definitely not the right way to implement this. this should only have one association with user_trade

HexaField (changes requested):

packages/server-core/src/user/user-trade/user-trade.hooks.ts
```diff
@@ -0,0 +1,63 @@
+import { disallow } from 'feathers-hooks-common'
+import addAssociations from '@xrengine/server-core/src/hooks/add-associations'
+import { HookContext } from '@feathersjs/feathers'
+import * as authentication from '@feathersjs/authentication'
+import app from "../../../../server/src/app"
```
HexaField, line 5: `server-core` cannot import from `server` package

packages/server-core/src/user/user-trade/user-trade.model.ts
```diff
@@ -0,0 +1,41 @@
+import { Sequelize, DataTypes } from 'sequelize'
+import { Application } from '../../../declarations'
+
+export default (app: Application): any => {
```
HexaField, line 4: why do we need to store user trade in the database? this should just be a feathers service rather than a feathers-sequelize service & model

HexaField (changes requested):

Issue #3702, opened 2021-10-14
3D Loading Screen


HexaField:
It would be awesome to have a sick 3d loading screen, maybe something that fades from a blank 3d scene with a 3d loading icon into the scene as it loads.

speigg:
Closing in favor of #4076

PR #3895, opened 2021-10-28
3D Loading Screen #3702

frozencrystal:
Hi,
Please check out the new [3D Loading Screen](https://github.com/XRFoundation/XREngine/issues/3702) that uses the Engine's loop.
Thanks.

### DIFF
--- a/packages/engine/src/scene/systems/LoadingSystem.ts
+++ b/packages/engine/src/scene/systems/LoadingSystem.ts
@@ -1,0 +1,90 @@
+import { AmbientLight, Box3, PerspectiveCamera, Scene, WebGLRenderer } from 'three'
+import { GLTFLoader } from '@xrengine/engine/src/assets/loaders/gltf/GLTFLoader'
+import { EngineEvents } from '@xrengine/engine/src/ecs/classes/EngineEvents'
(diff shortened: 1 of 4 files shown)


packages/engine/src/scene/systems/LoadingSystem.ts
```diff
@@ -0,0 +1,90 @@
...
+  gltfScene: any
+  needsResize: boolean
+  constructor() {
+    LoadingRenderer.instance = this
+    //Todo
+    this.onResize = this.onResize.bind(this)
+
+    const loaderRendererCanvasId = 'loader-renderer-canvas'
+    const canvasElement = document.getElementById(loaderRendererCanvasId) as any
+    const url = 'https://172.160.10.156:3000/models/devices/razer_laptop.glb'
```
kinjalravel, line 26: @frozencrystal  Why is this hard coded ? can you please change

HexaField: @kinjalravel I believe this is a temporary solution - this will not be the model we use when this is fully implemented

kinjalravel: @frozencrystal @HexaField How about we pass this as props?

PR #3910, opened 2021-10-31
Hamza/add gltf transformation in editor

hamzzam:
## Summary

--

## Checklist

## References

## Reviewers


packages/engine/src/assets/functions/LoadGLTF.ts
```diff
@@ -71,6 +74,50 @@ export async function LoadGLTF(url: string): Promise<LoadGLTFResultInterface> {
   })
 }
 
+/**
+ * Loads an Instanced Asset which is in GLTF format and uses EXT_mesh_gpu_instancing extension.
+ *
+ * @param url URL of the asset.
+ * @returns a promise of {@link LoadGLTFResultInterface}.
+ */
+export async function LoadInstancedGLTF(url: string): Promise<LoadGLTFResultInterface> {
```
NPatel10, line 83: Logic in this function is similar to the logic in Asset Loader. Check if these 2 can be merged, otherwise PR is good.

packages/engine/src/assets/classes/AssetLoader.ts
```diff
@@ -197,6 +200,55 @@ export class AssetLoader {
     load(params, onLoad, onProgress, onError)
   }
 
+  static async loadInstancedGLTF(
```
patreeceeo, line 203: Can this just return promise instead of accepting an `onLoad` callback arg _and_ returning a promise? Like what [this guy](https://github.com/XRFoundation/XREngine/pull/3910/files#diff-0101def9fedb24e71b1585b997ad39b8f4fe0bf91c09a332d8e1ad194cf0bda3R83) is doing.

patreeceeo: Also this seems almost the same as that function. Why both?

packages/engine/src/assets/functions/LoadGLTF.ts
patreeceeo:
```suggestion
export async function loadInstancedGLTF(url: string): Promise<LoadGLTFResultInterface> {
```

hamzzam:
Made the required changes and completed the pending changes.

HexaField:
Looks like there is a missing file that is causing checks to fail, also need to resolve the conflict and merge dev in.

HexaField:
Looking good, just needs some instructions and a few files to test with so we can QA

hamzzam:
> Looking good, just needs some instructions and a few files to test with so we can QA

Added.

hamzzam:
GPU instancing can be used to instantiate multiple copies of an object without incurring extra draw calls. This functionality has been added in the XREngine with a focus on static scene objects.
Kindly use the following guidelines to use this feature:

1. Model your object in some modelling tool like blender etc and make the copies as instanced versions of the original mesh.
    **NOTE**: When modelling your object, remember to only assign one material to one mesh. Mesh that have multiple materials assigned to them cannot be GPU instantiated currently.
2. Export the glb for your model and load it in the XREngine editor using Model Node.
3. Set the _Using GPU instancing Flag_ in the given Model Node to true.
    **NOTE**: The model node can be selected as a complete object. The individual instances in the glb cannot be selected/translated in the editor.
4. Some test models are present on this path _models/debug/glTF-instancing._

HexaField:
Seems to work great! Thanks a lot for this!

HexaField (approved):

PR #3915, opened 2021-11-01
Storage provider tests

Abhishek-Pathak-Here:


packages/server-core/src/media/storageprovider/local.storage.ts
```diff
@@ -15,6 +15,29 @@ import { FileContentType } from '@xrengine/common/src/interfaces/FileContentType
 const keyPathRegex = /([a-zA-Z0-9/_-]+)\/[a-zA-Z0-9]+.[a-zA-Z0-9]+/
 
 export class LocalStorage implements StorageProviderInterface {
+  beforeTest(): Promise<any> {
```
HexaField, line 18: these helper functions should be in the test files rather than here.

packages/server-core/src/media/storageprovider/s3.storage.ts
```diff
@@ -97,13 +127,12 @@ export class S3Provider implements StorageProviderInterface {
     })
   }
 
-  listObjects = async (prefix: string, recursive = false): Promise<StorageListObjectInterface> => {
+  listObjects = async (prefix: string): Promise<StorageListObjectInterface> => {
```
HexaField, line 130: we need to keep this recursive option and the delimiter

HexaField (changes requested):

packages/server-core/src/media/storageprovider/s3.storage.ts
Abhishek-Pathak-Here: @HexaField  Should we do it like this?

HexaField: yes, though we want to keep the default to false, not true

packages/server-core/src/media/storageprovider/storageprovider.interface.ts
```diff
@@ -51,9 +51,9 @@ export interface StorageProviderInterface {
 
   /**
    * Gets the provider
-   * @returns {StorageProviderInterface}
+   * @returns {any}
```
HexaField, line 54: this should return type `StorageProviderInterface`

HexaField (changes requested):

HexaField (approved):

PR #3928, opened 2021-11-01
bitecs+threejs preliminary integration

NateTheGreatt:
this is all that is needed in order to optimize network serialization for now

should toss this code and integrate [threecs](https://github.com/thirdroom/threecs-next) when it's stable


speigg (approved):

NateTheGreatt:
issues found during QA, reverting to draft to investigate

packages/engine/src/avatar/functions/createAvatar.ts
```diff
@@ -55,14 +125,21 @@ export const createAvatar = (spawnAction: typeof NetworkWorldAction.spawnAvatar.
...
-    rotation: new Quaternion().copy(spawnAction.parameters.rotation),
-    scale: new Vector3(1, 1, 1)
-  })
+
+  const position = createVector3Proxy(TransformComponent.position, entity).copy(spawnAction.parameters.position)
+
+  const rotation = createQuaternionProxy(TransformComponent.rotation, entity).copy(spawnAction.parameters.rotation)
+
+  // todo: figure out why scale makes avatar disappear
+  // const scale = createVector3Proxy(TransformComponent.scale, entity)
```
HexaField, line 134: Scale needs to be set to (1, 1, 1) by default, is the avatar disappearing because this line is setting all scale values to 0?

NateTheGreatt: strangely enough, it happens regardless of value! i just forgot to include `.copy(new Vector3(1, 1, 1))` in the commented out code

NateTheGreatt: this is something that we can figure out in the `threecs-next` repo

PR #3930, opened 2021-11-02
Mappa#4: improve perf

patreeceeo:
Finally got a breakthrough on this one! What really did it was changing some code where I was iterating over the keys of a map. The code is a class that allows for using a combination of primitive values as the key into a map (what I'm calling `ArrayKeyedMap` for now). It accomplishes this by turning each array key in to a string key, so that it maps by value rather than by reference. In the `keys` method of this class, I iterate over the keys of the underlying ES6 Map and I was looking up the original array stored in yet another ES6 Map. Realized that I could instead just store a set of the original array keys and iterate over that.


patreeceeo:
building failing on a test which seems to have nothing to do with this PR: 

```   1) Uncaught error outside test suite:
     Uncaught TSError: ⨯ Unable to compile TypeScript:
Error: src/common/classes/GenerateMeshBVHWorker.ts(48,65): error TS2345: Argument of type 'false' is not assignable to parameter of type 'MeshBVHDeserializeOptions | undefined'.

      at createTSError (/home/runner/work/XREngine/XREngine/node_modules/ts-node/src/index.ts:744:12)
      at reportTSError (/home/runner/work/XREngine/XREngine/node_modules/ts-node/src/index.ts:748:19)
      at getOutput (/home/runner/work/XREngine/XREngine/node_modules/ts-node/src/index.ts:935:36)
      at Object.compile (/home/runner/work/XREngine/XREngine/node_modules/ts-node/src/index.ts:1237:30)
      at Module.m._compile (/home/runner/work/XREngine/XREngine/node_modules/ts-node/src/index.ts:1364:30)
      at Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
      at Object.require.extensions.<computed> [as .ts] (/home/runner/work/XREngine/XREngine/node_modules/ts-node/src/index.ts:1368:12)
      at Module.load (node:internal/modules/cjs/loader:981:32)
      at Function.Module._load (node:internal/modules/cjs/loader:822:12)
      at Module.require (node:internal/modules/cjs/loader:1005:19)```

**Comment by patreeceeo** [comment]
Just found another performance boost: Plain ol' JS objects are faster than ES6 Maps in some cases.

**Comment by patreeceeo** [comment]
Just found another performance boost: Plain ol' JS objects are faster than ES6 Maps in some cases. Willl clean up these changes I have locally and push them up.

**Comment by patreeceeo** [comment]
Working on fixing the tests now. But this is looking really good.

File: packages/engine/src/map/MapUpdateSystem.ts [hunk 1]
```diff
@@ -82,21 +86,21 @@ export default async function MapUpdateSystem(world: World): Promise<System> {
 
     if (mapState.needsUpdate) {
       // Perf hack: Start with an empty array so that any children that have been purged or that do not meet the criteria for adding are implicitly removed.
```
NateTheGreatt, line 88: is this comment still relevant after the change made below?

packages/engine/src/map/MapUpdateSystem.ts
```diff
@@ -82,21 +86,21 @@ export default async function MapUpdateSystem(world: World): Promise<System> {
...
-      const subSceneChildren = []
+      object3dComponent.value.children.length = 0
```
NateTheGreatt, line 89: just for my understanding: was this change for performance or did you need to maintain other references to this array (or both)?

NateTheGreatt (approved):
looks good to me, great work!

patreeceeo:
That was for really for simplicity I guess and I assumed it helps slightly
with performance but didn’t test that specifically.

packages/engine/src/map/MapUpdateSystem.ts
patreeceeo: That was really just for simplicity, I assumed it would help with performance but didn't test that specifically.

packages/engine/src/map/MapUpdateSystem.ts
patreeceeo: Yeah, just decided (somewhat impulsively) to do it another way

patreeceeo:
thanks!

PR #3965, opened 2021-11-04
Location Not loading correct scene bug fix

zulqarnainhanif:
pass correct location name to the world and other component from loction sub page route


HexaField (approved):
Works for me!

PR #4019, opened 2021-11-09
Portal Refactor

HexaField:
## Summary

- Avatars persist when moving between scenes
- When on local dev, gameserver process shuts down and restarts with new scene
- Adds `EngineState` client state
- Fix lazy loading reloading components even when using react router
- Improvements to mocha test configuration

## Checklist

## References

## Reviewers


packages/engine/tests/portal/portal.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import dotenv from 'dotenv-flow'
+
+dotenv.config({
+  path: appRootPath.path,
+  silent: true
+})
+
+const sceneData = parseSceneDataCacheURLs(sceneJson, process.env.LOCAL_STORAGE_PROVIDER!)
+
+describe.skip('Portal', () => {
```
speigg, line 22: skip intentional?

packages/server-core/src/world/scene/scene.service.ts
```diff
@@ -53,19 +78,23 @@ export default (app: Application) => {
...
+  })
+  app.use('/cubemap/:entityId', getCubemapBake(app))
 
   /**
    * Get our initialized service so that we can register hooks
    *
    * @author Vyacheslav Solovjov
    */
   const service = app.service('scene')
+  app.service('portal')
```
speigg, line 97: does this do anything?

speigg (approved):

packages/engine/tests/portal/portal.test.ts
HexaField: yeah, it's breaking because PhysX is persisting between tests. gotta figure out a better solution to global values with mocha

packages/server-core/src/world/scene/scene.service.ts
HexaField: oops!

HexaField: it appears not. thanks

Issue #3134, opened 2021-09-05
3D Interactables UI


shawwwwwwww:
Rewrite modal with ethereal.js (in engine)

volumetric, video & audio objects have 3D ‘PLAY’ and ‘PAUSE’ popup instead of ‘INTERACT’

link modal

interactable modal

 

Essentially we want to create an XRUI that displays these models directly within our engine, rather than in a DOM overlay

In a way that would also be compatible with a VR experience, for example.

The existing avatar username labels are a simple example of how our XRUI system works. You can see how this is implemented by looking at AvatarUISystem.tsx

So, probably first step would be to remove our model-viewer dependency and display these models entirely in our engine

PR #4044, opened 2021-11-10
3D Interactables UI #3134

rondoor124:
This is the PR for the [3D Interactables UI](https://github.com/XRFoundation/XREngine/issues/3134)


packages/engine/src/scene/systems/RendererUpdateSystem.ts
```diff
@@ -10,12 +11,18 @@ import { System } from '../../ecs/classes/System'
...
+const renderedUpdatableQuery = defineQuery([UpdatableComponent, RenderedComponent])
 
 export default async function RendererUpdateSystem(world: World): Promise<System> {
   return () => {
     for (const entity of renderedQuery()) {
       const obj = getComponent(entity, Object3DComponent)
       ;(obj.value as unknown as Updatable).update(world.delta)
     }
+
+    for (const entity of renderedUpdatableQuery()) {
```
HexaField, line 23: do we need a second query here? If my understanding is correct, Object3DComponent should be the only component that holds the Object3D, the Updateable component a tag component for running an update function on the fixed loop, and RenderedComponent is a tag component for running an update function on the free loop. I think for the interactable UI, we want to only use the RenderedComponent and not use Updateable component at all.

HexaField (changes requested):

packages/engine/src/scene/systems/RendererUpdateSystem.ts
rondoor124: Okay got it.

packages/editor/src/nodes/ModelNode.ts
```diff
@@ -103,7 +104,7 @@ export default class ModelNode extends EditorNodeMixin(Model) {
   boundingSphere = new Sphere()
   gltfJson = null
   isValidURL = false
-  _matrixAutoUpdate = false
+  _matrixAutoUpdate = true
```
HexaField, line 107: should be default to false

packages/editor/src/components/inputs/InteractableGroup.tsx
```diff
@@ -0,0 +1,216 @@
...
+  node?: object
+  t: Function
+}
+
+//Declaring TriggerVolumeNodeEditor state
+type InteractableGroupState = {
+  options: any[]
+}
+
+export class InteractableGroup extends Component<InteractableGroupProps, InteractableGroupState> {
```
HexaField, line 63: please convert to a functional component

packages/editor/src/components/inputs/ArrayInputGroup.tsx
```diff
@@ -0,0 +1,100 @@
...
+export interface ArrayInputGroupState {
+  count: number
+  values: any
+}
+
+/**
+ *
+ * @author Ron Oyama
+ */
+export class ArrayInputGroup extends Component<ArrayInputGroupProp, ArrayInputGroupState> {
```
HexaField, line 22: please convert to a functional component

HexaField (changes requested):

rondoor124:
Hi @HexaField @speigg 
I fixed the above issue. please check and give me feedback.
Thanks.

HexaField:
> Hi @HexaField @speigg I fixed the above issue. please check and give me feedback. Thanks.

Just need to resolve the conflict then we can give this a review & QA

rondoor124:
Yes got it

packages/engine/src/interaction/functions/interactUI.ts
```diff
@@ -0,0 +1,237 @@
...
+  addComponent(ui.entity, TransformComponent, {
+    position: new Vector3(transform.position.x, transform.position.y + 2.5, transform.position.z),
+    rotation: new Quaternion(),
+    scale: new Vector3(1, 1, 1)
+  })
+
+  //create text
+  const interactTextEntity = createInteractText(interactiveComponent.data.interactionText)
+
+  const timer = setInterval(() => {
```
speigg, line 54: Using setInterval seems like an anti-pattern; It would be better to use an enter() query w/ XRUIComponent in the InteractionSystem

packages/engine/src/interaction/ui/InteractiveModalView.tsx
```diff
@@ -0,0 +1,433 @@
+import React from 'react'
+import { createState } from '@hookstate/core'
+import { createXRUI } from '../../xrui/functions/createXRUI'
+import { useXRUIState } from '@xrengine/engine/src/xrui/functions/useXRUIState'
+import { NavigateNext, NavigateBefore } from '@mui/icons-material'
+
+let callback: any
+
+export function createInteractiveModalView(data: any) {
```
speigg, line 9: Can we add types to the data parameter here?

packages/engine/src/interaction/ui/InteractiveModalView.tsx
```diff
@@ -0,0 +1,433 @@
+import React from 'react'
+import { createState } from '@hookstate/core'
+import { createXRUI } from '../../xrui/functions/createXRUI'
+import { useXRUIState } from '@xrengine/engine/src/xrui/functions/useXRUIState'
+import { NavigateNext, NavigateBefore } from '@mui/icons-material'
+
+let callback: any
```
speigg, line 7: We should avoid creating components that rely on global state unless we know for certain we only ever want a single instance active at a time. This callback can be passed in to the `createInteractiveModalView` function, and then passed along as a prop to `InteractiveModalView`. E.g.:

```ts
export function createInteractiveModalView(data: <type this>) {
return createXRUI(() => <InteractiveModalView callback={data.callback}></InteractiveModalView>), createInteractiveModalState(data))
}
```

packages/engine/src/interaction/ui/InteractiveModalView.tsx
```diff
@@ -0,0 +1,433 @@
...
+    }
+  }
+
+  const linkClick = (urls) => {
+    for (let url of urls) {
+      window.open(url, '_blank')!.focus()
+    }
+  }
+
+  const renderMedia = (width, height, position?) => {
```
speigg, line 94: FYI, if you want to add :hover styles, you can use `<style dangerouslySetInnerHTML={}></style>` to add normal CSS. Hover styles will be rendered automatically.

packages/engine/src/interaction/functions/interactUI.ts
```diff
@@ -0,0 +1,237 @@
...
+      }
+      hideInteractUI(entity)
+      clearInterval(timer)
+
+      //rendered for model rotation
+      addComponent(ui.entity, RenderedComponent, {})
+
+      if (object3D && object3D.value) {
+        //@ts-ignore
+        object3D.value.update = () => {
```
speigg, line 69: This looks like a dangerous thing to do; ethereal objects already have an `update` function (called in the XRUISystem), and we are overriding that here. In general, this pattern makes me nervous. We should be putting this kind of logic inside of a system (e.g., InteractiveSystem).

speigg (changes requested):

rondoor124:
Thanks for the feedback. I will fix it very soon.

rondoor124:
@speigg I fixed the feedback. please review.
Thanks.

speigg (approved):

PR #4064, opened 2021-11-12
added chatState in the globalThis, added navmesh to ground plane and …

alextitonis:
…made an override system for the movement commands

## Summary

A summary of changes being made in this PR

## Checklist

## References

## Reviewers


patreeceeo:
Hey Alex, would you mind adding some explanation in the PR description?

packages/engine/src/networking/utils/getUser.ts
```diff
@@ -7,11 +7,21 @@ import { UserNameComponent } from '../../scene/components/UserNameComponent'
...
-export function getUserEntityByName(name: string) {
+export function getUserEntityByName(name: string, localUserId) {
+  /*console.log('looking for entity: ' + name)
+  console.log('values: ' + JSON.stringify(Engine.defaultWorld.clients.values()))
   const client = Array.from(Engine.defaultWorld.clients.values()).find((c) => {
     return c.name === name
   })
-  return client ? useWorld().getUserAvatarEntity(client.userId) : undefined
+  return client ? useWorld().getUserAvatarEntity(client.userId) : undefined*/
+  const world = useWorld()
```
speigg, line 17: If this function is meant to be called outside of a System, then it's preferable to use Engine.defaultWorld rather than `useWorld`

speigg (approved):

packages/client-core/src/social/services/ChatService.ts
```diff
@@ -215,6 +215,8 @@ export const accessChatState = () => state
 
 export const useChatState = () => useState(state) as any as typeof state
 
+globalThis.chatState = state
```
HexaField, line 218: why does chat state need to be on the global state? can we use the accessChatState hook instead?

packages/engine/src/scene/functions/createGround.ts
```diff
@@ -33,5 +37,15 @@ export const createGround = async function (entity: Entity, args: GroundProps, i
 
   createCollider(entity, mesh)
 
+  if (isClient) {
```
HexaField, line 40: lets move this to the navigation system and add a GroundPlaneTagComponent component to query for, to keep this decouple from the core engine functionality

HexaField (changes requested):

packages/engine/src/scene/functions/createGround.ts
HexaField: For now, maybe we can just have an option in the editor on the Scene node for if the scene should have navmeshes or not. That way we can optimise scenes that don't need it, and have a single source of truth per-secne.

alextitonis:
> 
> 
> Hey Alex, would you mind adding some explanation in the PR description?

In the PR, there are 2 things done, first i've added the chat state in the globalThis, so i can access it through puppeteer.evaluate
And added navmesh based on the Map object in the ground plane and added an override for the /move /follow commands, to set exact target position ignoring the raycasting (x,y,z)

packages/engine/src/networking/utils/getUser.ts
alextitonis: updated it

shawwwwwwww:
lgtm

patreeceeo:
Can we get rid these console logs if they aren't used anymore? And any unused args/imports/declarations

packages/editor/src/nodes/GroundPlaneNode.ts
```diff
@@ -70,5 +72,8 @@ export default class GroundPlaneNode extends EditorNodeMixin(GroundPlane) {
       receive: this.receiveShadow,
       cast: false
     })
+    this.addGLTFComponent('navmesh', {
```
HexaField, line 75: this is not needed, generateNavmesh will be serialised into `ground-plane` and `navmesh` has no deserialization in either the location nor the editor

HexaField (changes requested):

HexaField (approved):

PR #4098, opened 2021-11-15
Camera system updates

mohsenheydari:
## Summary

- Follow camera raycast bugfix
- Added ray hit results caching
- Fixed avatar opacity update issue

## Checklist

## References

## Reviewers

@HexaField


packages/engine/src/scene/systems/SceneObjectSystem.ts
```diff
@@ -50,6 +50,11 @@ export default async function SceneObjectSystem(world: World): Promise<System> {
 
       ;(object3DComponent.value as any).entity = entity
 
+      // Enable scene layer
+      object3DComponent.value.traverse((obj: Object3D) => {
```
HexaField, line 54: setting scene layer by default will cause all objects to be overridden - this has been observed in the golf game with a lot of weirdness due to custom avatars. rather than setting it here, we should set it whereever these objects are created.

packages/engine/src/camera/systems/CameraSystem.ts
```diff
@@ -248,6 +252,7 @@ export default async function CameraSystem(world: World): Promise<System> {
           coneDebugHelpers.push(arrow)
           arrow.traverse((obj: Object3D) => {
             obj.layers.set(CameraLayers.Gizmos)
+            obj.layers.enable(CameraLayers.Render)
```
HexaField, line 255: my intention with the new camera layers, was that rather than adding the render layer to gizmos, we add gizmos to the camera when we want to render them, this way we dont have to set gizmos as often, and we have a single source of truth (the camera) for if gizmos should be rendered (rather than each gizmo)

HexaField (changes requested):

packages/engine/src/camera/systems/CameraSystem.ts
mohsenheydari: What about the parent/child transform relationship that the camera object introduces. For example, if I want to give a gizmo static world position in the world I need to convert the world to local position in order to make the gizmo stay at the same place.

HexaField (changes requested):

packages/engine/src/camera/systems/CameraSystem.ts
HexaField: This is a seperate issue i think, layers just allow for a bitmask to optionally render things by category. It has no bearing on render order or transforms.

packages/engine/src/avatar/functions/avatarFunctions.ts
```diff
@@ -96,8 +97,10 @@ const setupAvatar = (entity: Entity, model: any, avatarURL?: string) => {
...
 
-    if (typeof object.material !== 'undefined') {
-      // object.material = object.material.clone()
+    if (object.material) {
+      // Transparency fix
+      object.material.format = RGBAFormat
+
       materialList.push({
         id: object.uuid,
         material: object.material.clone()
```
NateTheGreatt, line 106: thanks to @HexaField's excellent catch here, he has exposed a great target for a unit test for `setupAvatar`. it would be great to add a small test to assert that the material is indeed a clone, and not an identical reference.

e.g. `it('should make a new material instance per avatar')`

packages/engine/src/scene/functions/loadGLTFModel.ts
```diff
@@ -93,6 +94,11 @@ export const parseGLTFModel = (
 ) => {
   // console.log(sceneLoader, entity, component, sceneProperty, scene)
 
+  // Enable scene layer
+  scene.traverse((obj: Object3D) => {
+    obj.layers.enable(CameraLayers.Scene)
+  })
```
NateTheGreatt, line 100: this might also be a good target for a unit test for `parseGLTFModel`

e.g. `it('should enable scene layer for camera')`

NateTheGreatt (commented):
pulled this down and did some QAT, feels great. nice work @mohsenheydari!

while here, perhaps we should take this opportunity to add some test coverage to some of the modified functions? i marked a couple possible places

packages/engine/src/avatar/functions/avatarFunctions.ts
HexaField: put the skeleton for these tests in - loading an avatar from tests is non trivial, as we have to load it using fs of which i can see there being some roadblocks

packages/engine/src/scene/functions/loadGLTFModel.ts
HexaField: good idea! i fixed the loadGLTFModel test and made it now check for this. will need to extend this in the future, but this particular issue will be caught now.

PR #4125, opened 2021-11-16
implement hotkey functionality

zulqarnainhanif:
## Summary

implement ctrl/cmd+s functionality
implemented the functionality for grouping and duplicating nodes from hierarchy as well.

## Checklist

## References

https://github.com/XRFoundation/XREngine/issues/4113

## Reviewers

@HexaField


speigg:
Awesome

packages/editor/src/functions/utils.ts
```diff
@@ -34,8 +34,23 @@ export function unique(arr, maybeComp) {
...
+
+  const iOS13_iPad = navigator.platform === 'MacIntel'
+
+  const iOS1to12quirk = () => {
+    var audio = new Audio() // temporary Audio object
+    audio.volume = 0.5 // has no effect on iOS <= 12
+    return audio.volume === 1
+  }
+
+  return iOS_1to12 || iOS13_iPad || iOS1to12quirk()
```
speigg, line 49: Can you explain why this is necessary?

speigg (approved):

packages/editor/src/functions/utils.ts
zulqarnainhanif: I have the latest mac and it kept showing ctrl + s for it while it should show cmd + s.

The reason behind this is that for latest macs it return `MacIntel` as navigator.platform which wasn't being caught in the previous solution.

So i found an interesting solution on stackoverflow for this which i implemented above.

speigg: Does this solution also work with older Macs?

HexaField (approved):
Group and duplication functionality seems to not be working for me, but I don't mind as we are currently refactoring the way the editor nodes work, so we can revisit the specific funcitonality at a later time. The keybindings work, so I'm happy with this!

PR #4134, opened 2021-11-16
add-msa-db-models

aleksmaz:
-add db models/interfaces for MSA-Basketball States

### DIFF
--- a/packages/server-core/src/match/team/team.docs.ts
+++ b/packages/server-core/src/match/team/team.docs.ts
@@ -1,0 +1,51 @@
+export default {
+  definitions: {
+    team: {
+      type: 'object',
+      properties: {
+        profileId: {
+          type: 'string'
+        },
+        title: {
+          type: 'string'
+        },
+        games_played: {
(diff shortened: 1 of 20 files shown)


packages/server-core/src/match/team/team.docs.ts
```diff
@@ -0,0 +1,51 @@
...
+        profileId: {
+          type: 'string'
+        },
+        title: {
+          type: 'string'
+        },
+        games_played: {
+          type: 'integer'
+        },
+        wins: {
```
HexaField, line 15: some of these properties seem specific to the project this is for, i wonder if we can replace all of these fields with a generic metadata json object for generic object storage. I think the 'team' funcitonality is good to have, given it is for actual users, otherwise i'm not sure if this should be in the base stack. if not, we will need to figure out an elegant way to extend the database model via projects.

HexaField (changes requested):

HexaField:
Closing in favour of project DB customisation talked about elsewhere

PR #4135, opened 2021-11-17
Added retry to S3 storage provider for NoSuchBucket errors.

barankyle:
## Summary

Some storageProvider tests have been intermittently failing due to NoSuchBucket
errors from S3. These seem to be entirely due to records of a new bucket's existence not
propagating everywhere instantaneously.

Added a new method to s3.storage.ts, which performs a headBucket call (which seems to be more reliable
than when some other S3 method is called) and, if that doesn't return an error, calls waitFor('bucketExists').
When this resolves, it resolves the returned promise.

Most calls now use this in their error handling if they get a NoSuchBucket error. If it resolves successfully,
those functions re-call themselves.

## Checklist

## References

## Reviewers


HexaField (commented):
I wonder if instead of this verbose retry, we can wrap all these in a helper function that abstracts the bucket check

HexaField:
Another thought, if this is also happening with the tests, should we instead add a promise that returns when we can verify the bucket does exist? This way it's scoped to the test that is failing (as this doesnt appear to be occuring in local development or on deployment)

PR #4142, opened 2021-11-17
Fix remote emotes

alextitonis:
## Summary

A summary of changes being made in this PR

## Checklist

## References

## Reviewers


HexaField (changes requested):
I looked into this issue. The pattern matching definitely works, so there is no need to match the action type manually with the string. All that needs to happen is the original code needs to change from `NetworkWorldAction.avatarAnimation.matches` to `NetworkWorldAction.avatarAnimation.matchesFromAny`

Please revert the changes and make the new one (changing .matches to .matchesFromAny).

HexaField (approved):

PR #4165, opened 2021-11-18
Fixed issue with subdomain cleanup.

barankyle:
## Summary

During cleanupOldGameservers(), we were patching subdomain-provisions where instanceId
was null and gs_id didn't match an existing gameserver. If a GS crashes, though, the subdomain
record might still have the instanceId, meaning it'll be stuck as allocated forever. Removed
instanceId: null from query so that it'll clean up all subdomain-provisions if they don't
match a current GS.

## Checklist

## References

## Reviewers


HexaField (approved):
nice catch

PR #4174, opened 2021-11-19
Added a view to show location not found

zulqarnainhanif:
## Summary

add a screen that tells the user to check his location

## Checklist

## References

https://github.com/XRFoundation/XREngine/issues/3874

## Reviewers

@HexaField


barankyle:
I'd made a warning modal trigger that's supposed to do this already: https://github.com/XRFoundation/XREngine/blob/dev/packages/client/src/components/World/GameServerWarnings.tsx#L62

I'm not sure why it's not triggering, since it runs off the same state value. I'm not sure we need the new things in this PR if we could make the modal work again.

HexaField (changes requested):
Kyle's comment is correct, we should be re-using the modal we already have for this.

zulqarnainhanif:
@HexaField @barankyle update the warning component to stay there when value for noCountDown is true and remove my changes.

HexaField (approved):
Seems to work now, thanks.

barankyle (approved):

PR #4248, opened 2021-11-24
Editor ECS Scene Loading with Serializers & Deserializers

NPatel10:
## Summary

- [x] scene hierarchy into entity tree (temporarily until we do fix up child transform system and implement ECS renderer)
- [x] mixin classes into (de)serializer functions & ECS components
- [x] editor properties use ECS components instead of node classes
- [x] 'Elements' to be 'prefabs' and add list of prefabs that house the necessary scene types to create a new node from
- [x] all editor commands to work with new format

## Checklist

## References

## Reviewers


packages/editor/src/functions/getIntersectingNode.ts
```diff
@@ -1,27 +1,37 @@
...
-        if (curObject.isNode) {
-          break
-        }
-        curObject = curObject.parent
-      }
-      if (curObject && curObject !== scene && !curObject.ignoreRaycast) {
-        ;(result as RaycastIntersectionNode).node = curObject
-        return result as RaycastIntersectionNode
-      }
+function getParentEntity(obj: Object3DWithEntity): Object3DWithEntity {
```
HexaField, line 12: could we do an Object3D.traverseParents to avoid the while loop?

HexaField (changes requested):
I think we should avoid making a system for each node, serialise & deserialise functions should be enough.

packages/editor/src/services/EditorServices.ts
```diff
@@ -1,9 +1,12 @@
...
 import { createState, useState } from '@hookstate/core'
 
-const state = createState({
-  projectName: null as string,
-  sceneName: null as string
+const state = createState<{
+  projectName: string | null
+  sceneName: string | null
+}>({
+  projectName: null,
```
HexaField, line 8: we cn do `null! as string` here to avoid manual typing :)

packages/editor/src/managers/SceneManager.ts
```diff
@@ -186,24 +193,38 @@ export class SceneManager {
...
-    Engine.renderer = screenshotRenderer
+    Engine.renderer = this.screenshotRenderer
     this.disableUpdate = true
-    let scenePreviewCamera = sceneNode.findNodeByType(ScenePreviewCameraNode)
-    if (!scenePreviewCamera) {
-      scenePreviewCamera = new ScenePreviewCameraNode()
-      Engine.camera.matrix.decompose(scenePreviewCamera.position, scenePreviewCamera.rotation, scenePreviewCamera.scale)
-      CommandManager.instance.executeCommandWithHistory(EditorCommands.ADD_OBJECTS, [scenePreviewCamera])
+
+    const entities = getAllEntitiesWithComponent(ScenePreviewCameraTagComponent)
```
HexaField, line 200: We have Engine.activeCameraEntity for storing the location camera entity, we should leverage this in the editor too for this

packages/engine/src/common/constants/ComponentNames.ts
```diff
@@ -0,0 +1,52 @@
+export enum ComponentName {
```
HexaField, line 1: good idea, just a few changes for clarity and types
```
export const SceneJSONComponentName = {
  MT_DATA: 'mtdata' as const,
  ...
}
```

packages/engine/src/initializeEngine.ts
```diff
@@ -327,7 +317,8 @@ export const initializeEngine = async (initOptions: InitializeOptions = {}): Pro
   } else if (options.type === EngineSystemPresets.EDITOR) {
     Engine.userId = 'editor' as UserId
     Engine.isEditor = true
-    // Engine timer should not start here for eidtor since we load scene after selecting.
+    // Engine timer should not start for eidtor since we load scene after selecting.
+    Engine.engineTimer.stop()
```
HexaField, line 321: the engine should be always running. we will need it for other things. loading and unloading scenes should only be changes in state and add/removing entities/components

HexaField (changes requested):

packages/editor/src/components/properties/NameInputGroup.tsx
```diff
@@ -18,58 +23,63 @@ const StyledNameInputGroup = (styled as any)(InputGroup)`
...
  * @type {class component}
  */
-export const NameInputGroup = (props: Types) => {
-  let [name, setName] = useState(props.node.name)
-  let [focusedNode, setFocusedNode] = useState(null)
+export const NameInputGroup: EditorComponentType = (props) => {
+  const nodeName = getComponent(props.node.entity, NameComponent).name
+
+  let [name, setName] = useState(nodeName)
+  let [focusedNode, setFocusedNode] = useState<EntityTreeNode>()
```
zulqarnainhanif, line 36: Its better to use `const` for react component states

packages/editor/src/components/properties/ParticleEmitterNodeEditor.tsx
```diff
@@ -185,8 +63,8 @@ export const ParticleEmitterNodeEditor = (props: ParticleEmitterNodeEditorProps)
         smallStep={0.01}
         mediumStep={0.1}
```
zulqarnainhanif, line 64: we can define a const for the same props being passed again and again, it would reduce the file length and increase code readability.

const StepProps = {
  smallStep: 0.01,
  mediumStep: 0.1,
  largeStep: 1,
 ....
}

and can be used as

     <NumericInputGroup
        {...StepProps}
        value={particleComponent.ageRandomness}
        onChange={updateProperty(ParticleEmitterComponent, 'ageRandomness')}
      />

packages/editor/src/components/properties/ParticleEmitterNodeEditor.tsx
```diff
@@ -172,8 +50,8 @@ export const ParticleEmitterNodeEditor = (props: ParticleEmitterNodeEditorProps)
...
         largeStep={1}
-        value={props.node.ageRandomness}
-        onChange={onChangeAgeRandomness}
+        value={particleComponent.ageRandomness}
+        onChange={updateProperty(ParticleEmitterComponent, 'ageRandomness')}
```
zulqarnainhanif, line 54: we can define a const for the same props being passed again and again, it would reduce the file length and increase code readability.

const StepProps = {
smallStep: 0.01,
mediumStep: 0.1,
largeStep: 1,
....
}

and can be used as

 <NumericInputGroup
    {...StepProps}
    value={particleComponent.ageRandomness}
    onChange={updateProperty(ParticleEmitterComponent, 'ageRandomness')}
  />

packages/editor/src/functions/getIntersectingNode.ts
NPatel10: In `traverseParents` breaking the loop and early exit is not possible.

packages/editor/src/services/EditorServices.ts
NPatel10: `null` is also assigned during runtime in some cases.

packages/editor/src/components/properties/NameInputGroup.tsx
NPatel10: Fixed.

packages/engine/src/initializeEngine.ts
NPatel10: Fixed!

packages/editor/src/classes/MeshCombinationGroup.ts
```diff
@@ -74,7 +78,7 @@ async function meshStandardMaterialComparator(group, a, b) {
     a.aoMapIntensity === b.aoMapIntensity &&
     a.normalScale.equals(b.normalScale) &&
     a.emissive.equals(b.emissive) &&
-    (await meshBasicMaterialComparator(group, a, b)) &&
+    (await meshBasicMaterialComparator(group, a as any, b as any)) &&
```
zulqarnainhanif, line 81: If these promises aren't codependent we can use `await Promise.all([promise1,promise2])` that would be a lot quicker

packages/editor/src/commands/ModifyPropertyCommand.ts
```diff
@@ -2,113 +2,126 @@ import Command, { CommandParams } from './Command'
...
-
-        if (object.onChange) {
-          object.onChange(propertyName)
+  updateProperties(nodes: EntityTreeNode[], properties: PropertyType, component: ComponentConstructor<any, any>): void {
+    for (let i = 0; i < nodes.length; i++) {
+      const entity = nodes[i].entity
+      if (component) {
+        const comp = getComponent(entity, component)
+        if (comp) {
+          for (const propertyName in properties) {
```
zulqarnainhanif, line 92: avoid using the `for in` loop it is the slowest of them all. you can run a loop with Object.keys(properties)

packages/editor/src/commands/ModifyPropertyCommand.ts
```diff
@@ -2,113 +2,126 @@ import Command, { CommandParams } from './Command'
...
         }
       }
+      const nodeComponent = getComponent(entity, EntityNodeComponent)
+      for (const component of nodeComponent.components) {
+        const update = useWorld().sceneLoadingRegistry.get(component)?.update
+        if (update) update(nodes[i].entity, properties)
+      }
+    }
 
+    for (const propertyName in properties) {
```
zulqarnainhanif, line 112: same as above avoid usage of `for in`

packages/editor/src/commands/RotationCommand.ts
```diff
@@ -69,44 +67,39 @@ export default class RotationCommand extends Command {
...
-        spaceMatrix = tempMatrix.identity()
+        const obj3d = getComponent(lastSelectedObject.entity, Object3DComponent).value
+        obj3d.updateMatrixWorld()
+        spaceMatrix = obj3d.parent!.matrixWorld
       }
     }
 
     for (let i = 0; i < objects.length; i++) {
       const object = objects[i]
+      let obj3d = getComponent(object.entity, Object3DComponent).value
```
zulqarnainhanif, line 86: const can be used here for both variable i think

packages/editor/src/components/hierarchy/HierarchyPanelContainer.tsx
```diff
@@ -918,231 +130,147 @@ export default function HierarchyPanel() {
...
-        }
-        // check if pressed key equals to left arrow and node object children node length is greator then 0
-      } else if (e.key === 'ArrowLeft' && node.object.children.filter((o) => o.isNode).length > 0) {
-        if (e.shiftKey) {
-          collapseChildren(node)
-        } else {
-          collapseNode(node)
-        }
+        case 'ArrowLeft':
+          if (!node.entityNode.children || node.entityNode.children.length === 0) return
```
zulqarnainhanif, line 198: `node.entityNode?.children?.length === 0` will return the same response

packages/editor/src/components/hierarchy/HierarchyPanelContainer.tsx
```diff
@@ -918,231 +130,147 @@ export default function HierarchyPanel() {
...
-        //check if pressed key equals to arrow right
-        //and node property object children node length ius greator then zero.
-      } else if (e.key === 'ArrowRight' && node.object.children.filter((o) => o.isNode).length > 0) {
-        if (e.shiftKey) {
-          expandChildren(node)
-        } else if (node.object.children.filter((o) => o.isNode).length > 0) {
-          expandNode(node)
-        }
+        case 'ArrowRight':
+          if (!node.entityNode.children || node.entityNode.children.length === 0) return
```
zulqarnainhanif, line 205: `node.entityNode?.children?.length === 0` will return the same response

packages/editor/src/components/hierarchy/HierarchyTreeNode.tsx
```diff
@@ -0,0 +1,269 @@
...
+                )}
+              </div>
+              {/* {node.entitynode.issues && node.entitynode.issues.length > 0 && <NodeIssuesIcon node={node.entitynode} />} */}
+            </div>
+          </div>
+
+          <div
+            className={styles.nodeDropTraget}
+            style={{
+              marginLeft: node.depth > 0 ? node.depth * 8 + 20 : 0,
```
zulqarnainhanif, line 260: styling same as above for `beforeDropTarget`

packages/editor/src/components/inputs/InteractableGroup.tsx
```diff
@@ -44,162 +49,127 @@ const InteractableTheme = [
...
+  //     component: InteractableComponent,
+  //     properties: { payloadHtmlContent: sanitizedHTML }
+  //   })
+  // }
+
+  const interactableComponent = getComponent(props.node.entity, InteractableComponent)
+
+  const renderInteractableTypeOptions = () => {
+    switch (interactableComponent.interactionType) {
+      case 'infoBox':
```
zulqarnainhanif, line 84: why do we need a switch statement here if there is only 1 case that we need to handle?

packages/editor/src/components/properties/MediaSourceProperties.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { MediaComponent } from '@xrengine/engine/src/scene/components/MediaComponent'
+
+export const MediaSourceProperties: EditorComponentType = (props) => {
+  const { t } = useTranslation()
+
+  const mediaComponent = getComponent(props.node.entity, MediaComponent)
+
+  return (
+    <>
+      <InputGroup
```
zulqarnainhanif, line 17: we can populate views of such components from a JSON, since same HTML is being repeated again and again

packages/editor/src/components/properties/PortalNodeEditor.tsx
```diff
@@ -30,132 +28,101 @@ type PortalFilterOption = {
...
 /**
  * PortalNodeEditor provides the editor for properties of PortalNode.
  *
  * @author Josh Field <github.com/HexaField>
  * @type {class component}
  */
-export const PortalNodeEditor = (props: PortalNodeEditorProps) => {
-  let [portals, setPortals] = useState([])
+export const PortalNodeEditor: EditorComponentType = (props) => {
+  let [portals, setPortals] = useState<Array<{ value: string; name: string }>>([])
```
zulqarnainhanif, line 40: use of `const` here

packages/editor/src/components/properties/PropertiesPanelContainer.tsx
```diff
@@ -111,29 +119,20 @@ const NoNodeSelectedMessage = (styled as any).div`
...
-    setActiveNode({ node: selected[selected.length - 1] })
-  }, [selected])
+  const onSelectionChanged = () => setSelected([...CommandManager.instance.selected])
 
   const onObjectsChanged = (objects, property) => {
     const selected = CommandManager.instance.selected
 
-    setActiveNode({ node: selected[selected.length - 1] })
-
     if (property === 'position' || property === 'rotation' || property === 'scale' || property === 'matrix') {
```
zulqarnainhanif, line 129: ['position', 'rotation', 'scale', 'matrix'].includes(property)

packages/editor/src/managers/SceneManager.ts
```diff
@@ -385,6 +363,50 @@ export class SceneManager {
...
+
+    return target
+  }
+
+  removeUnusedObjects = (object3d: Object3D) => {
+    computeAndSetStaticModes(object3d)
+
+    function hasExtrasOrExtensions(object) {
+      const userData = object.userData
+      for (const key in userData) {
```
zulqarnainhanif, line 375: avoid usage of `for in` here as well

HexaField:
@zulqarnainhanif this feedback is great thanks. we can follow up with these in subsequent PRs.

packages/editor/src/components/hierarchy/HierarchyPanelContainer.tsx
NPatel10: When the `children` will be `undefined` then condition will become `false` instead of `true`

packages/editor/src/components/hierarchy/HierarchyPanelContainer.tsx
NPatel10: When the `children` will be `undefined` then condition will become `false` instead of `true`

packages/editor/src/components/properties/MediaSourceProperties.tsx
NPatel10: This is a good way to go. But currently all the editors for all nodes are implemented in this way and refactoring all of them is a big task in itself. Hence I am taking a note and will revisit this later.

HexaField (approved):

PR #4271, opened 2021-11-25
Fix CORS Issue

rondoor124:
I fixed cors issue using [cors-anywhere node module](https://www.npmjs.com/package/cors-anywhere) and this is tested with woocommerce URLs that have cors issue.
Please review and give me feedback.
Thanks


speigg:
@barankyle probably need your eyes on this

.env.local.default
```diff
@@ -56,6 +57,7 @@ SERVER_MODE=local
 SERVER_HOST=localhost
 SERVER_PORT=3030
 SERVER_URL=https://localhost:3030
+CROS_SERVER_PORT=3035
```
barankyle, line 60: Typo here - should be 'CORS' (and elsewhere that it's used)

packages/social/config/default.json
```diff
@@ -5,6 +5,7 @@
     "loginDisabled": false,
     "logo": "./logo.svg",
     "apiServer": "https://${VITE_SERVER_HOST}:${VITE_SERVER_PORT}",
+    "corsServer": "https://${VITE_SERVER_HOST}:${VITE_CORS_SERVER_PORT}",
```
barankyle, line 8: I don't think this is necessary. social is our ARC-related codebase, which is getting pulled out very soon and made into a separate XREngine project.

.env.local.default
barankyle, line 60: Typo here - should be 'CORS_SERVER_PORT', as well as down in server-core/src/appconfig.ts where it's used.

packages/social/config/default.json
barankyle, line 8: I don't believe this is needed, since the `social` package is our ARC-related code that's getting pull out of the main repo soon and moved to its own package.

packages/server-core/src/createCorsServer.ts
```diff
@@ -0,0 +1,42 @@
+import cors_proxy from 'cors-anywhere'
+import config from './appconfig'
+import fs from 'fs'
+import net from 'net'
+
+const createCorsServer = (port: number) => {
```
barankyle, line 6: Our use of SSL is different between development and production, as can be seen in server/src/index.ts; this will need to be accounted for here. In dev mode, we have the server use SSL, so that we can simulate a production environment with https://URLs. When in dev mode, the https server uses the self-signed certs that are pointed to here.

In production, we start up the servers with SSL off, and do SSL termination at the load balancer. Unless there's something I'm missing, the cors-anywhere server should not have httpOptions if the server's being started in production mode - again, see server/src/index.ts for determining whether SSL should be turned on.

packages/client-core/src/util/cors.tsx
```diff
@@ -0,0 +1,14 @@
+import i18n from 'i18next'
```
barankyle, line 1: I'm probably missing something here, but where is this file being used?

barankyle (changes requested):

speigg:
@barankyle CI/CD failed on your commit

packages/server-core/src/createCorsServer.ts
```diff
@@ -0,0 +1,45 @@
...
+  isPortTaken(port, () => {
+    createCorsServer(useSSL, certOptions, port)
+  })
+}
+
+const isPortTaken = (port, fn) => {
+  const tester = net
+    .createServer()
+    .once('error', (err) => {
+      if (err.name === 'EADDRINUSE') return fn()
```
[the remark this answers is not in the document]
rondoor124: @barankyle 
https://github.com/XRFoundation/XREngine/blob/dev/packages/server-core/src/createFileServer.ts#L29
I just reference this function without checking...
How about the incrase the port number?
`if (err.name === 'EADDRINUSE') return fn(port + 1, true)`
can it be a solution?

barankyle (approved):

speigg (approved):

Issue #4082, opened 2021-11-13
Improve performance auto tuning system


HexaField:
in WebGLRendererSystem.ts we have a framerate auto-tuner which changes the resolution and graphics settings depending on the framerate it detects. it needs some tweaking as it is a bit over-sensitive.

PR #4316, opened 2021-11-29
Add more granular sensitivity for #4082

rondoor124:
I added more granular sensitivity for [#4082](https://github.com/XRFoundation/XREngine/issues/4082)
Please review and give me feedback.
Thanks.


speigg:
Our `calculateMovingAverage` function doesn't make sense, which is why the sensitivity is probably failing right now. 

This is how we should be doing it:

```ts
class ExponentialMovingAverage {

    mean: number
    multiplier: number

    constructor(timePeriods = 10, startingMean = 0) {
        this.mean = startingMean
        this.multiplier = 2 / (timePeriods + 1)
    }

    update(newValue: number) {
        const meanIncrement = this.multiplier * (newValue - this.mean)
        const newMean = this.mean + meanIncrement
        this.mean = newMean
    }
}

const movingAverage = new ExponentialMovingAverage(60*3)

// in render loop
movingAverage.update(delta)
const avgDelta = movingAverage.mean
```

packages/engine/src/renderer/WebGLRendererSystem.ts
```diff
@@ -272,18 +287,19 @@ export class EngineRenderer {
...
-    const averageDelta = this.calculateMovingAverage(delta)
+    // const averageDelta = this.calculateMovingAverage(delta)
+    const averageDelta = this.onUpdateMovingAverage(delta)
 
     // dont downgrade when scene is still loading in
-    if (useWorld().elapsedTime > 5) {
+    if (Engine.sceneLoaded) {
       if (averageDelta > this.minRenderDelta) {
-        this.qualityLevel--
+        this.qualityLevel -= this.qualityLevelSensitivity
```
speigg, line 296: Is this still necessary w/ the updated moving average logic?

packages/engine/src/renderer/WebGLRendererSystem.ts
```diff
@@ -122,6 +124,10 @@ export class EngineRenderer {
   timeSamples = new Array(60 * 1).fill(1000 / 60) // 3 seconds @ 60fps
   index = 0
 
+  averageMean: number = 0
+  averageMultiplier: number = 0
+  averageTimePeriods: number = 60
```
speigg, line 129: I don't think 1 second is enough, probably need 3-5 seconds

packages/engine/src/renderer/WebGLRendererSystem.ts
```diff
@@ -255,13 +265,18 @@ export class EngineRenderer {
...
     this.averageFrameTime =
       (this.averageFrameTime * this.timeSamples.length + delta - this.timeSamples[this.index]) / this.timeSamples.length
-
     this.timeSamples[this.index] = delta
     this.index = (this.index + 1) % this.timeSamples.length
-
     return this.averageFrameTime
   }
 
+  onUpdateMovingAverage(newValue: number): number {
```
speigg, line 273: I think we should just put the moving average logic in a utility class like the one I provided; it will likely be used in multiple places, and it's very self-contained

speigg (changes requested):

rondoor124:
Hi @speigg 
Thanks for the quick feedback. I updated the ExponentialMovingAverage into common class.

rondoor124:
Hi, @speigg @HexaField
Please let me know the details to improve this feature more.
Thank you!

Issue #4198, opened 2021-11-22
Git Authentication Oauth


HexaField:
We want Github app access for projects and the editor to enable proper version control and repo access on deployment. 
https://docs.github.com/en/developers/apps/getting-started-with-apps/differences-between-github-apps-and-oauth-apps

We need to be able to allow users to link private repositories such that they can be cloned, pulled, add commits, pushed etc with user credentials specific to their xrengine auth session.

PR #4423, opened 2021-12-09
Github App OAuth #4198

rondoor124:
This is the prototype of the Git Authentication Oauth https://github.com/XRFoundation/XREngine/issues/4198

**How to test**
- Reinstall node module
- Download the Pem file from the GitHub app and add it to the certs folder as git.pem.


HexaField:
This PR also adds a `gitPem` column to `server-settings` table

.env.local.default
```diff
@@ -119,7 +120,8 @@ FACEBOOK_CLIENT_ID=262344435023143
 FACEBOOK_CLIENT_SECRET=
 
 GITHUB_CALLBACK_URL=https://localhost:3030/auth/oauth/github
-GITHUB_CLIENT_ID=4b3e9e2c5fca8073809a
+GITHUB_APP_ID=156701
+GITHUB_CLIENT_ID=Iv1.2b4c5a211bf7ab05
```
HexaField, line 124: is this supposed to have changed?

packages/server-core/src/githubapp/githubapp/githubapp.class.ts
```diff
@@ -0,0 +1,44 @@
...
+    const repo = await getGitHubAppRepos()
+    return repo as any
+  }
+
+  async get(id: number): Promise<any> {
+    const repo = getGitRepoById(id)
+    return repo as any
+  }
+
+  async create(): Promise<any> {}
```
HexaField, line 31: lets remove these override functions if they are unused

packages/server-core/src/githubapp/services.ts
```diff
@@ -0,0 +1,3 @@
+import GithubApp from './githubapp/githubapp.service'
```
HexaField, line 1: rather than creating a new `githubapp` folder lets move the sevice and class to `/server-core/src/projects/githubapp/...` instead

packages/server-core/src/projects/project/project.class.ts
```diff
@@ -220,8 +221,11 @@ export class Project extends Service {
     })
     if (existingProjectResult != null) await super.remove(existingProjectResult.id, params)
 
+    const repoPath = await getAuthenticatedRepo(data.url)
```
HexaField, line 224: we should still allow public repos to be cloned if unauthenticated. lets let this fail gracefully and notify the user they are unauthorised if the repo is private.

HexaField (changes requested):

HexaField:
We should be able to store the PEM in the database entry provided the column has sufficient length.
https://stackoverflow.com/questions/49839814/saving-a-pem-certificate-in-a-sql-server-database

.env.local.default
rondoor124: Yes. we need this information to get the GitHub app on the [service](https://github.com/XRFoundation/XREngine/pull/4423/files#diff-33f0167016721cfe6302b67b36443ff8f9850da6336a2bb2d15523791450fc5eR10).
```
const app = new App({
  appId: config.authentication.oauth.github.appid,
  privateKey: config.server.gitPem,
  oauth: {
    clientId: config.authentication.oauth.github.key,
    clientSecret: config.authentication.oauth.github.secret
  }
})
```

packages/server-core/src/githubapp/githubapp/githubapp.class.ts
rondoor124: I tried to remove the unused override functions, but it shows this error:
```
Class 'GitHubAppService' incorrectly implements interface 'ServiceMethods<Data, Partial<Data>>'.
  Type 'GitHubAppService' is missing the following properties from type 'ServiceMethods<Data, Partial<Data>>': create, update, patch, removets(2420)
```

packages/server-core/src/githubapp/services.ts
rondoor124: Yes. got it!

packages/server-core/src/githubapp/githubapp/githubapp.class.ts
HexaField: ok dont worry about it then

packages/server-core/src/updateAppConfig.ts
```diff
@@ -17,8 +17,52 @@ const nonFeathersStrategies = ['emailMagicLink', 'smsMagicLink']
...
+      console.warn('[updateAppConfig]: Failed to read serverSetting')
+      console.warn(e)
+    })
+  promises.push(serverSettingPromise)
+  await Promise.all(promises)
+}
+
 export const updateAppConfig = async (): Promise<void> => {
-  if (process.env.APP_ENV === 'development') return
+  // if (process.env.APP_ENV === 'development') return
```
HexaField: please revert this comment

packages/server-core/src/projects/githubapp/githubapp-helper.ts
```diff
@@ -39,6 +40,7 @@ export const createGitHubApp = () => {
 
 export const getGitHubAppRepos = async () => {
   try {
+    await refreshAppConfig()
```
HexaField, line 43: rather than calling this here, we should listen to the `server-setting` feathers service patch calls and update then, rather than lazily when `getGitHubAppRepos` is called

HexaField (changes requested):

packages/server-core/src/appconfig.ts
```diff
@@ -192,6 +193,7 @@ const authentication = {
       secret: process.env.FACEBOOK_CLIENT_SECRET!
     },
     github: {
+      appid: process.env.GITHUB_APP_ID!,
```
zulqarnainhanif, line 196: This is added here but doesn't seem to be displayed in `XREngine/packages/client-core/src/admin/components/Setting/Authentication.tsx` file

rondoor124: @zulqarnainhanif Yep. I just realized that we have an option to change the Authentication env on the admin page. Thank you so much!

barankyle (approved):

PR #4439, opened 2021-12-10
Loading Screen & XRUI example improvements

sheflywithwings:


zulqarnainhanif:
@speigg can you please take a look into the latest code i pushed in this PR. The state in engine service is working fine but XRUILoadingDetailView doesn't get updated with the state. Please let me know what might be the issue so i can fix this.

PR #4448, opened 2021-12-12
Refactor Database Seeding

HexaField:
## Summary

- [x] refactors the feathers seeder to be entirely async and removes unused functionality
- [x] seeder must have 'FORCE_DB_REFRESH=true` explicitly set to run
- [x] change deployment flow to do check for unseeded database in builder rather than server pods
- [x] default project is no longer read-only on deployment, as all logical operations and tests are not dependent on it

Adds a couple new commands to the root package.json
`check-db-exists` which queries the database for if the 'users' table exists, if it does, it exits, if it does not, it runs the command
`init-db-production` which runs the API server with `APP_ENV=production` `FORCE_DB_REFRESH=true` which will seed the database and exit

## Checklist

## QA Steps
```
docker container stop xrengine_db
docker container rm xrengine_db
docker container prune
npm run dev-docker
npm run check-db-exists
npm run install-projects
```
should destroy, recreate and seed database, then exit

## Reviewers

@barankyle

### DIFF
--- a/scripts/check-db-exists.ts
+++ b/scripts/check-db-exists.ts
@@ -1,0 +1,58 @@
+import dotenv from 'dotenv'
+import cli from 'cli'
+import { Sequelize } from 'sequelize'
+import { spawn } from 'child_process'
+
+dotenv.config()
+const db = {
+  username: process.env.MYSQL_USER ?? 'server',
+  password: process.env.MYSQL_PASSWORD ?? 'password',
+  database: process.env.MYSQL_DATABASE ?? 'xrengine',
+  host: process.env.MYSQL_HOST ?? '127.0.0.1',
+  port: process.env.MYSQL_PORT ?? 3306,
+  dialect: 'mysql'
+} as any
+
+db.url = process.env.MYSQL_URL ?? `mysql://${db.username}:${db.password}@${db.host}:${db.port}/${db.database}`
+
+cli.enable('status')
+
+const sequelize = new Sequelize({
+  ...db,
+  logging: true,
+  define: {
+    freezeTableName: true
+  }
+})
+
+cli.main(async () => {
+  await sequelize.sync()
+
+  const [results] = await sequelize.query("SHOW TABLES LIKE 'user';")
+
+  if (results.length === 0) {
+    console.log('User table not found, seeding the database...')
+
+    const initPromise = new Promise((resolve) => {
+      const initProcess = spawn('npm', ['run', 'init-db-production'])
+      initProcess.once('exit', resolve)
+      initProcess.once('error', resolve)
+      initProcess.once('disconnect', resolve)
+      initProcess.stdout.on('data', (data) => console.log(data.toString()))
+    }).then(console.log)
+
+    await Promise.race([
+      initPromise,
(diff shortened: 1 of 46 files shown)


scripts/check-db-exists.ts
```diff
@@ -0,0 +1,58 @@
...
+      initProcess.stdout.on('data', (data) => console.log(data.toString()))
+    }).then(console.log)
+
+    await Promise.race([
+      initPromise,
+      new Promise<void>((resolve) => {
+        setTimeout(() => {
+          console.log('WARNING: Initialisation too long to launch!')
+          resolve()
+        }, 3 * 60 * 10000) // timeout after 3 minutes - it needs to be this long as the default project is uploaded to the storage provider in this time
```
barankyle, line 50: 3 minutes should be 3* 60 * 1000

scripts/check-db-exists.ts
```diff
@@ -0,0 +1,58 @@
+import dotenv from 'dotenv'
+import cli from 'cli'
+import { Sequelize } from 'sequelize'
+import { spawn } from 'child_process'
+
+dotenv.config()
```
barankyle, line 6: Currently we only run dotenv.config if process.env.KUBERNETES is not true, i.e. not in a production environment where all the ENV_VARS are supposed to be set directly rather than provided by a .env file. `.env` is in the .dockerignore, so dotenv.config should normally do nothing, but for consistency maybe we should only conditionally run it on KUBERNETES !== true here as well.

HexaField: gotcha, i will also use dotenv-flow and copy the dotenv configuration params from appconfig.ts for consistency

barankyle (approved):

barankyle (changes requested):
I was trying it out and had some minor concerns about being able to force a reinit after the database is already seeded. check-db-exists only checks for user table existence; even if you set FORCE_DB_REFRESH to true on the builder, it won't do a reinit unless `user` doesn't exist. I'm not sure how often anyone's going to be running a reinit on their own, but I feel like that should be an option.

barankyle:
Also, if locally I run `check-db-exists` on a clean database, which seeds the database, then `install-projects`, the default project is not being installed. I'm not sure if this would also happen in production, but I would expect a local run to install default-project.

HexaField:
> I was trying it out and had some minor concerns about being able to force a reinit after the database is already seeded. check-db-exists only checks for user table existence; even if you set FORCE_DB_REFRESH to true on the builder, it won't do a reinit unless `user` doesn't exist. I'm not sure how often anyone's going to be running a reinit on their own, but I feel like that should be an option.

Reinit should still be available by forcing in the api server. The check-db-production command executes that after all.

> Also, if locally I run check-db-exists on a clean database, which seeds the database, then install-projects, the default project is not being installed. I'm not sure if this would also happen in production, but I would expect a local run to install default-project.

Right yes, since we have an implicit install for the default project when the api server runs in dev mode, we need to seed the database still in production. Making these changes now.

barankyle (approved):

barankyle (approved):

PR #4515, opened 2021-12-19
git clone command for getting started has incorrect formatting 

seang:
## Summary
The two dash characters used aren't getting correctly parsed on copy and run, replacing with standard dash characters

## Reviewers


HexaField (approved):
Thank you!

NateTheGreatt (approved):

PR #4774, opened 2022-01-13
Spawnpoint skybox redundancy fix

dinomut1:
## Summary

Fixed error caused when loading skybox or spawn point components from gltf models. Added a simple check to see that an object3d has not already been created for a given entity to prevent redundancy.

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers


package.json
```diff
@@ -113,5 +113,8 @@
     "trace-unhandled": "2.0.1",
     "ts-node": "10.4.0",
     "ts-node-dev": "1.1.8"
+  },
+  "dependencies": {
```
HexaField, line 117: this should be added to dev dependencies rather than normal dependencies

HexaField (changes requested):

HexaField:
I believe this is a misuse of the scene format. Each entity / object should have a specific function, if the intention here is to be able to have a single object contain the skybox and the spawn position information, they should be split into multiple files. If this is a bug, we should find out why this is occuring rather than adding a quick fix.

PR #4783, opened 2022-01-14
Fix Scene Loader Tests

HexaField:
## Summary

Updates all the scene loader tests that were disabled, solving a few bugs along the way.

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers


packages/engine/src/scene/functions/loaders/DirectionalLightFunctions.ts
```diff
@@ -61,7 +66,7 @@ export const deserializeDirectionalLight: ComponentDeserializeFunction = (
 
 export const updateDirectionalLight: ComponentUpdateFunction = (entity: Entity) => {
   const component = getComponent(entity, DirectionalLightComponent)
-  const light = getComponent(entity, Object3DComponent)?.value as DirectionalLight
+  const light = component.light
```
NPatel10, line 69: Do you think this will be a good way or this is for some special purpose? We can also do this to nearly all other components types.

NPatel10 (approved):

packages/engine/src/scene/functions/loaders/DirectionalLightFunctions.ts
HexaField: Since it's a reference to an object and not a value itself, it's probably fine to do this, as there is still a single source of truth. 

The problem in this case is that CSM introduces some complexities with how the directional light rendering works, so we have to store it on the directional light component instead of the object3d component.

PR #4796, opened 2022-01-15
lightmap import fix

dinomut1:
## Summary
Fixed a race condition in the loading of GLTF assets with lightmaps causing them to partially or not load lightmaps. Changed the onload callback for gltfs to be async and wait for the lightmaps to be loaded and set onto asset materials before continuing. Previously the GLTFParser would have its materials changed to have the lightmaps, but the actual materials on the meshes themselves were never being updated to match the parser. Parity is now enforced within the onload callback. 

Also refactored a misspelling of extension.

## Checklist

## References

## QA Steps

Load any scene with gltf lightmapped assets in the editor or client. The lightmaps now load correctly when they would exhibit undefined behavior before.

## Reviewers


HexaField (commented):
Looks good to me, could you please write a basic unit test for `loadLightmaps` to prevent future regressions?

packages/engine/tests/assets/functions/loadGLTF.test.ts
```diff
@@ -0,0 +1,17 @@
+import assert from "assert"
+
+import {GLTF, GLTFLoader} from "../../../src/assets/loaders/gltf/GLTFLoader"
+import {AssetLoader} from "../../../src/assets/classes/AssetLoader"
+import {Mesh, Material, MeshBasicMaterial, Texture} from "three"
+
+describe("LoadGLTF", () => {
+    describe("LoadLightmaps", () => {
+        it("should load lightmapped gltf", async () => {
+            AssetLoader.load({url:"assets/lightmap_test.glb"}, async (gltf:GLTF) => {
```
HexaField, line 10: while this is fine, an integration test is not really what we need, i think we can avoid the bloat of a glb file by just unit testing the `loadLightmaps` function by mocking some GLTF data

HexaField (changes requested):

packages/engine/tests/assets/functions/loadGLTF.test.ts
dinomut1: Mocking up gltf data for this test is extremely complex and I don't feel confident I can mock up data that correctly replicates the state of a gltf passed to loadLightmaps.

loadLightmaps works with both the loaded scene hierarchy in the gltf.scene object and parsed json in the GLTFParser object gltf.parser, along with loading indexed texture dependencies that are created during an import. In order to mock this data up I'll end up needing to write a GLTF file anyways for the texture dependencies to be set up right. Loading a glb file automatically sets up the scene, parser and dependencies to the right state for the test to be done. 

I can certainly make the glb file smaller though.

HexaField: No stress. 😊

PR #5178, opened 2022-02-12
fix entity undefined in data reader

HexaField:
## Summary

Fixes a crash where a pose packet can come after a user entity has been destroyed.

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_

### DIFF
--- a/packages/engine/src/networking/serialization/DataReader.test.ts
+++ b/packages/engine/src/networking/serialization/DataReader.test.ts
@@ -449,6 +449,55 @@ describe('DataReader', () => {
     strictEqual(TransformComponent.rotation.y[entity], 0)
     strictEqual(TransformComponent.rotation.z[entity], 0)
     strictEqual(TransformComponent.rotation.w[entity], 0)
+
+    // should update the view cursor accordingly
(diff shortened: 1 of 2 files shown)


hamzzam:
image
  Seems to be still occurring, I had tried spending some time on this too. I think this happens when a non-avatar object is also present in the packet but the offset is always set according avatar entity data. But not sure mostly a guess at this point .

HexaField:
> Seems to be still occurring, I had tried spending some time on this too. I think this happens when a non-avatar object is also present in the packet but the offset is always set according avatar entity data. But not sure mostly a guess at this point .

Ah, this is a different error, I am still unsure why this one is occuring.

hamzzam:
> > Seems to be still occurring, I had tried spending some time on this too. I think this happens when a non-avatar object is also present in the packet but the offset is always set according avatar entity data. But not sure mostly a guess at this point .
> 
> Ah, this is a different error, I am still unsure why this one is occuring.

Oh ok

PR #5348, opened 2022-02-26
Replace EditorEvents with Hookstate

zulqarnainhanif:
## Summary

- Replace all the editor events with hookstate

## Checklist

## References

https://github.com/XRFoundation/XREngine/issues/4340

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

@HexaField


packages/editor/src/services/ModeServices.ts
```diff
@@ -0,0 +1,102 @@
...
+  transformPivotModeChanged: boolean
+  snapSettingsChanged: boolean
+}
+
+const state = createState<ModeServiceStateType>({
+  renderModeChanged: false,
+  playModeChanged: false,
+  flyModeChanged: false,
+  transformMode: '',
+  transformModeChanged: false,
```
NPatel10, line 21: Since there is `transformMode` state, having `transformModeChanged` is redudent in my opinion. We can use value of `transformMode` to determine the change which was not possible before this refactor hence we had 2 properties.

NPatel10 (commented):
Everything looks good. At one place I like to draw your attention. Let me know your thought on that

zulqarnainhanif:
@NPatel10 @HexaField 2 issues left in this PR which i will be working at night.
- Heirarchy Tree isn't updating and showing (once you expand its view it shows up)
- Loading another scene over current scene creates the same problem we were facing before (might be some problem with scene not being unloaded properly etc)

packages/editor/src/services/ModeServices.ts
zulqarnainhanif: Will update this flow, Thanks.

NPatel10 (approved):

speigg:
One more thing; I added `useHookedEffect()`, which should be used for creating effects with hookstate deps. The issue being that, you don't want to pass `state.value` as a dep if `state.value` refers to an object (rather than a primitive). Passing `state` as a dep is correct, but if `state.value` is never used in the component, then hookstate dependency tracking won't cause that component to rerender when that state changes. `useHookEffect()` fixes this.

zulqarnainhanif:
> One more thing; I added `useHookedEffect()`, which should be used for creating effects with hookstate deps. The issue being that you don't want to pass `state.value` as a dep if `state.value` refers to an object (rather than a primitive). Passing `state` as a dep is correct, but if `state.value` is never used in the component, then hookstate dependency tracking won't cause that component to rerender when that state changes. `useHookEffect()` fixes this.

@speigg  Can you please point out with an example if such changes needs to be done in my PR.

speigg:
@zulqarnainhanif  In EditorContainer.ts, you added 

```ts
  useEffect(() => {
     if (editorError) {
       onEditorError(editorError.value)
     }
   }, [editorError])
   ```
   
This should be `useHookedEffect`, otherwise the component won't re-render when the editorError state changes. This is the only case where it's strictly necessary, but it doesn't hurt to use `useHookedEffect` by default (especially if using hookstate deps).

PR #5350, opened 2022-02-28
Update 01-installation.md

michaelsharpe:
## Summary

Just a small change to the installation instructions. Building the image fails on a new pull of the repo since you cant commit an empty folder.  So an empty folder for `project-package-jsons` needs to be created for the image build to succeed.  There might be a better way to fix this, like a default file in that folder.  However I wanted to raise this as a flag as I walk through installing the engine.


HexaField:
Thank you!

Issue #4857, opened 2022-01-20
Asset Loader cache issue - keeps reference of the object


NPatel10:
Asset loader cache is storing the object reference without cloning it so any operation further down the time is being reflected in the cache. Furthermore, upon creating second object from cache it is same object as first one.

Also there is Texture cache and GLTF cache in editor, find out if those can be replace with Asset Loader. If so then probably Cache Manager is also not required.

PR #5379, opened 2022-03-02
added deep copy when getting the cached asset

FabriceIRANKUNDA:
## Summary

## Checklist

## References
#4857 

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_


packages/editor/src/caches/GLTFCache.ts
```diff
@@ -8,7 +8,7 @@ export default class GLTFCache {
   get(url) {
     const absoluteURL = new URL(url, (window as any).location).href
     if (this.cache.has(absoluteURL)) {
-      return this.cache.get(absoluteURL)
+      return new Map(JSON.parse(JSON.stringify([...this.cache]))).get(absoluteURL)
```
NPatel10, line 11: Making a whole new map is not great because it will make copy of all assets loaded in cache. But we only need copy of requested object. Also using `JSON` to make copy of asset is not desirable because `JSON` only serialize non functional properties. So all the methods which were in the assets will be gone. So find other way like, `Object.assign` or `clone` method of asset if available (check for deep cloning).

NPatel10 (changes requested):
Making a whole new map is not great because it will make copy of all assets loaded in cache. But we only need copy of requested object. Also using JSON to make copy of asset is not desirable because JSON only serialize non functional properties. So all the methods which were in the assets will be gone. So find other way like, Object.assign or clone method of asset if available (check for deep cloning).

PR #5409, opened 2022-03-04
Project Configurable Env Settings

zulqarnainhanif:
## Summary

Add project settings seeded from xrengine.config.ts of each project and allow updating from Admin Panel Settings

This PR will add a column to Project Table named `settings` which can store and array of objects ({key: "", value: ""}). The key is the name of the setting and value can consist of any any string value you want stored against it.
The Admin Panel will have a `Project Tab` in `/settings` route now which will show settings config loaded from xrengine.config.ts , which then can be updated and stored in the database against each project.
By default the Settings column of Project Table will be hidden from `project service` but can be accessed through `project-settings service` like its used in the admin panel routes. Only admins currently have access to these api but later scopes can be used to allow fetching them according to the scehma available with each project.

## Checklist
- [x] Changes reviewed by at least 2 approved reviewers

## References

https://github.com/XRFoundation/XREngine/issues/5183

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_


packages/client-core/src/admin/services/Setting/ProjectSettingService.ts
```diff
@@ -0,0 +1,72 @@
...
+//State
+export const PROJECT_PAGE_LIMIT = 100
+
+export interface ProjectSettingValue {
+  key: string
+  value: string
+}
+
+export const state = createState({
+  projectSetting: [] as Array<ProjectSettingValue>
```
speigg, line 15: this state can just be an array, e.g., 

```ts
createState([] as Array<ProjectSettingValue>)
```

packages/projects/ProjectConfigInterface.ts
```diff
@@ -67,4 +69,10 @@ export interface ProjectEventHooks {
   onUninstall?: InstallFunctionType
 }
 
+export interface ProjectSettingSchema {
+  key: string
+  type: string
+  scopes: Array<string>
```
speigg, line 75: not being used?

speigg (approved):
Looks great, nice job!

packages/projects/ProjectConfigInterface.ts
zulqarnainhanif: this will be used when we start adding code to check whether a person has access to any project setting or not

HexaField (approved):

PR #5436, opened 2022-03-08
Add a minimalist mode with some show/hide buttons on location UI

FabriceIRANKUNDA:
## Summary

Clean up buttons
Added animations and auto-hide 
Tested and enabled media button by default

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_


HexaField:
Screenshot from 2022-03-15 12-14-25

Mobile needs improvement.

HexaField:
Screenshot from 2022-03-15 12-15-47
bothbuttons need to be moved to the right a bit, such that the distance between the edge of the screen and the button are the same on both sides

HexaField (changes requested):
The state of the button should be saved to local browser data, such that when i reload the page it remembers whether i left it shown or hidden.

HexaField (changes requested):
Getting close, when the panels are hidden, they should also move in the opposite way to when they are shown, moving up or down. it also seems a bit slow, it should happen much faster, the same timing as the buttons rotating.

HexaField (changes requested):
- video, audio & face tracking automatically start without being clicked
- on mobile the screen is unclickable except for the middle third, the section holding all the media icons and settings icons is being clicked instead, this should click through to the canvas
- thumbpad to move is overlapping with the settings icons
- if the fullscreen button is not available, the refresh should sit over to the left so it looks right
- avatar thumbnail in top right can be made half the size

HexaField:
Screenshot from 2022-03-22 14-12-26
please fix this

PR #5468, opened 2022-03-12
Improve scene loading typings & fix loading race

speigg:
More accurate typings for scene data / metadata.
More accurate typings for scene service state, and more understandable action names.
Ensure that scene data is available before systems can initialize.
Move loading progress logic into EngineService receptor.
Use useHookedEffect in LoadEngineWithScene and NetworkInstanceProvisioning.

## Summary

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_


packages/engine/src/ecs/functions/SystemFunctions.ts
```diff
@@ -30,26 +30,37 @@ export type SystemInstanceType = {
...
+            console.warn(`Long system execution detected. System: ${name} \n Duration: ${duration}`)
+          }
         }
-      }
-    } as SystemInstanceType
+      } as SystemInstanceType
+    } catch (e) {
+      console.error(`System ${name} failed to initialize! `)
+      console.error(e)
+      return {
```
HexaField, line 57: should we return an empty system? or return null and filter it out?

packages/engine/src/ecs/classes/EngineService.ts
```diff
@@ -55,11 +55,25 @@ export function EngineEventReceptor(action: EngineActionType) {
...
-        return s.merge({ joinedWorld: true })
+        return s.merge({
+          sceneLoaded: false,
+          sceneLoading: true,
+          loadingProgress: 0,
+          loadingDetails: 'common:loader.loadingObjects'
+        })
+      case EngineEvents.EVENTS.SCENE_LOADED: {
+        const message = s.joinedWorld.value ? 'common:loader.loadingComplete' : 'common:loader.joiningWorld'
+        return s.merge({ sceneLoaded: true, sceneLoading: false, loadingProgress: 100, loadingDetails: message })
```
HexaField, line 66: i would advise against pushing i18n into the engine package, this is a property of the client rather than the engine itself, we should have a LoadingStage enum or something that the client can then interpret

packages/client-core/src/components/World/LocationLoadHelper.tsx
```diff
@@ -32,13 +33,12 @@ import { updateNearbyAvatars } from '@xrengine/engine/src/networking/systems/Med
 import { loadSceneFromJSON } from '@xrengine/engine/src/scene/functions/SceneLoading'
 import { getSystemsFromSceneData } from '@xrengine/projects/loadSystemInjection'
 
-export const retriveLocationByName = (authState: any, locationName: string, history: any) => {
-  if (
-    authState.isLoggedIn?.value === true &&
-    authState.user?.id?.value != null &&
-    authState.user?.id?.value.length > 0
-  ) {
+import { AuthState } from '../../../src/user/services/AuthService'
```
HexaField, line 36: shorten import

HexaField (changes requested):

packages/engine/src/ecs/classes/EngineService.ts
speigg: Good idea

packages/engine/src/ecs/functions/SystemFunctions.ts
speigg: We can filter it out, sure

HexaField (approved):

PR #5482, opened 2022-03-14
Fixed a bug with minikube client local file loading.

barankyle:
## Summary

getCachedAsset.ts always uses cacheDomain for generating URLs. On minikube, though, when using local storage,
the client and server need different addresses. The server needs to use 10.0.2.2 to punch out of minikube to
the host machine's localhost, while the client just needs to use localhost.

New logic assumes that localStorageProvider will be 10.0.2.2, and so alterations to that are done if new parameter
'clientFetch' is explicitly true (defaults to false). scene.class.ts now accepts query parameter clientFetch and passes
it on.

Fixed a bug in adminLocations fetching that rebrands location_setting to locationSettings, as we do with non-admin locations.

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_


packages/client-core/src/world/services/SceneService.ts
```diff
@@ -24,7 +24,7 @@ export const useSceneState = () => useState(state) as any as typeof state
 
 export const SceneService = {
   fetchCurrentScene: async (projectName: string, sceneName: string) => {
-    const sceneData = await client.service('scene').get({ projectName, sceneName })
+    const sceneData = await client.service('scene').get({ projectName, sceneName, clientFetch: true })
```
HexaField, line 27: clientFetch through feathers api could instead of isInternal in params i think

packages/server-core/src/media/storageprovider/getCachedAsset.ts
```diff
@@ -1,4 +1,7 @@
-export const getCachedAsset = (path: string, cacheDomain: string) => {
+import config from '../../appconfig'
+
+export const getCachedAsset = (path: string, cacheDomain: string, clientFetch = false) => {
   if (!cacheDomain) throw new Error('No cache domain found - please check the storage provider configuration')
+  if (config.server.storageProvider === 'local' && clientFetch) cacheDomain = 'localhost:8642'
```
HexaField, line 5: we must also check for `config.kubernetes.enabled` here

HexaField (changes requested):
we should try to avoid passing clientFetch aroundas much as this

HexaField (approved):
cool

PR #5512, opened 2022-03-17
Miscellaneous VR bug fixes

HexaField:
## Summary

- physics debug not showing in VR
- VR camera layers not being set up properly
- controller error on server
- stop crash when spamming VR button

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_

### DIFF
--- a/packages/engine/src/xr/functions/WebXRFunctions.ts
+++ b/packages/engine/src/xr/functions/WebXRFunctions.ts
@@ -109,13 +109,15 @@ export const proxifyXRInputs = (entity: Entity, inputData: XRInputSourceComponen
  */
 
 export const startWebXR = async (): Promise<void> => {
+  const world = useWorld()
+  if (hasComponent(world.localClientEntity, XRInputSourceComponent)) return
+
   const container = new Group()
   const head = new Group()
   const controllerLeft = new Group()
   const controllerRight = new Group()
   const controllerGripLeft = new Group()
   const controllerGripRight = new Group()
-  const world = useWorld()
 
   removeComponent(world.localClientEntity, FollowCameraComponent)
   container.add(Engine.camera)
--- a/packages/engine/src/xr/functions/WebXRFunctions.ts
+++ b/packages/engine/src/xr/functions/WebXRFunctions.ts
@@ -109,15 +109,13 @@ export const proxifyXRInputs = (entity: Entity, inputData: XRInputSourceComponen
  */
 
 export const startWebXR = async (): Promise<void> => {
-  const world = useWorld()
-  if (hasComponent(world.localClientEntity, XRInputSourceComponent)) return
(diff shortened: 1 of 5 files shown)


packages/engine/src/xr/functions/WebXRFunctions.ts
```diff
@@ -109,13 +109,15 @@ export const proxifyXRInputs = (entity: Entity, inputData: XRInputSourceComponen
  */
 
 export const startWebXR = async (): Promise<void> => {
+  const world = useWorld()
+  if (hasComponent(world.localClientEntity, XRInputSourceComponent)) return
```
mohsenheydari, line 113: I think this should probably be handled in upper calling layers (e.g `startXRSession`)  if redundant function calls are concerned. There are more function calls there that should probably not be fired twice per session.

packages/engine/src/xr/systems/XRSystem.ts
```diff
@@ -30,7 +31,15 @@ const startXRSession = async () => {
     dispatchLocal(EngineActions.xrSession() as any)
 
     // Current WebXRManager.getCamera() typedef is incorrect
-    ;(Engine.xrManager as any).getCamera().layers.enableAll()
+    // @ts-ignore
+    const cameras = Engine.xrManager.getCamera() as ArrayCamera
+    cameras.layers.enableAll()
+    cameras.cameras.forEach((camera) => {
```
mohsenheydari, line 37: Not very sure but shouldn't we restore the camera layers to the previous state on `EngineEvents.EVENTS.XR_END`?

mohsenheydari, line 37: Looks like these layers are overridden after they are set, I did put the layer setup in a `setTimeout` and the avatar appeared again.

mohsenheydari, line 37: `cameras.cameras` array is empty at this point. The VR camera layers are also the reason that the avatar is not showing.
I made a fix by running the same code in a world receptor as it is running a few frames later. At that point, the camera array is already populated.


`  Engine.currentWorld.receptors.push((action: EngineActionType) => {
    switch (action.type) {
      case NetworkWorldAction.setXRMode.type:
        // Current WebXRManager.getCamera() typedef is incorrect
        // @ts-ignore
        const cameras = Engine.xrManager.getCamera() as ArrayCamera
        cameras.cameras.forEach((camera) => {
          camera.layers.disableAll()
          camera.layers.enable(ObjectLayers.Scene)
          camera.layers.enable(ObjectLayers.Avatar)
          camera.layers.enable(ObjectLayers.UI)
        })
        break
    }
  })`

PR #5566, opened 2022-03-26
Update threejs to r139

HexaField:
## Summary

- updates threejs from r136 to r139
- refactors all scene loader update functions to be synchronous, utilising asset loader precaching
- breaks and disables FXAA (see https://github.com/XRFoundation/XREngine/issues/5568)

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_


packages/engine/package.json
```diff
@@ -78,7 +78,7 @@
     "@types/offscreencanvas": "2019.6.4",
     "@types/proxyquire": "^1.3.28",
     "@types/sinon": "10.0.11",
-    "@types/three": "0.136.1",
+    "@types/three": "0.138.0",
```
NPatel10, line 81: Why it is different from `three` version?

HexaField: three types are always a few weeks behind, as it is managed entirely separately from threejs itself. there are no typing updates between these versions, so it is all fine.

packages/engine/src/scene/functions/loaders/AudioFunctions.ts
```diff
@@ -115,14 +113,13 @@ export const updateAudio: ComponentUpdateFunction = async (entity: Entity, prope
 
   if (properties.audioSource) {
     try {
-      const { url, contentType } = await resolveMedia(component.audioSource)
```
NPatel10, line 118: By removing `resolveMedia` we are relying on file extension to determine extension instead of content type. This can cause problems for URL sources with no extensions. If this is intentional then avoid this.

HexaField: yeah this was something i realised, since all files being uploaded in the editor will have a file extension, i'm not too worried, it's pretty standard now. it also enforces ensuring file extensions correspond to mime types, which is what we want.

NPatel10 (approved):

speigg (approved):

speigg:
Looks good!

PR #5580, opened 2022-03-29
Improve login by magiclink

FabriceIRANKUNDA:
## Summary
Updated link sent via email 
Improved magic link to be used once and before expiration

## Checklist

## References

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

_Reviewers for this PR_

### DIFF
--- a/packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
+++ b/packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
@@ -26,9 +26,9 @@ const AuthMagicLink = (props: Props): JSX.Element => {
       AuthService.loginUserByJwt(token, '/', '/')
     } else if (type === 'connection') {
       AuthService.loginUserMagicLink(token, '/', '/')
-      if (user !== null) {
-        AuthService.refreshConnections(user.id.value!)
-      }
+      // if (user !== null) {
+      //   AuthService.refreshConnections(user.id.value!)
+      // }
       // window.location.href = '/profile-connections'
     }
   }, [])
--- a/packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
+++ b/packages/client-core/src/user/components/MagicLink/AuthMagicLink.tsx
@@ -20,16 +20,16 @@ interface Props {
 const AuthMagicLink = (props: Props): JSX.Element => {
   const { token, type } = props
   const { t } = useTranslation()
-
+  const user = useAuthState().user
   useEffect(() => {
     if (type === 'login') {
       AuthService.loginUserByJwt(token, '/', '/')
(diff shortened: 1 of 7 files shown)


barankyle (approved):

barankyle:
@FabriceIRANKUNDA I fixed some issues I'd found. The major one was that, seemingly for a while (well before you started working on this), submitting an email/sms was calling AuthService.addConnectionByEmail/BySMS, which is not correct. Our authentication system supports have multiple identity providers associated with a single user, but we've never really implemented it to date. addConnectionBy<Service> is only used when adding something to an existing User, not when logging in from a Guest account.

I also changed up how loginUserMagicLink works. I have it doing a GET on the 'login' service, which is what was handling the magiclink logins before, only now it's being called from the front-end rather than been hit directly by the browser. This returns a JWT generated from the associated identity-provider, which is then passed into loginUserByJWT(), which will save the JWT into localStorage and actually log the user in. the loadUserData() call you were trying to do doesn't persist over refreshes, since the client is set up to use the JWT stored in localStorage to authorize all of the calls to the API.

I made 'login-token'.FIND disallowed externally again, as that was leaking the login_token table. 'login'.GET should be more secure about only providing a valid JWT when a valid login token is passed to it. I also got rid of what you'd written for login-token.FIND since it wasn't needed anymore; the only lookup we do in login.GET is a findOne using the model.

barankyle (approved):

FabriceIRANKUNDA:
> @FabriceIRANKUNDA I fixed some issues I'd found. The major one was that, seemingly for a while (well before you started working on this), submitting an email/sms was calling AuthService.addConnectionByEmail/BySMS, which is not correct. Our authentication system supports have multiple identity providers associated with a single user, but we've never really implemented it to date. addConnectionBy(Service) is only used when adding something to an existing User, not when logging in from a Guest account.
> 
> I also changed up how loginUserMagicLink works. I have it doing a GET on the 'login' service, which is what was handling the magiclink logins before, only now it's being called from the front-end rather than been hit directly by the browser. This returns a JWT generated from the associated identity-provider, which is then passed into loginUserByJWT(), which will save the JWT into localStorage and actually log the user in. the loadUserData() call you were trying to do doesn't persist over refreshes, since the client is set up to use the JWT stored in localStorage to authorize all of the calls to the API.
> 
> I made 'login-token'.FIND disallowed externally again, as that was leaking the login_token table. 'login'.GET should be more secure about only providing a valid JWT when a valid login token is passed to it. I also got rid of what you'd written for login-token.FIND since it wasn't needed anymore; the only lookup we do in login.GET is a findOne using the model.

Awesome, thanks.

PR #5604, opened 2022-03-31
HyperFlux, Engine.store, and Engine.currentWorld.store

speigg:
## Summary

**HyperFlux** — brings together various state management strategies in XREngine, in a way that makes it easy to introspect and test. This PR adds ENGINE and WORLD hyperflux stores w/ migrated action dispatch and receptors; states still need be deglobalized and registered into their respective stores. A separate hyperflux store for the client state management also remains to be implemented. 

The ENGINE store is _non-networked_, meaning actions are dispatched directly on the _**incoming**_ queue, and run on the Engine timer. 

```ts
createHyperStore({
    name: 'ENGINE',
    getDispatchId: () => 'engine',
    getDispatchTime: () => Engine.elapsedTime
})
// Engine timer callback:
const executeWorlds = (delta, elapsedTime) => {
  Engine.elapsedTime = elapsedTime
  ActionFunctions.applyIncomingActions(Engine.store)
  // ...
}
```

The WORLD store is _networked_, meaning actions are dispatched directly on the _**outgoing**_ queue, and run on the world's fixed tick.

```ts
createHyperStore({
    name: 'WORLD',
    networked: true,
    getDispatchId: () => Engine.userId,
    getDispatchTime: () => this.fixedTick, // world.fixedTick
    defaultDispatchDelay: 1
})
// IncomingActionSystem
import { applyIncomingActions } from '@xrengine/hyperflux'
export default async function IncomingActionSystem(world) {
  return () => {
    applyIncomingActions(world.store)
  }
}

```

The CLIENT store is (or will be, when it's implemented) _non-networked_, and runs on a setInterval. 

In any case, the appropriate store must be provided when dispatching an action:

```ts
dispatchAction(world.store, NetworkWorldAction.spawnAvatar({ parameters }))
  ```

Likewise when adding or removing receptors:
```ts
addActionReceptor(world.store, (a) =>
    matches(a).when(NetworkWorldAction.spawnObject.matches, (a) => recepted.push(a))
)
```

State objects can also be defined and registered on a store:

```ts
const PeerState = defineState('peers', () => {
    return [] // initial state
})

registerState(world.store, PeerState)

// get immutable state
const peerState = getState(world.store, PeerState)

// or, get mutable state (if and only if in a receptor function)
const mutablePeerState = getMutableState(world.store, PeerState)
```

All incoming, outoing, and historical actions are accessible on the `store.actions` object.


HexaField (approved):

HexaField (changes requested):
It seems like there is a bug where the users take a long time to join the world. The join world function in NetworkFunctions seems to be dispatching far into the future, so i am imagining there is an incorrect time calculation being done in the action dispatching. It is also peculiar that the time property in engine actions is showing as a float below 1, whereas the network actions are dispatching as an integer which looks like it corresponds to world tick. perhaps there is a mistake somewhere here.

Let's also get this fully tested before we merge it in to ensure we have no regressions.

speigg:
> It seems like there is a bug where the users take a long time to join the world. The join world function in NetworkFunctions seems to be dispatching far into the future, so i am imagining there is an incorrect time calculation being done in the action dispatching. It is also peculiar that the time property in engine actions is showing as a float below 1, whereas the network actions are dispatching as an integer which looks like it corresponds to world tick. perhaps there is a mistake somewhere here.
> 
> Let's also get this fully tested before we merge it in to ensure we have no regressions.



> It seems like there is a bug where the users take a long time to join the world. The join world function in NetworkFunctions seems to be dispatching far into the future, so i am imagining there is an incorrect time calculation being done in the action dispatching.

That's correct; I thought I pushed the fix last night before I went to sleep haha. 

> Let's also get this fully tested before we merge it in to ensure we have no regressions. 

Writing tests today

HexaField (approved):

PR #5710, opened 2022-04-11
Added local storage support for editor helpers

NPatel10:
## Summary

now state of the editor helpers are stored in localstorage. Currently supported helpers are,
- [x] Transform mode
- [x] Transform Space
- [x] Transform Pivot
- [x] Grid
- [x] Render Mode
- [x] Snap Tool
- [x] Node helper
- [x] Physics helper

## Checklist

## References
- [x] #5685 

## QA Steps

2. `npm install`
3. `npm run dev-reinit`
4. `npm run dev`

## Reviewers

@HexaField


HexaField (changes requested):

packages/editor/src/services/ModeServices.ts
```diff
@@ -38,35 +48,117 @@ const state = createState<ModeServiceStateType>({
...
+  gridVisibility: true,
+  gridHeight: 0,
+  nodeHelperVisibility: true,
+  physicsHelperVisibility: true
 })
 
+function storeEditorHelperData(): void {
+  if (typeof window !== 'undefined') {
+    localStorage.setItem(
+      globalThis.process.env['VITE_EDITOR_LOCAL_STORAGE_KEY'] || 'theoverlay-editor-store-key-v1',
```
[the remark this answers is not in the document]
NPatel10: I changed this to use `ClientStorage` from `engine` package.

packages/editor/src/services/EditorHelperState.ts
```diff
@@ -0,0 +1,269 @@
...
+  nodeHelperVisibility: true,
+  physicsHelperVisibility: true
+})
+
+export async function restoreEditorHelperData(): Promise<void> {
+  if (typeof window !== 'undefined') {
+    const s = {} as EditorHelperStateType
+
+    await Promise.all([
+      ClientStorage.get(EditorHelperKeys.TRANSFORM_MODE).then((v) => {
```
HexaField, line 59: will this wait for the setter or just the getter? do we need to wrap this in a promise that resolves when the set promise is fulfilled?

packages/engine/src/debug/systems/DebugHelpersSystem.ts
```diff
@@ -47,56 +42,19 @@ const quat = new Quaternion()
...
-  const helpersByEntity = {
-    viewVector: new Map(),
-    ikExtents: new Map(),
-    box: new Map<Entity, Box3Helper>(),
-    helperArrow: new Map(),
-    velocityArrow: new Map(),
-    navmesh: new Map(),
-    navpath: new Map()
-  }
+export const helpersByEntity = {
```
HexaField, line 45: we want to keep this in the system scope rather than global scope such that we can write tests more easily. we want to avoid mutable global scope where possible.

packages/engine/src/renderer/WebGLRendererSystem.ts
```diff
@@ -233,29 +236,12 @@ export class EngineRenderer {
...
-  }
 }
 
 export default async function WebGLRendererSystem(world: World) {
   new EngineRenderer()
 
-  receiveActionOnce(Engine.store, EngineEvents.EVENTS.JOINED_WORLD, () =>
-    EngineRenderer.instance.loadGraphicsSettingsFromStorage()
-  )
+  receiveActionOnce(Engine.store, EngineEvents.EVENTS.JOINED_WORLD, async () => await restoreEngineRendererData())
```
HexaField, line 244: no need to make this async, the function will be called synchronously anyway, and this restore function does not need to be awaited

HexaField (changes requested):

packages/editor/src/services/EditorHelperState.ts
NPatel10: This will not be awaited. Since setting the variables will occur only once when loading the engine all the promises will be fulfilled. Same this is happening during the state update in the receptors.

packages/engine/src/debug/systems/DebugHelpersSystem.ts
NPatel10: We can freeze the object or we have to revert back to previously implemented logic where receptors are in the `DebugHelperSystem`

HexaField: oh i see, yes let's move the receptors back and keep it scoped to the system

HexaField (approved):

Issue #5485, opened 2022-03-14
Add check and UI popup for unsupported or missing hardware acceleration


HexaField:
- only needed on windows
- detect consistent low framerate across a few seconds once world has joined using engine frame timer (World.ts)
- display modal saying to check GPU settings

PR #5735, opened 2022-04-13
Hardware acceleration

FabriceIRANKUNDA:
## Summary
Added a popuo UI to inform user to enabled hardware acceleration for better experience
## References
#5485 
closes #_insert number here_

## Checklist

## QA Steps

## Reviewers

_Reviewers for this PR_


package.json
```diff
@@ -120,6 +120,8 @@
     "typescript": "4.6.2"
   },
   "dependencies": {
+    "@xrengine/engine": "^0.4.13",
```
HexaField, line 123: please remove the `@xrengine/engine` dep from here, and move `detect-gpu` to /packages/engine/package.json

packages/client-core/src/common/components/WarningMessage/BrowserMessages.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+
+import Typography from '@mui/material/Typography'
+
+export const BrowserMessages = ({ styles, browser }) => [
```
HexaField, line 5: lets use an object here where the key is the browser type rather than an array

packages/client-core/src/common/components/WarningMessage/WarningMessage.tsx
```diff
@@ -0,0 +1,35 @@
...
+import MuiDialogActions from '@mui/material/DialogActions'
+import MuiDialogContent from '@mui/material/DialogContent'
+import MuiDialogTitle from '@mui/material/DialogTitle'
+
+import styles from '../index.module.scss'
+import { BrowserMessages } from './BrowserMessages'
+
+const WarningMessage = ({ setIsHardwareAccelerationEnabled }) => {
+  const browser = {
+    isFirefox: navigator.userAgent.indexOf('Firefox') != -1,
```
HexaField, line 15: this probably is not good enough, we need to have a more robust way of detecting which browser is being used

packages/client/src/pages/index.tsx
```diff
@@ -7,15 +8,25 @@ import {
...
 
 const ROOT_REDIRECT: any = globalThis.process.env['VITE_ROOT_REDIRECT']
 
 export const HomePage = (): any => {
   const { t } = useTranslation()
   const clientSettingState = useClientSettingState()
+  const [isHardwareAccelerationEnabled, setIsHardwareAccelerationEnabled] = useState(true)
   const [clientSetting] = clientSettingState?.client?.value || []
 
+  useEffect(() => {
```
HexaField, line 22: we want to put this in the location page, not the home page. 
https://github.com/XRFoundation/XREngine/tree/dev/packages/client-core/src/components/World/LoadEngineWithScene.tsx

HexaField (changes requested):

packages/client/src/pages/index.tsx
FabriceIRANKUNDA: Basically, I added the check on the home page because I thought that having hardware acceleration enabled will be also important when the user navigates to the editor as well, then I preferred to add a check on the home page instead of adding in both location and editor. what do you think?

PR #5947, opened 2022-05-03
Network xr hands

hamzzam:
## Summary

Networking of XR hands

## Checklist

## QA Steps
Open 2 instances of a test location.
Set avatar controller type to XR hands from the menu.
Launch VR mode. 
XR Hands movements should be synced across all sessions.

## Reviewers
@HexaField


packages/engine/src/avatar/AvatarSystem.ts
```diff
@@ -75,6 +77,16 @@ function avatarActionReceptor(action) {
...
 
           addComponent(entity, XRInputSourceComponent, inputData as any)
+
+          dispatchAction(
+            Engine.instance.store,
+            AvatarInputSettingsAction.setControlType(a.avatarInputControllerType as any)
+          )
+          // This is required because state is updated in the next frame and some functionality that requires updated controller type might run in the current frame.
+          // This component can be used at such places.
+          addComponent(entity, AvatarInputControllerTypeUpdatePendingComponent, {
```
HexaField, line 87: let's try and refactor around not needing this component, we should rely more heavily on actions than adding and removing components

HexaField (changes requested):

packages/engine/src/avatar/AvatarSystem.ts
speigg: "some functionality that requires updated controller type might run in the current frame" which functionality?
"This component can be used at such places." What places?

hamzzam: > "some functionality that requires updated controller type might run in the current frame" which functionality? "This component can be used at such places." What places?

Hey sorry for the confusion I have updated the comment now to be more understandable.

dmitrizagidulin:
PR Moved to https://github.com/XRFoundation/XREngine/pull/6111

PR #5983, opened 2022-05-06
Refactor/#na update installation docs

ashishguptasanu:
## Summary

Updated minikube setup docs
Added var & instruction for host.minikube.internal

## References

Not Available

## Checklist
- [x] Changes reviewed by at least 2 approved reviewers

## QA Steps

## Reviewers

_Reviewers for this PR_


barankyle (changes requested):
This isn't right. minikube should actually be placing the `host.minikube.internal` entry into the hostfile on its own, so maybe we don't need to be specifying that line. But the IP that `host.minikube.internal` points to is definitely not the output of `minikube ip`.

ashishguptasanu:
> This isn't right. minikube should actually be placing the `host.minikube.internal` entry into the hostfile on its own, so maybe we don't need to be specifying that line. But the IP that `host.minikube.internal` points to is definitely not the output of `minikube ip`.

During my installation, this automated entry was not there with Ubuntu 20. I had to do it manually, that's why I thought.

Issue #6004, opened 2022-05-09
new message dot on chat is always present, even when there are no new messages


Griffisaurus:
![image](https://user-images.githubusercontent.com/92340542/167472443-a70abd68-9388-46a2-a0d8-00d80b96a4d4.png)

PR #6025, opened 2022-05-10
hide new massage dot and added alert sound on new message

FabriceIRANKUNDA:
## Summary

## References
#6004 
closes #_insert number here_

## Checklist

## QA Steps

## Reviewers

_Reviewers for this PR_


packages/client-core/src/components/InstanceChat/index.tsx
```diff
@@ -117,8 +118,14 @@ const InstanceChat = (props: Props): any => {
...
 
   React.useEffect(() => {
-    if (sortedMessages && sortedMessages[sortedMessages.length - 1]?.senderId !== user?.id.value) {
+    if (
+      sortedMessages &&
+      sortedMessages[sortedMessages.length - 1]?.senderId !== user?.id.value &&
+      chatState.messageCreated.value
+    ) {
       setNoUnReadMessage(false)
+      const audio = new Audio(notificationAlertURL)
```
HexaField, line 127: we should see if we need to manually clean this up or if it will be handled with the garbage collector, or better yet if we can re-use the audio node

packages/common/src/constants/URL.ts
```diff
@@ -0,0 +1 @@
+export const notificationAlertURL = 'http://codeskulptor-demos.commondatastorage.googleapis.com/pang/pop.mp3'
```
HexaField, line 1: we should download this file and server it from /packages/client/public folder

HexaField (changes requested):

HexaField (approved):

HexaField (changes requested):
- notification plays when loading messages before a user has joined
- audio settings slider should influence the volume of the alert

we may want to handle this audio via the engine, so we have greater control over it, and spatialize it when we implement it in ethereal

HexaField (changes requested):
this is on the right path, but we need to separate these audio settings out of the render settings. instead of engine/src/renderer/EngineRendererState we need these new variables to be in engine/src/audio/AudioState

dmitrizagidulin:
PR Moved to https://github.com/XRFoundation/XREngine/pull/6114

PR #6035, opened 2022-05-11
Fixes for several A/V problems

barankyle:
## Summary

Many A/V listeners were being registered more than once, especially on reconnections,
which were not clearing old listeners. Listeners should all now be registered exactly
once at any given time.

Prior behavior for receiving list of producers was to re-create consumers for each one
while clearing the existing ones. This was unnecessarily resource-intensive. Consumers
are now created only if there isn't already one for that producer.

Fixed issue with transports not always being cleared on disconnect/reconnect, which
was causing some problems with creating new consumer on some quick reconnects.

Made hover-expansion of user A/V controls occur anywhere on the locationUser container.
Hover CSS just on the individual users left gaps between them that would cause everything
to shrink, which was annoying.

## References

closes #_insert number here_

## Checklist

## QA Steps

## Reviewers

_Reviewers for this PR_

### DIFF
--- a/packages/client-core/src/components/Layout/index.module.scss
+++ b/packages/client-core/src/components/Layout/index.module.scss
@@ -75,6 +75,18 @@ body {
     }
   }
 
+  @media (min-width: 769px) {
+    &:hover,
+    &:focus {
+      [class*="user-controls"] {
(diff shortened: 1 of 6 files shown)


HexaField (approved):
LGTM!

PR #6036, opened 2022-05-11
Faster Engine init & customizable injected systems

speigg:
Moved Engine initialization into the client root component (system initializations are still route-specific for now)

Move injected location systems into [locationName].tsx, via Engine.instance.injectedSystems

This allows custom routes to avoid including these systems and still rely on LoadEngineWithScene and LoadLocationScene helpers.

Moved engine state to ENGINE hyperflux store

**Most relevant changes for reviewers:**

- canvas is only created in main.tsx now (removed from adminRoutes.tsx, EditorPage.tsx, and [locationName].tsx)
- Engine - injectedSystems added
- LoadEngineWithScene - location systems moved to [locationName].tsx, via Engine.instance.injectedSystems 
- LoadLocationHelper - initClient() decoupled from scene state

Other stuff:
EngineService.ts -> EngineState.ts
accessEngineState -> getEngineState


packages/client/src/pages/location/[locationName].tsx
```diff
@@ -30,6 +32,19 @@ const LocationPage = () => {
   useEffect(() => {
     dispatch(LocationAction.setLocationName(locationName))
     AuthService.listenForUserPatch()
+
```
Comment (HexaField) [inline packages/client/src/pages/location/[locationName].tsx:35]: lets put these injected system in a helper function so we can easily modularise and extend this without having to duplicate changes in the offline route and any custom routes in projects that wish to use this too

HexaField (approved):

PR #6068, opened 2022-05-13
Cleanup time conventions and world time origins

speigg:
- Standardized and clarified time units with property name conventions:
Variables with *Time = UTC millisecond timestamp
All other time is expressed in seconds, with *Seconds naming convention
- Cleaned up Timer and timing logic, passing standard high-resolution timestamps everywhere rather than passing delta time or otherwise manipulated (scaled) time values


HexaField (approved):
nice work, LGTM

PR #6133, opened 2022-05-18
Changes to add support for IPFS storage provider

hanzlamateen:
## Summary

This PR adds support for IPFS storage provider. If IPFS is configured in cluster then it will use it by default. Else it will fallback to the one provided in environment.

## References

closes #_insert number here_

## Checklist

## QA Steps

## Reviewers

_Reviewers for this PR_


packages/server-core/src/media/storageprovider/ipfs.storage.ts
```diff
@@ -37,6 +38,16 @@ export class IPFSStorage implements StorageProviderInterface {
       .then((res) => res.type === 'directory')
       .catch(() => false)
   }
+  async getCachedAsset(assetPath: string): Promise<string> {
+    if (!this.cacheDomain) throw new Error('No cache domain found - please check the storage provider configuration')
+
+    const filePath = path.join(this._pathPrefix, assetPath)
+
+    return this._client.files
```
HexaField, line 46: we should use async here rather than returning the promise that has then and catch attached

also, why is it necessary that this uses the stat function? we should keep this synchronous and we do not need to check that the file exists, it is just a helper file to return a URL, we do not want to mutate the path if it cannot be found

HexaField (changes requested):

packages/server-core/src/media/storageprovider/ipfs.storage.ts
hanzlamateen: Well, I am using promise which in itself resolves or reject just like things have been done in `local.storage.ts`

I am making use of stats method to get `CID` which is used to access resources in IPFS MFS

HexaField (changes requested):
now that we have established it is too problematic to run projects entirely off the IPFS storage provide, lets revert to [01f2de7](https://github.com/XRFoundation/XREngine/pull/6133/commits/01f2de7a7b8545cdad088cae225aa131799a3ce8)

packages/analytics/src/start.ts
```diff
@@ -21,7 +21,7 @@ process.on('unhandledRejection', (error, promise) => {
 const analyticsServerPipe = pipe(configureSocketIO())
 
 export const start = async (): Promise<Application> => {
-  const app = createFeathersExpressApp(analyticsServerPipe)
+  const app = await createFeathersExpressApp(analyticsServerPipe)
```
dmitrizagidulin, line 24: Whoa, good catch! :) (How did this work before??)

packages/server-core/src/media/storageprovider/ipfs.storage.ts
```diff
@@ -0,0 +1,399 @@
...
+
+    // const decodedData = new TextDecoder().decode(chunksArray).toString();
+    // console.log(decodedData)
+
+    // const file=`https://ipfs.io/ipfs/${hash}`;
+    // const req = await fetch(file, {method:'HEAD'})
+    // console.log(req.headers.get('content-type'))
+
+    return {
+      Body: Buffer.from(chunksArray),
```
dmitrizagidulin, line 59: @HexaField @speigg - our getObject() interface is a full on buffer and not a stream? This seems potentially disastrous, given large enough files (that may overflow node memory constraints). 
But that's unrelated to this PR; I'd love to revisit that later.

packages/server-core/src/media/storageprovider/ipfs.storage.ts
```diff
@@ -0,0 +1,399 @@
...
+
+    return ''
+  }
+
+  async _forwardIPFS(podName: string, forward: k8s.PortForward, forwardPort: number): Promise<string> {
+    return new Promise((resolve) => {
+      const server = net.createServer((socket) => {
+        forward.portForward('default', podName, [forwardPort], socket, socket, socket)
+      })
+      server.listen(0, '127.0.0.1', () => {
```
dmitrizagidulin, line 227: Should this IP be hardcoded here?

dmitrizagidulin (approved):
:+1:  (only minor comments inline)

packages/server-core/src/media/storageprovider/ipfs.storage.ts
HexaField: this is how it communicates internally with k8s

HexaField (approved):

Issue #5531, opened 2022-03-19
Docker Backend and JS frontend for Ripple Grant (XRPL, IPFS), ILP & Coil


mrhegemon:
Add containers to helm chart and make sure they are working (rippled, go-ipfs, ilp something)

Add coil and ilp connector (?) to client

Might be best to break this down into smaller isssues!

xrpl docker
https://hub.docker.com/u/xrpllabsofficial
old?
https://twitter.com/xrpllabs/status/1431756688676241418

https://github.com/WietseWind/docker-rippled
https://github.com/WietseWind/docker-rippled-validator

ipfs docker - add file driver alt / no S3
https://docs.ipfs.io/how-to/run-ipfs-inside-docker/
https://github.com/yeasy/docker-ipfs
https://hub.docker.com/r/ipfs/go-ipfs/

coil
https://coil.com/creator/how-to-monetize

ilp
https://interledger.org/
https://hub.docker.com/u/interledgerjs
https://github.com/interledgerjs/ilp-connector

dmitrizagidulin:
Further notes from convo with Liam:
* Add a 'Enable Coil' settings in admin panel. Add a textbox for the coil address.

mrhegemon:
we need to make it configurable
enable ipfs as a file driver - deployment config
enable coil - an admin config
enable xrpl nodes  - deployment config

PR #6164, opened 2022-05-20
Coil Payment Pointer Integration

dmitrizagidulin:
Addresses issue #5531

* Adds Coil payment pointer env vars
* Adds a Coil section to Admin Settings screen
* Adds payment pointer to client html meta tag (for use with Coil)


dmitrizagidulin:
Adding the COIL_PAYMENT_POINTER env var displays it in the Admin Settings screen:

image

and adds it to the header `<meta>` tag (see screenshot at the bottom, above title):

image

PR #6167, opened 2022-05-21
Fixed some bugs in WebRTCFunctions.ts and project retrieval

barankyle:
## Summary

Despite checking that transport exists, apparently it can be undefined/null by the time
router checks are performed. This could lead to the whole media server crashing because it
was trying to read (null).internal. Added conditional check in each occurrence to prevent that.

Made handleSendTrack check if there's already a producer with the exact same appData, and close it
if so.

Added much higher default pagination on calls to project.find. We're normally using the results as
if they're the entirety of the projects in the database, so the default of 10 could easily cause
projects to not be found by that call.

## References

closes #_insert number here_

## Checklist

## QA Steps

## Reviewers

_Reviewers for this PR_


HexaField:
Can we instead selectively disable pagination in certain feathers calls? I believe we are doing this elsewhere in the codebase already.

speigg (approved):

PR #6203, opened 2022-05-25
hide interact button when interactable object is not available and fi…

### LINKED ISSUE #5971: On mobile, hide interact button unless interacting
The interact button should be hidden unless an interactable object is available to interact with. We can play with turning the interact button into a "jump" button when there is nothing to interact with. In either case, the yellow glow should be removed.

FabriceIRANKUNDA:
…xed some css animations

## Summary

## References

closes #5971 

## Checklist

## QA Steps

## Reviewers

_Reviewers for this PR_


packages/client-core/src/common/components/TouchGamepad/index.tsx
```diff
@@ -15,6 +19,9 @@ export const TouchGamepad: FunctionComponent<TouchGamepadProps> = () => {
     document.dispatchEvent(event)
   }
 
+  const entity = Engine.instance.currentWorld.scene.id as Entity
```
HexaField, line 22: we should use `Engine.instance.currentWorld.entityTree.rootNode.entity` instead of `Engine.instance.currentWorld.scene.id`

HexaField (changes requested):

HexaField (approved):

HexaField (changes requested):
an interactable is never added to the scene node, we should instead get the InteractorComponent on `Engine.instance.currentWorld.localClientEntity` which has focusedEntity which will have an InteractableComponent

HexaField (changes requested):
this is not quite right still, please re-read the previous comment

HexaField (changes requested):

HexaField:
It still seems like this is not working properly. The button should only show up if there is an object that can be interacted with, and should hide when there is none. It might just be a case of it not updating properly.
Screenshot from 2022-06-01 16-49-15
Screenshot from 2022-06-01 16-50-24

HexaField:
Screenshot from 2022-06-04 12-59-03
the button is showing up when not on mobile, please fix this

HexaField (approved):
great stuff, all working properly!

PR #6264, opened 2022-05-31
i18n.ts to new version

johnfelipe:
/root/XREngine/packages/client-core/src/i18n.ts
https://gist.github.com/johnfelipe/76ac9686f141bac72c92c02b40f8971b


packages/client-core/src/i18n.ts
```diff
@@ -1,20 +1,15 @@
-import admin from '../i18n/en/admin.json'
-import common from '../i18n/en/common.json'
-import editor from '../i18n/en/editor.json'
-import social from '../i18n/en/social.json'
-import user from '../i18n/en/user.json'
+import { merge } from 'lodash'
```
speigg, line 1: Do we need this? Can't we use Object.assign ?

speigg (changes requested):

packages/client-core/src/i18n.ts
johnfelipe: import { merge } from 'lodash'

	// @ts-ignore
	const resources = Object.entries(import.meta.globEager('../i18n/**/*.json')).reduce((obj, [key, translation]) => {
	  const [language, namespace] = key.replace('../i18n/', '').replace('.json', '').split('/')
	  const result = { [language]: { [namespace]: translation } }
	  return merge(result, obj)
	}, {} as any)

	export const getI18nConfigs = () => {
	  return {
		resources,
		namespace: Object.keys(resources['es'])
	  }
	}

speigg (approved):

PR #6368, opened 2022-06-07
Compress linear velocity

hamzzam:
## Summary

Compress linear velocity.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/engine/src/networking/serialization/Utils.ts
+++ b/packages/engine/src/networking/serialization/Utils.ts
@@ -70,9 +70,9 @@ export const getVector4IndexBasedComponentValue = (vector4: Vector4SoA, entity: 
 }
 
 // Used when compressing float values, where the decimal portion of the floating point value
-// is multiplied by this number prior to storing the result. Doing this allows
-// us to retain three decimal places.
-export const FLOAT_PRECISION_MULT = 1000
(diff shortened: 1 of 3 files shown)


packages/engine/src/networking/serialization/Utils.ts
```diff
@@ -81,14 +81,16 @@ export const FLOAT_PRECISION_MULT = 1000
 // giving more precision with the same number of bits.
 export const QUAT_MAX_RANGE = 1 / Math.sqrt(2)
 
+export const VEC3_MAX_RANGE = 1 / 32
```
speigg, line 84: Why is this 1/32?

hamzzam: Gaffer suggested that a speed of 32 meters per-second is a good starting point. I tested it out using our physics simulation test system and everything was working as expected.

HexaField: this seems fine for now, we can see if puttclub ends up having a ball speed faster than this and we can adjust as needed

speigg: We should rename this constant if it's strictly related to Velocity. Vec3_Max_Range is ambiguous

Issue #4296, opened 2021-11-28
Migrate from PhysX to Rapier


HexaField:
PR to be reviewed by @speigg - Gheric, please message him when this branch is ready for review.

https://rapier.rs/docs/user_guides/javascript/getting_started_js/
https://docs.rs/rapier3d/latest/rapier3d/
https://github.com/dimforge/rapier
https://github.com/dimforge/rapier.js

Each of the following should have associated tests:

- [x] a `createPhysicsWorld` function that does the same as the existing `Physics.createScene` but returns a physics world that is  populated to `Engine.instance.currentWorld.physics` - 1
- [x] helper function to create various rigidbody types - 2
- [x] helper function to add colliders to rigid bodies - 2
- [x] raycast helper functions - 2
- [x] wrap collision events with hyperflux action dispatch & receptors - 4
- [x] update physics system to work with rapier physics - 4
- [x] debug helpers derived from physics data - 4
- [x] create ground plane for scene loading - 2
- [x] collision groups - 1
- [x] avatar kinematic controller & moveAvatar - 4
- [x] box colliders - 2
- [x] basic scene loading works with sky-station - 4
- [x] remove physx stuff - 4
- [x] complete implementation of RapierPhysicsSystem - 2
- [x] trigger volumes - 2
- [x] update avatar tests - 2
- [x] update assets pipeline doc for rapier stuff

Follow up work
- [ ] equippables - 2
- [ ] add missing physics tests - 2

we can push colliders straight into the components without a subobject to optimize memory usage and references, we aren't this for most components, but there is no reason not to.

this can be wrapped in a helper function `createPhysicsCollider` that takes an entity & config and adds a rigidbody to the world & the ECS

```ts
const rigidBody = addComponent(entity, RigidbodyDynamicComponent, RigidBodyDesc.dynamic())`
```

hamzzam:
https://github.com/XRFoundation/XREngine/issues/5275

barankyle:
Per someone from the community discovering that PhysX doesn't work properly with Node 18.1 and newer, please make sure that Rapier does work with the latest version of Node (I'm guessing it does, just want to make sure) and also update docs/docs/1_installation/readme.md to remove the notes that were added about being restricted to < 18.1.

PR #6382, opened 2022-06-09
Physx to rapier

hamzzam:
## Summary

Migrate physics engine to Rapier3D.

## References

addresses #4296

## Checklist

## QA Steps


packages/engine/src/physics/classes/PhysicsRapier.ts
```diff
@@ -193,8 +193,9 @@ function createRigidBodyForObject(
 
 function removeCollidersFromRigidBody(entity: Entity, world: World) {
   const rigidBody = getComponent(entity, RigidBodyComponent)
-  for (let index = 0; index < rigidBody.numColliders(); index++) {
-    const collider = rigidBody.collider(index)
+  const numColliders = rigidBody.numColliders()
+  for (let index = 0; index < numColliders; index++) {
+    const collider = rigidBody.collider(0)
```
HexaField, line 198: careful here, should be using the index

HexaField (changes requested):

packages/engine/src/physics/classes/PhysicsRapier.ts
hamzzam: the collider() function indexes into an array whose size is decreased with each removal.
I was actually using the index originally but then updated in this commit (https://github.com/XRFoundation/XREngine/pull/6382/commits/8db8e382585648dd7933464015bf6bb92f01eabb)

HexaField: Ah, cool

PR #6412, opened 2022-06-13
Wallet login re-enable.

dmitrizagidulin:
## Summary
* Add in-browser Credential Handler API wallet support
* Add `VITE_LOGIN_WITH_WALLET` env var that controls Login with Wallet button, defaults to false
* Remove unused/deprecated `locationAdmins` references
* Add `did-auth` authentication strategy

## References

* Partly addresses https://github.com/XRFoundation/XREngine/issues/6453 (item 4)

## Checklist

Provided this doesn't break anything existing, I'd like to get this into `dev`, to continue building on it.


barankyle:
Why is OAuth login being disabled (on the front-end, anyway) when wallet login is enabled?

I feel like did-auth should be integrated into authentication settings so that it's on par with all of the others - it can be enabled/disabled from the admin settings page, and whether or not it's in use can just be read from the auth settings rather than needing VITE_LOGIN_WITH_WALLET.

dmitrizagidulin:
@barankyle - great question. The aim is to move the oauth login stuff to the wallet itself.

dmitrizagidulin:
@barankyle Though now that I say that, I realize that we might want to enable the wallet in Dev for testing. Which means we also prolly want to keep the oauth stuff. So, I'll remove the part that disables the social sign-ins.

barankyle (changes requested):
I would still like to see wallet auth integrated into authentication settings, so that its enabled/disabled status can be controlled from the admin panel and does not rely on an environment variable.

barankyle (approved):

mrhegemon (approved):

PR #6424, opened 2022-06-15
add simple material option to models. Trigger volumes can now be scaled

dinomut1:
Added "Use Simple Materials" toggle to each Model component. 
Trigger volume collision meshes are now able to be scaled.


packages/engine/src/scene/functions/loaders/TriggerVolumeFunctions.ts
```diff
@@ -24,6 +24,8 @@ export const deserializeTriggerVolume: ComponentDeserializeFunction = (
   entity: Entity,
   json: ComponentJson<TriggerVolumeComponentType>
 ): void => {
+  const xform = getComponent(entity, TransformComponent)
```
HexaField, line 27: we should keep the convention of camel case `<componentName>Component` we have elsewhere in the codebase for component instances

packages/engine/src/scene/functions/loaders/ModelFunctions.ts
```diff
@@ -66,6 +68,15 @@ export const updateModel: ComponentUpdateFunction = (entity: Entity, properties:
   if (typeof properties.textureOverride !== 'undefined') {
     overrideTexture(entity)
   }
+
+  if (properties.useBasicMaterial !== undefined) {
```
HexaField, line 72: this should be `if (typeof properties.useBasicMaterial == 'boolean' && properties.useBasicMaterial) {`

HexaField (changes requested):

packages/engine/src/scene/functions/loaders/ModelFunctions.ts
dinomut1: This is equivalent, and the next line is required for toggle functionality

Issue #6272, opened 2022-05-31
Shader Dev for Interior Scene: Water, Grass


dinomut1:
Water: Add water shader underneath caustics effect:
Base: water vertex shader for ripples. Then Caustics blended with water texture / sprites

Caustic Lighting Shader: Standard lit material, with environment map, instancing, and then blended caustic lighting on top of diffuse channel.

Grass: Instances mesh onto surface of target mesh
growing parameters. width, height, leaf texture, direction, curvature, randomness, wind parameters, density

dinomut1:
Basalt: Glow with texture
Remove platform and add transparent pedestal atop more basalts

God rays: find example, add to postprocessing stack

PR #6466, opened 2022-06-20
Instancing component

dinomut1:
Adds Instancing component to engine. 

Three sampling modes: 
Scatter: instances are scattered onto the target surface based on UV
Vertices: one instance per vertex on target surface
Nodes: one instance per transform of children of root node

Two instancing modes:
Grass: generates instanced banners with custom shader
Mesh: instances provided mesh to sample points

Tests + UX polish forthcoming

Addresses first part of #6272

### DIFF
--- a/packages/editor/src/components/properties/InstancingNodeEditor.tsx
+++ b/packages/editor/src/components/properties/InstancingNodeEditor.tsx
@@ -30,6 +30,8 @@ import {
 } from '@xrengine/engine/src/scene/functions/loaders/InstancingFunctions'
 import getFirstMesh from '@xrengine/engine/src/scene/util/getFirstMesh'
 
+import AcUnitIcon from '@mui/icons-material/AcUnit'
+
 import { PropertiesPanelButton } from '../inputs/Button'
 import { ImagePreviewInputGroup } from '../inputs/ImagePreviewInput'
 import InputGroup from '../inputs/InputGroup'
@@ -278,4 +280,5 @@ export const InstancingNodeEditor: EditorComponentType = (props) => {
   )
 }
 
+InstancingNodeEditor.iconComponent = AcUnitIcon
 export default InstancingNodeEditor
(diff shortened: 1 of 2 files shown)


NPatel10:
@dinomut1, I am not able to make instance of the grass at my end. If you can provide all the details necessary to produce grass it would be great for testing. Thanks.
SS1
SS

dinomut1:
> @dinomut1, I am not able to make instance of the grass at my end. If you can provide all the details necessary to produce grass it would be great for testing. Thanks. SS1 SS

@NPatel10 I think the ground plane is too large. I have only tested it using Model nodes as the target surface.

NPatel10:
@dinomut1, Above suggestions works, But now if I select spawn point it is giving me error suggesting there is no uvs. I think which is fine but after getting this error the component freezes and does not let me stage or restage anything.

dinomut1:
> @dinomut1, Above suggestions works, But now if I select spawn point it is giving me error suggesting there is no uvs. I think which is fine but after getting this error the component freezes and does not let me stage or restage anything.

@NPatel10 Added stricter checks for the target surface options check

NPatel10:
> @NPatel10 Added stricter checks for the target surface options check

This is fixed but it shows itself in the list and same problem occurs when selecting itself. And another one is the icon is not being showed in the hierarchy panel for instance node.

dinomut1:
> > @NPatel10 Added stricter checks for the target surface options check
> 
> This is fixed but it shows itself in the list and same problem occurs when selecting itself. And another one is the icon is not being showed in the hierarchy panel for instance node.

@NPatel10 good finds. The icon not showing up in hierarchy is a problem I do not know how to fix. The other component I've made (Asset component) also has no icon in the hierarchy. Low priority

I will debug the selecting itself as target and freezing bug

NPatel10 (approved):

NPatel10:
@dinomut1, I fixed the icon issue for instancing component.

PR #6474, opened 2022-06-22
Multiplayer in Editor MVP

### LINKED ISSUE #6080: changing instance viewed in the editor makes a world instance connection

### LINKED ISSUE #6081: avatars & dynamic objects can be seen in real time

HexaField:
## Summary

- [x] adds spectate mode instance provisioning logic to the editor based on dropdown selection
- [x] unscubscribing from a hyperflux topic deletes all action history from that topic
- [x] fixes various vulnerabilities with bugs with instance connection logic
- [x] text message events handling for spectate via custom world server system

This PR adds a spectate network connection mode that does not spawn an avatar nor notifies the text message channel of the location that a user has joined. It refactores various parts of hyperflux and instance connection logic, including patching some bugs and security vulnerabilities around connection.

Screenshot from 2022-06-22 12-12-46

## References
closes #6080
closes #6081

## Checklist

## QA Steps


barankyle (changes requested):
If I go to /editor as an admin, then open a project and go to a scene that has an instance running, I do not see that instance in the instance selector. Same thing if I go to /editor/<project_name>. It seems that the instances are only retrieved if I go directly to /editor/<project_name>/<scene_name>.

I tested on minikube with two users on separate instances of different scenes. It works, but the instance list not being up-to-date was an issue. I started observing the sky-station scene, then switched to the default scene. The instance list remained what it was for sky-station. I had to reload when the editor was on /editor/default-project/default to get the list of instances for the default scene.

If the admin is in the instance in another tab, they will not be able to load that instance in the editor, surely because of our restrictions on one connection to an instance per user. We may not want to address this now.

User A sees the '<User B> joined the layer' notification (not the message, the notification in the top right) when User B loads the instance in the editor.

HexaField:
> If I go to /editor as an admin, then open a project and go to a scene that has an instance running, I do not see that instance in the instance selector. Same thing if I go to /editor/<project_name>. It seems that the instances are only retrieved if I go directly to /editor/<project_name>/<scene_name>.

Yes, this is intentionally. The instances are retrieved by first retrieving all locations using the scene loaded, and then retrieving all instances of each of those locations, otherwise we'd have to iterate all scenes of a project as well.

> I tested on minikube with two users on separate instances of different scenes. It works, but the instance list not being up-to-date was an issue. I started observing the sky-station scene, then switched to the default scene. The instance list remained what it was for sky-station. I had to reload when the editor was on /editor/default-project/default to get the list of instances for the default scene.
Yes, the feathers event listener work we discussed still needs to be done, I will do this in a follow up PR.

> If the admin is in the instance in another tab, they will not be able to load that instance in the editor, surely because of our restrictions on one connection to an instance per user. We may not want to address this now.
I have thought about this, we will need a more sophistocated connection flow, and this can be done after the instance server lifecycle action/receptor refactor.

> User A sees the ' joined the layer' notification (not the message, the notification in the top right) when User B loads the instance in the editor.
This is likely due to subscribing to not being able to opt out of the notification event. We can add a check to `Engine.instance.isEditor` or something. Will do this in the follow up.

PR #6633, opened 2022-07-22
Add client logging API endpoint.

dmitrizagidulin:
* Adds a `/api/log` endpoint to server-core, which receives all log events from `client`, `engine`, and `hyperflux` and sends them to Elastic for aggregation (only when `APP_ENV !== 'development'`) 
* Fixes `server-core` log events not having a `component` property
* Adds a `multiLogger` export to `common`.
* Adds error logging to `NotificationService.dispatchNotify()` (only when `variant === 'error'`)

Usage:

```ts
import multiLogger from '@xrengine/common/src/logger'
const logger = multiLogger.child({ component: 'client-core:ClientSettingService' }) // or appropriate component name

logger.error(err)
logger.info('Message')
logger.info('Message %o', { interpolation: 'values' })
logger.info({merge: 'object'}, 'Message goes here')
logger.info({merge: 'object'}, 'Message goes here %o', { interpolation: 'values' })
```

To test sending to Elasticsearch:

Edit your `.env.local`, set `VITE_FORCE_CLIENT_LOG_AGGREGATE=true`.
(You can then see the log events in local Kibana http://localhost:5601/app/kibana )

### DIFF
--- a/packages/client-core/src/admin/components/Project/GithubRepoDrawer.tsx
+++ b/packages/client-core/src/admin/components/Project/GithubRepoDrawer.tsx
@@ -44,7 +44,6 @@ const GithubRepoDrawer = ({ open, project, onClose }: Props) => {
       setProcessing(false)
       onClose()
     } catch (err) {
-      console.error(err)
       NotificationService.dispatchNotify(err.message, { variant: 'error' })
       setProcessing(false)
       throw err
@@ -59,7 +58,6 @@ const GithubRepoDrawer = ({ open, project, onClose }: Props) => {
       setProcessing(false)
       onClose()
     } catch (err) {
-      console.error(err)
       NotificationService.dispatchNotify(err.message, { variant: 'error' })
       setProcessing(false)
       throw err
--- a/.env.local.default
+++ b/.env.local.default
@@ -22,6 +22,7 @@ APP_HOST=localhost:3000
 APP_PORT=3000
 
 # Vite variables - Exposed to client side
+VITE_FORCE_CLIENT_LOG_AGGREGATE=false
 VITE_GA_TRACKING_ID=
 VITE_SENTRY_DSN=
 VITE_APP_HOST=localhost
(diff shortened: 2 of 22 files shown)


packages/client-core/src/admin/components/Project/GithubRepoDrawer.tsx
```diff
@@ -44,7 +44,6 @@ const GithubRepoDrawer = ({ open, project, onClose }: Props) => {
       setProcessing(false)
       onClose()
     } catch (err) {
-      console.error(err)
```
dmitrizagidulin, line 47: Since we're doing `NotificationService.dispatchNotify()`, it will log the error, no need to do it here.

barankyle:
Are there any security concerns with having /api/log accessible to anyone?

dmitrizagidulin:
> Are there any security concerns with having /api/log accessible to anyone?

Great question. Yes and no. 

So, on the one hand, that's the tragedy of client-side logging, right. Any time a browser-based JS app is using an API, it's basically accessible to anyone (cause there's no good way for a JS app to store any secrets). This is true of commercial off-the-shelf logging startups like [Rollbar](https://rollbar.com/) and others, too -- a hostile user can use your app, examine source, and just use your Client ID to make bogus error and log entries. It's a price you pay / tradeoff you make -- in exchange for recording all the telemetry you want, you're exposing yet another API endpoint to the wild internets.

Fwiw, to put it in perspective -- the `/api/log` endpoint is no different (meaning, no more or less vulnerable) than any of our other Ethereal Engine API endpoints (basically, everything on server-core, analytics, instance server, etc, is accessible to anyone).

Given that we kind of have to live with it, what are the threats here, and is there anything we can do to mitigate?
The way I see it, the main threats are: 1) DDOS, and 2) Opportunity for spoofed log entries.

**In terms of DDOS**: one, the logging endpoint is no different than any of our other endpoints (which can be spammed / DDOS'd), and two (more importantly), all the standard DOS mitigation techniques apply (IP filtering, rate limiting, ML, etc).

**Spoofed log entries:** much like with denial of service, a determined attacker could spam the logging endpoint with bogus events / errors etc. 
The solution there, again, is more telemetry. Obv we can record the IP addresses of the requests easily. But even more useful - we can add a persistent session ID that is transmitted with all events. Which would let us make all sorts of interesting queries, heuristics and ML filters, regarding whether an event is likely to be organic, or spoofed.

Anyways, great to think about this. On the whole, I think this is a decent risk for us to take at this point.

PR #6636, opened 2022-07-25
Editor object explode

HexaField:
## Summary
Adds basic model hierarchy inspector to editor


dinomut1 (approved):

packages/engine/src/scene/functions/loaders/ModelFunctions.ts
```diff
@@ -119,11 +119,7 @@ export const serializeModel: ComponentSerializeFunction = (entity) => {
 
 const parseModelProperties = (props): ModelComponentType => {
   return {
-    src: props.src ?? SCENE_COMPONENT_MODEL_DEFAULT_VALUE.src,
-    materialOverrides: props.materialOverrides ?? SCENE_COMPONENT_MODEL_DEFAULT_VALUE.materialOverrides,
-    matrixAutoUpdate: props.matrixAutoUpdate ?? SCENE_COMPONENT_MODEL_DEFAULT_VALUE.matrixAutoUpdate,
-    useBasicMaterial: props.useBasicMaterial ?? SCENE_COMPONENT_MODEL_DEFAULT_VALUE.useBasicMaterial,
-    isUsingGPUInstancing: props.isUsingGPUInstancing ?? SCENE_COMPONENT_MODEL_DEFAULT_VALUE.isUsingGPUInstancing,
-    isDynamicObject: props.isDynamicObject ?? SCENE_COMPONENT_MODEL_DEFAULT_VALUE.isDynamicObject
+    ...SCENE_COMPONENT_MODEL_DEFAULT_VALUE,
```
HexaField, line 122: rather than spreading objects, we should keep it as it is. the original method gives us strict control of what data enters the components

dinomut1: It doesn't make a difference with the previous syntax. If we want control of the data entering the component we would need type checks and validation, which we should just do with JSON schema instead of in our parse functions

PR #6719, opened 2022-08-03
Replace Old Particle Emitter with three-nebula

### LINKED ISSUE #6498: [Epic]: Particle system 2.0
https://discourse.threejs.org/t/nebula-a-fully-featured-particle-system-designer-for-three/21854

https://github.com/creativelifeform/three-nebula 

V1 - get our particle system to just use what this thing exports
Might be cool to buy a $5 nebula license and see what happens when we put their "fire" or something in our engine.

V2 - try to recreate it

https://github.com/needle-tools

Effekseer

https://github.com/hmans/three-vfx

## Dependent on

- [ ] 

## Stories

- [ ]

dinomut1:
Closes #6498


package.json
```diff
@@ -127,6 +127,7 @@
     "pino-pretty": "^7.6.1",
     "qrcode.react": "^3.1.0",
     "react-json-view": "^1.21.3",
+    "three-nebula": "^10.0.3",
```
HexaField, line 130: this should be added in the engine package, rather than the root

packages/engine/src/scene/components/ParticleEmitterComponent.ts
```diff
@@ -0,0 +1,15 @@
...
+import { SceneJson } from '@xrengine/common/src/interfaces/SceneInterface'
+
+import { createMappedComponent } from '../../ecs/functions/ComponentFunctions'
+
+export type ParticleGenerationMode = 'LIBRARY' | 'JSON'
+
+export type ParticleEmitterComponentType = {
+  mode: ParticleGenerationMode
+  src: any
+  args?: any
```
HexaField, line 12: lets prefer non-nullable properties on all components

packages/engine/src/scene/systems/ParticleSystem.ts
```diff
@@ -0,0 +1,87 @@
...
+      emitter.removeAllBehaviours()
+    })
+    system.renderers.map((renderer) => {
+      system.removeRenderer(renderer)
+    })
+    system.update()
+    system.emitters.map((emitter) => {
+      emitter.stopEmit()
+    })
+    return new Promise<void>((resolve) =>
```
HexaField, line 44: why do we return a promise with a setTimeout here?

HexaField (changes requested):

packages/engine/src/scene/systems/ParticleSystem.ts
dinomut1: Nebula is running its own async logic to clear the particles. It needs an additional frame execution to correctly clear the particle buffer

PR #6720, opened 2022-08-03
Fixed a bug with instance provisioning and unreachable instanceservers.

barankyle:
## Summary

On a couple of occasions, instanceservers were not externally reachable -
they could be kubectl exec'd into and responded to internal requests, but
any request from outside was hanging. Updated instance-provision to make
a request to the server it's about to return. If there's an error, or it
does not receive any response within two seconds, it removes the instance
record, deletes the pod in a K8s environment, and then tries to get another
free instanceserver.

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/server-core/src/networking/instance-provision/instance-provision.class.ts
```diff
@@ -202,6 +204,46 @@ export async function checkForDuplicatedAssignments(
...
+  // it assumes the pod is unresponsive. Locally, it just waits half a second and tries again - if the local
+  // instanceservers are rebooting after the last person left, we just need to wait a bit for them to start.
+  // In production, it attempts to delete that pod via the K8s API client and tries again.
+  let responsivenessCheck: boolean
+  responsivenessCheck = await Promise.race([
+    new Promise<boolean>((resolve) => {
+      setTimeout(() => {
+        logger.warn(`Instanceserver at ${ipAddress} too long to respond, assuming it is unresponsive and killing`)
+        resolve(false)
+      }, 2 * 1000) // timeout after 2 seconds
```
HexaField, line 218: do we want to make these times configurable?

barankyle: Made it so.

HexaField (approved):

PR #6817, opened 2022-08-20
Refactor AuthService service.

dmitrizagidulin:
* In-place refactor of AuthService.
* Convert promise syntax to await/async.
* Move avatar-related methods from AuthService to its own AvatarService
* Move actions to their own separate AuthActions file.
* Extract duplicated 'reset guest token' code to its own function.
* Add logging.


HexaField:
Hmm, i feel like it would have been better to split this file up categorically rather than by different parts of the flux pattern.

for instance, `usernameUpdatedAction` and other user actions could go in a UserProfileService, all the did stuff could go in a DIDService, all the other login stuff could go in a LoginService.

this seems a bit unnecessary, and it is nice to have all the relevant functionality of the flux pattern in one file.

barankyle (approved):

PR #6871, opened 2022-09-05
Wallet upgrade for demo

dmitrizagidulin:


packages/server-core/src/createApp.ts
```diff
@@ -184,6 +185,12 @@ export const createFeathersExpressApp = (
     return res.status(204).send()
   })
 
+  app.post('/api/credentials/request', IssueRequest.post)
```
HexaField, line 188: is there a reason we are not using feathers for this?

packages/engine/src/credentials/CredentialSystem.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export const CredentialSystem = () => {
+  const didVerifyCredentialActions = createActionQueue(CredentialAction.verificationResult.matches)
+  const nameQuery = defineQuery([NameComponent])
+
+  return () => {
+    for (const entity of nameQuery.enter()) {
+      const name = getComponent(entity, NameComponent).name
+      if (name === 'Door' || name === 'VCKey Trigger Volume') {
```
HexaField, line 38: we can't rely on component names here, we should be creating a tag component, and use that to identify an object that is requiring this functionality

HexaField (changes requested):

packages/engine/src/credentials/CredentialSystem.ts
dmitrizagidulin: @HexaField yeah, this was a temporary hack me and Gheric used for the demo -- we figured we'd discuss with you re how to do it properly

packages/engine/src/initializeEngine.ts
```diff
@@ -54,6 +55,10 @@ export const setupEngineActionSystems = () => {
     type: SystemUpdateType.FIXED_LATE,
     systemFunction: OutgoingActionSystem
   })
+  initSystemSync(world, {
```
HexaField, line 58: system should not be initialized like this, these are core internal engine systems, should be initialized in `initializeCoreSystems` asynchronously instead

packages/instanceserver/src/channels.ts
```diff
@@ -253,6 +254,12 @@ const loadEngine = async (app: Application, sceneId: string) => {
     const sceneResultPromise = app.service('scene').get({ projectName, sceneName, metadataOnly: false }, null!)
 
     await initializeCoreSystems()
+
+    initSystemSync(world, {
```
HexaField, line 258: remove this, initialize in initializeCoreSystems

HexaField (changes requested):

packages/engine/src/initializeEngine.ts
dmitrizagidulin: Thanks, moved.

packages/instanceserver/src/channels.ts
dmitrizagidulin: Done.

speigg (approved):

packages/engine/src/credentials/CredentialSystem.ts
HexaField: this still needs to be addressed

packages/engine/src/credentials/CredentialSystem.ts
```diff
@@ -0,0 +1,73 @@
...
+        addComponent(entity, CallbackComponent, {
+          receiveVC: async () => {
```
HexaField: lets pull these out into their own functions

packages/engine/src/initializeEngine.ts
```diff
@@ -160,6 +155,11 @@ export const initializeCoreSystems = async (injectedSystems?: SystemModuleType<a
       uuid: 'xre.engine.AssetSystem',
       type: SystemUpdateType.FIXED_LATE,
       systemLoader: () => import('./scene/systems/AssetSystem')
+    },
+    {
+      uuid: 'xre.engine.OutgoingActionSystem',
```
HexaField, line 160: moved the wrong system... OutgoingActionSystem needs to stay where it was

packages/engine/src/networking/systems/OutgoingActionSystem.ts
```diff
@@ -10,7 +10,7 @@ const sendOutgoingActions = (world: World) => {
   }
 }
 
-export default function OutgoingActionSystem(world: World) {
+export default async function OutgoingActionSystem(world: World) {
```
HexaField, line 13: revert

packages/server-core/src/createApp.ts
```diff
@@ -185,6 +186,12 @@ export const createFeathersExpressApp = (
...
```
HexaField, line 189: lets use a feathers route instead of a raw REST endpoint

HexaField (changes requested):

packages/engine/src/initializeEngine.ts
dmitrizagidulin: Ah, that was a rebasing error, reverting.

packages/server-core/src/createApp.ts
dmitrizagidulin: Main reason is - I don't know how to do it using feathers. Happy to refactor in the next PR tho.

packages/engine/src/credentials/CredentialSystem.ts
dmitrizagidulin: Future PR.

packages/engine/src/initializeEngine.ts
dmitrizagidulin: Reverted

packages/server-core/src/createApp.ts
dmitrizagidulin: Don't know how to do that, tho happy to refactor in the next PR.

packages/engine/src/credentials/CredentialSystem.ts
dmitrizagidulin: Extracted

packages/projects/default-project/apartment.scene.json
```diff
@@ -997,12 +997,6 @@
         {
           "name": "visible",
           "props": {}
-        },
```
HexaField, line 1000: the mount point needs to stay here. if this was due to saving the scene from the editor, then this is a bug that an be addressed elsewhere

HexaField (changes requested):

packages/engine/src/credentials/CredentialSystem.ts
```diff
@@ -0,0 +1,73 @@
+import { NotificationService } from '@xrengine/client-core/src/common/services/NotificationService'
```
HexaField, line 1: engine cannot be importing client-core, the action should be processed in a separate system in client-core

HexaField (changes requested):

packages/common/src/serverHost.ts
```diff
@@ -0,0 +1,6 @@
+export const hostDefined = !!globalThis.process.env['VITE_SERVER_HOST']
```
HexaField, line 1: we have these already in client-core/src/util/config.ts lets unify these into a common, as this file has done here

HexaField (changes requested):

packages/projects/default-project/apartment.scene.json
dmitrizagidulin: Yeah, not sure what happened there. Reverting.

dmitrizagidulin:
Ok, the ECS / CredentialSystem stuff is holding up wallet updates. I'm closing this PR and will break this into two separate PRs (put the ecs update into a separate branch).

PR #6872, opened 2022-09-06
Added static resources section in admin panel

### LINKED ISSUE #6708: [Story] Add search for all types of static resources

hanzlamateen:
## Summary

This PR introduces static resources section in admin panel and adds search to it.

image

## References

closes #6708

## Checklist

## QA Steps


barankyle:
It looks like the selector in the new drawer for staticResourceType is using the list of types that can be filtered on. I think the admin ResourceService should get the full list of accepted static-resource-types from the static-resource-type table, stored separately from the list of filterable types, and that that should be used for populating the list of resources.

Let's say we're starting with a clean database. The only types of static resources in the database are avatars and user-thumbnails. As it currently works from what I tested, you can't create an audio, or video, or volumetric static resource from the new admin page since none of those are in the list of filters.

barankyle:
Letting admins set the key might also warrant reconsideration. If I'm uploading a static resource, I don't think I want to consider what the filename is. When we're creating static resources in other services, we're hard-coding the key based on what we're uploading, like how avatars get put into /avatars/<public or a specific user id>. Maybe new static resources should have a 'name' field in the drawer, and the key is assembled behind-the-scenes into /static-resources/<name> (though that could collide, so maybe add a UUID onto the end of the name like we do with avatar resources now).

hanzlamateen:
@barankyle all above comments resolved.

mrhegemon (approved):

PR #6878, opened 2022-09-07
Dynamic Loading Improvements

dinomut1:
*duplicate object now works on object3Ds
*add SceneObjectInput component
*dynamic scene loading now correctly handles children of dynamic nodes
*add lock toggle to properties panel


packages/engine/src/scene/systems/SceneObjectDynamicLoadSystem.ts
```diff
@@ -63,7 +82,34 @@ export default async function SceneObjectDynamicLoadSystem(world: World) {
             position: data.position
           })
 
+          hasComponent(entity, RigidBodyComponent) && Physics.removeCollidersFromRigidBody(entity, world.physicsWorld)
```
HexaField, line 85: this should be happening in the physics system, not here

packages/engine/src/scene/systems/SceneObjectDynamicLoadSystem.ts
```diff
@@ -63,7 +82,34 @@ export default async function SceneObjectDynamicLoadSystem(world: World) {
...
+          const targetNode = nodeMap.get(entity)
+          if (targetNode) {
+            iterateEntityNode(targetNode, (node) => {
```
HexaField, line 88: what is the case in which a child entity is separated in distance to it's parent? the dynamic load should only ever be used in leaf nodes of the entity tree, or alternatively parent nodes should with dynamic loading should stop their children from being loaded, and then handle the loading/unloading of their children

packages/engine/src/scene/systems/SceneObjectDynamicLoadSystem.ts
```diff
@@ -63,7 +82,34 @@ export default async function SceneObjectDynamicLoadSystem(world: World) {
...
+                    const data = { name: node.uuid, components: serializeEntity(node.entity) }
+                    if (node.parentEntity) {
+                      const parentNode = nodeMap.get(node.parentEntity!)!
+                      data['parent'] = parentNode.uuid
+                    }
+                    createSceneEntity(node.uuid, data)
+                  }
+                )
+
+              hasComponent(node.entity, RigidBodyComponent) &&
```
HexaField, line 102: once again, should be done in physics system

HexaField (changes requested):

packages/engine/src/scene/systems/SceneObjectDynamicLoadSystem.ts
dinomut1: Any child that has a position offset will be separated in distance from its parent. Used when wanting to dynamically load an object or group of objects as the player approaches a position in the level.

packages/engine/src/scene/systems/SceneObjectDynamicLoadSystem.ts
dinomut1: now in physics system

PR #6890, opened 2022-09-09
Model Export Improvements

dinomut1:
*model export as gltf now saves buffers as separate files
*model export as gltf now saves images as references
*ee-material extension saves arbitrary material properties


dinomut1:
addresses #6859

PR #6957, opened 2022-09-14
Rename to ServerLogger.

dmitrizagidulin:
Rename `logger.ts` to `serverSideLogger` per @HexaField's request (to avoid confusion with the client-side/isomorphic logger).


HexaField (changes requested):
`ServerLogger` is a bit cleaner than `serverSideLogger`

dmitrizagidulin:
> `ServerLogger` is a bit cleaner than `serverSideLogger`

K, renamed.

PR #6977, opened 2022-09-19
Button Click Refactor

HexaField:
## Summary

Replaces hardcoded handlePrimaryButton and handleSecondaryButton input bindings with action dispatches in their respective systems.

Implements XRUI clickaway for the avatar profile menu.

Implements a react context for the ActiveMenu, allowing us to easily set the active menu without passing it around props.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/client-core/src/systems/AvatarUISystem.tsx
+++ b/packages/client-core/src/systems/AvatarUISystem.tsx
@@ -1,24 +1,32 @@
 import { Not } from 'bitecs'
 import { Consumer } from 'mediasoup-client/lib/Consumer'
-import { Group, Vector3 } from 'three'
+import { Group, Vector2, Vector3 } from 'three'
 
 import { UserId } from '@xrengine/common/src/interfaces/UserId'
 import multiLogger from '@xrengine/common/src/logger'
 import { AvatarComponent } from '@xrengine/engine/src/avatar/components/AvatarComponent'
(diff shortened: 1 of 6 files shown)


speigg:
The more I think about it and study how we are handling input right now, the more I believe we should be directly reading input state in our systems. While this PR is an improvement, even dispatching this `buttonClicked` action is a level of indirection that isn't really necessary, and makes our input handling less flexible (e.g., we aren't exposing anything but the "started" lifecycle state for that button, and we are potentially creating multiple sources of truth for the the same input state). All we need is a way for a system to mark a given input state as "handled", so that subsequent systems don't also try to handle it.

speigg:
Anyway, happy to merge this in as-is, but let's think about this more. I think input behaviors as we've implemented them should all be refactored into systems.

speigg (approved):

HexaField:
> The more I think about it and study how we are handling input right now, the more I believe we should be directly reading input state in our systems. While this PR is an improvement, even dispatching this `buttonClicked` action is a level of indirection that isn't really necessary, and makes our input handling less flexible (e.g., we aren't exposing anything but the "started" lifecycle state for that button, and we are potentially creating multiple sources of truth for the the same input state). All we need is a way for a system to mark a given input state as "handled", so that subsequent systems don't also try to handle it.

Agreed. We can revisit this in the future. We do actually have all the data we need here though, `clicked: true` acts as started and `clicked: false` acts as ended button click actions.

PR #7044, opened 2022-10-04
Initial changes to support room

### LINKED ISSUE #6999: [Story] Room Invite Codes
We want simple alphanumeric room invite codes that can be used to invite a user to a specific instance, rather than just an invite link.

Example: I enter `ETHEREAL` into a UI and press a play button, and anyone else who does the same gets routed to the same instance.

For now, just a simple url based entry can be implemented, no need for a new UI

- [ ] room codes can be entered with `https://localhost:3000/?roomCode=myroom`
- [ ] entering a room code returns the location name and instanceId if the user is authorised for that location
  - [ ] if no instance is yet created, create an instance and return it
- [ ] when the last user exists the instance, the room code is cleared

hanzlamateen:
## Summary

## References

closes #6999

## Checklist

## QA Steps


barankyle (changes requested):
Following the comment on making the room code a unique field, it looks to me like there's a gap related to when room codes are made that could cause problems. 

Let's say two people are making new rooms, and both of their clients generate the same code. They both press the create room button at the same time. There won't be a room code in the database during that check, so both are redirected to the location with `?roomCode=<code>` in the querystring.

Now, both clients will call instance-provision with the same roomCode. One of them will happen after the other, and by the time it goes to create the room_code entry, that code will exist. If we haven't made roomCode a unique field, then now there will be two entries with the same code. If we do make it unique, then an error will be thrown. Either is not desired. This could also be more actively triggered if someone deliberately made two+ calls at the same time to instance-provision.find with the same room code.

Thinking about this some more, I'm wondering if the room code shouldn't be provided for new rooms before they're created. There'd be a separate new boolean query parameter, createNewRoom, that gets passed to instance-provision.find. This should indicate that a free instanceserver should always be created, as opposed to being assigned to an existing instanceserver*. The 6-digit room code would be generated randomly right before where room-code.create() is called, making sure it doesn't exist first, and re-generating until a free one is found. The code then gets passed back to the client alongside the id/ipAddress/port/podname as an optional return parameter, and the client would insert it into the address bar, similar to how if you just go to /location/<name>, it inserts the returned instanceId.

Thinking about other online games I've played with room codes, this is usually how I've seen it done - the room code isn't made until the room/table/etc. is made.

* see lines 585-614 of instance-provision.class.ts There, we check if there are instances for that location that have room, and add onto the response any potential instance-authorized-users if it's a private location. Allowed instances are ones where there are no instance-authorized-users, i.e. public locations, or ones that the user is authorized to go. Under this proposal to have a createNewRoom parameter, we could skip this whole check entirely and just call getFreeInstanceServer() if that parameter is present.

barankyle (approved):

HexaField (approved):

PR #7132, opened 2022-10-25
Fixed several avatar-related bugs

barankyle:
## Summary

Spawning of avatar now uses avatar object on user object. Client had previously been using fetchAvatarList and pulling the matching avatar off that, which was prone to failure if it was not on the first page. Instanceserver now gets user via feathers instead of sequelize model so that avatar is attached to user, and does not have to do a separate fetch of user's avatar.

Made project.update not ru onInstall hook if the project already exists in the database. This was causing avatars to be duplicated.

Updated fetchAvatarList to paginate. Updated avatar selection menus to paginate themselves by their page length and to make calls to fetchAvatarList when they need to get more from the API.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/client-core/src/user/components/UserMenu/menus/EmoteMenu.tsx
+++ b/packages/client-core/src/user/components/UserMenu/menus/EmoteMenu.tsx
@@ -39,7 +39,7 @@ export const useEmoteMenuHooks = ({ changeActiveMenu }: EmoteMenuHooksProps) => 
     {
       body: <img src="/static/sad.svg" alt="sad" />,
       containerProps: {
-        onClick: () => runAnimation(AvatarStates.CLAP)
+        onClick: () => runAnimation(AvatarStates.DANCE2)
       }
     },
     {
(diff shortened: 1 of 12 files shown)


packages/client-core/src/user/components/UserMenu/menus/EmoteMenu.tsx
```diff
@@ -39,7 +39,7 @@ export const useEmoteMenuHooks = ({ changeActiveMenu }: EmoteMenuHooksProps) =>
     {
       body: <img src="/static/sad.svg" alt="sad" />,
       containerProps: {
-        onClick: () => runAnimation(AvatarStates.CLAP)
+        onClick: () => runAnimation(AvatarStates.DANCE2)
```
HexaField, line 42: why this change?

HexaField (changes requested):

packages/client-core/src/user/components/UserMenu/menus/EmoteMenu.tsx
barankyle: There'd been a comment about how it was confusing that the animation with the hand clapping icon wasn't the animation that actually clapped, so I swapped them. This admittedly isn't related to the other avatar fixes, I just threw it in. If there's a reason to the icon/animation pairing I overlooked, it can be reverted.

HexaField: no prob, i'll look into it

PR #7159, opened 2022-10-31
Updated instance-provisioning to support private instances.

barankyle:
## Summary

Changed input query param 'createNewRoom' to 'createPrivateRoom'. If this is passed, then it mostly works as before, forcing a new instance to be provisioned, but it will also make the requesting user be added as an instance_authorized_user. This will prevent ordinary instance provision requests from putting users on this instance.

Changed params in most instance-provision.class.ts functions to be objects, as a long list of input variables when many of them are optional becomes unwieldy. If a user provisions with a room code and that instance exists, an instance_authorized_user will be made for them if it does not exist.

Changed default domain on minikube build from theoverlay.io to etherealengine.com

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/server-core/src/networking/instance-provision/instance-provision.class.ts
+++ b/packages/server-core/src/networking/instance-provision/instance-provision.class.ts
@@ -1,5 +1,5 @@
 import { BadRequest, NotAuthenticated } from '@feathersjs/errors'
-import { Id, NullableId, Params, ServiceMethods } from '@feathersjs/feathers'
+import { Id, NullableId, Paginated, Params, ServiceMethods } from '@feathersjs/feathers'
 import https from 'https'
 import _ from 'lodash'
 import fetch from 'node-fetch'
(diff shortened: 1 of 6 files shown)


packages/server-core/src/networking/instance-provision/instance-provision.class.ts
```diff
@@ -483,6 +545,19 @@ export class InstanceProvision implements ServiceMethods<any> {
             (!config.kubernetes.enabled || (config.kubernetes.enabled && !isCleanup)) &&
             instance.currentUsers < location.maxUsersPerInstance
           ) {
+            if (roomCode) {
```
HexaField, line 548: should we be checking here that the roomCode provided actually corresponds to the instance? or is this handled elsewhere?

barankyle: Fixed

PR #7188, opened 2022-11-06
Multiple User Connections

HexaField:
## Summary

- [x] replace UserId as peer identifier with PeerID
- [x] refactor avatar spawning to allow multiple connections but only one avatar
- [x] refactor authority to be peer driven instead of user driven
- [x] refactor WebRTC channels to use PeerID instead of UserId

## References

closes https://github.com/XRFoundation/XREngine/issues/7069

## Checklist

## QA Steps


packages/engine/src/avatar/AvatarControllerSystem.ts
```diff
@@ -94,6 +97,16 @@ export default async function AvatarControllerSystem(world: World) {
       const controller = getComponent(controlledEntity, AvatarControllerComponent)
       updateAvatarControllerOnGround(controlledEntity)
       if (controller.movementEnabled) {
+        /** Support multiple peers controlling the same avatar by detecting movement and overriding network authority.
+         *    @todo we may want to make this an networked action, rather than lazily removing the NetworkObjectAuthorityTag
```
speigg, line 101: I agree

packages/engine/src/avatar/functions/createAvatar.ts
```diff
@@ -37,8 +50,23 @@ export const defaultAvatarHalfHeight = defaultAvatarHeight / 2
...
+    const didSpawnEarlierThanThisClient = NetworkPeerFunctions.getCachedActionsForUser(userId).find(
+      (action) =>
+        WorldNetworkAction.spawnAvatar.matches.test(action) &&
+        action !== spawnAction &&
+        action.$time > spawnAction.$time
+    )
+    if (didSpawnEarlierThanThisClient) {
+      hasComponent(entity, NetworkObjectAuthorityTag) && removeComponent(entity, NetworkObjectAuthorityTag)
+    }
+    return UndefinedEntity
```
speigg, line 67: shouldn't we return the existing entity?

speigg: We should probably rename this function to `spawnAvatarReceptor`, for clarity, and to avoid confusing with `AvatarService.createAvatar()`

speigg (changes requested):

packages/engine/src/avatar/functions/createAvatar.ts
HexaField: > On the other hand, this function is functionally just an action receptor (for the spawnAvatar action), and it's odd for it to be returning anything at all.

we are using it in tests as the entities are hidden in the receiveSpawnObject receptor, we can easily get around this

barankyle (approved):

speigg (approved):

PR #7210, opened 2022-11-12
Add behave-graph and ee-behave-flow

### LINKED ISSUE #7211: https://github.com/beeglebug/behave-flow

dinomut1:
Integrate basic scaffold for node graph editor UI using behave-graph and forked version of behave-flow
Closes #7211 and #7212


packages/engine/src/behave-graph/components/RuntimeGraphComponent.ts
```diff
@@ -0,0 +1,36 @@
...
+  name: 'EE_runtimeGraph',
+
+  onInit: (entity) => {
+    const graphComponent = getComponent(entity, BehaveGraphComponent)
+    const registry = new Registry()
+    const logger = new DefaultLogger()
+    const ticker = new ManualLifecycleEventEmitter()
+    registerCoreProfile(registry, logger, ticker)
+    const systemState = getState(BehaveGraphSystemState)
+    systemState.domains[graphComponent.domain].value.register(registry, logger, ticker)
```
HexaField, line 26: components should be thought of as just data,data structures and must not have side effects like this, we should put all of this in an enter query or useEffect

HexaField (changes requested):

packages/engine/src/behave-graph/components/RuntimeGraphComponent.ts
dinomut1: None of this is side effecting, lines 24 and 26 only modify the registry argument they are passed.
Also, the Engine and ManualLifecycleEmitter types have no default value, so I'd need to be returning either nulls (bad) or unsafely cast dud objects (also bad). The Engine constructor must accept a fully initialized Graph, which requires a registry that is staged using the registerCoreProfile function and whatever additional domain-specific, also not side effecting register function that is stored in the systemState.domains record.

packages/instanceserver/src/index.ts
```diff
@@ -1,5 +1,6 @@
 import { updateAppConfig } from '@xrengine/server-core/src/updateAppConfig'
 
+require('fix-esm').register()
```
HexaField, line 3: Lets move this to createApp or appconfig in server-core, to also patch it for servers generally, since there may be imports to the engine from there in the future, and it seems generally good to have

HexaField (changes requested):

PR #7223, opened 2022-11-18
Fixed some issues with settings updates.

barankyle:
Most settings updates were not triggering an update of the api pods' appconfigs. All settings other than authentication now call updateAppConfig.

Authentications settings patches now restart the api deployment. OAuth keys/secrets appear to be baked into the express/feathers app middleware on startup; calling app.set('authentication', <new auth config>) does not change the tokens for when a client tries to log in. It's difficult and potentially buggy to change middleware on a machine that's in use, so replacing all the api pods is the safest option.

Removed refreshAppConfig because it wasn't being used for anything useful with gitPems no longer being used, and because you probably want to refresh all config anyway.

Fixed a visual bug with one auth method being displayed twice - stemmed from how we were splicing the list of auth methods.

Updated getContentType to return 'application/octet-stream' for 'ts', as .ts files are technically mapped to a video type.

## Summary

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/server-core/src/util/fileUtils.ts
```diff
@@ -4,7 +4,7 @@ import mime from 'mime-types'
 import path from 'path'
 
 export const getContentType = (url: string): string => {
-  return mime.lookup(url) || 'application/octet-stream'
+  return /\.ts$/.exec(url) ? 'application/octet-stream' : mime.lookup(url) || 'application/octet-stream'
```
HexaField, line 7: do we want to include any/all of js, jsx, tsx in this too?

HexaField (approved):

PR #7249, opened 2022-11-22
Add EE_ecs GLTF Extension

### LINKED ISSUE #7175: [Epic]: EE_ecs GLTF Extension
Loading of embedded ECS data must be moved from the node Extras field into its own dedicated extension. 

Backwards compatibility must be preserved

- implement EE_ecs GLTFExporter extension
- implement EE_ecs GLTFImporter extension
- implement EE_ecs GLTFTransform extension

dinomut1:
Closes #7175

### DIFF
--- a/packages/engine/src/assets/loaders/gltf/extensions/EEECSImporterExtension.ts
+++ b/packages/engine/src/assets/loaders/gltf/extensions/EEECSImporterExtension.ts
@@ -1,5 +1,9 @@
 import { Event, Object3D } from 'three'
 
+import { Entity } from '../../../../ecs/classes/Entity'
+import { addComponent } from '../../../../ecs/functions/ComponentFunctions'
+import { createEntity } from '../../../../ecs/functions/EntityFunctions'
+import { NameComponent } from '../../../../scene/components/NameComponent'
 import { parseECSData } from '../../../../scene/functions/loadGLTFModel'
 import { GLTF, GLTFLoaderPlugin } from '../GLTFLoader'
 import { ImporterExtension } from './ImporterExtension'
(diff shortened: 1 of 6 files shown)


packages/engine/src/assets/loaders/gltf/extensions/EEECSImporterExtension.ts
```diff
@@ -17,8 +21,16 @@ export default class EEECSImporterExtension extends ImporterExtension implements
     const nodeDef = json.nodes[nodeIndex]
     if (!nodeDef.extensions?.[this.name]) return null
     const extensionDef: EE_ecs = nodeDef.extensions[this.name]
-    if (!parser.options.entity) throw Error('error loading ecs data')
-    parseECSData(parser.options.entity, Object.entries(extensionDef.data))
+    const entityName = extensionDef.data?.['xrengine.entity'] as string
+    if (!entityName) return null
+    let entity: Entity
+    if (NameComponent.entitiesByName[entityName]) {
```
speigg, line 27: Names are not guaranteed to be unique

speigg (changes requested):

packages/engine/src/assets/loaders/gltf/extensions/EEECSImporterExtension.ts
dinomut1: entitiesByName is a record object, so names are guaranteed to be unique since only one entity can be mapped to any name. If that is not the intended behavior then the NameComponent needs changes

speigg: Names are there as a convenience, primarily for debugging purposes, and do not guarantee uniqueness. If a second entity is added with the same name, it will override the previous one in the record.

speigg: UUIDComponent provides a unique id

dinomut1: That should not be the behavior in any case. I've modified the entitiesByName object to map names to an array of objects, which allows multiple entities with the same name without loss of information. I've then removed the above check since multiple entities with the same name is no longer undefined behavior. There is a corresponding PR in the puttclub project that adjusts the code for this change

PR #7282, opened 2022-11-29
WebXR Input Sources Standardization

HexaField:
## Summary

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/engine/src/avatar/AvatarInputSystem.ts
```diff
@@ -0,0 +1,180 @@
...
+
+    useEffect(() => {
+      if (keys.value.ShiftLeft && world.localClientEntity) {
+        const controller = getComponent(world.localClientEntity, AvatarControllerComponent)
+        controller.isWalking = !controller.isWalking
+      }
+    }, [keys.ShiftLeft])
+
+    useEffect(() => {
+      if (keys.value.KeyE && world.localClientEntity) {
```
speigg, line 36: I am a little bit worried about turning input events into an input state that is handled reactively, and thus potentially delayed. I think this means it's possible, for example, to press the E key, and then by the time that this useEffect runs, the E key will no longer be pressed—meaning nothing happens.

packages/engine/src/input/systems/ClientInputSystem.ts
```diff
@@ -1,177 +1,212 @@
...
+    if (event.button === 1) button = 'AuxiliaryClick'
+    else if (event.button === 2) button = 'SecondaryClick'
+
+    if (down) keyState[button].set(true)
+    else keyState[button].set(none)
+
+    if (down) {
+      const now = Date.now()
+      if (button === 'PrimaryClick') {
+        if (now - lastPrimaryClickDown < 200 && now - lastPrimaryClickDown > 50)
```
speigg, line 73: why are we delaying the processing of clicks?

packages/engine/src/input/systems/ClientInputSystem.ts
```diff
@@ -1,177 +1,212 @@
...
+      if (button === 'AuxiliaryClick' && keyState['AuxiliaryMove'].value) keyState['AuxiliaryMove'].set(none)
     }
   }
 
-  // now that we have the data mapped, run the behaviors
-  for (const [key, value] of inputComponent.data) {
-    if (inputComponent.schema.behaviorMap.has(key)) {
-      inputComponent.schema.behaviorMap.get(key)!(entity, key, value)
+  const handleMouseMove = (event: MouseEvent) => {
+    if (keyState['PrimaryClick'].value) keyState['PrimaryMove'].set(true)
```
speigg, line 104: why are we resetting these values to their current state?

packages/engine/src/input/systems/ClientInputSystem.ts
```diff
@@ -1,177 +1,212 @@
...
+    )
+  }
+
+  const handleTouchMove = (event: TouchEvent) => {
+    if (keyState['PrimaryClick'].value) keyState['PrimaryMove'].set(true)
+    if (keyState['SecondaryClick'].value) keyState['SecondaryMove'].set(true)
+    if (keyState['AuxiliaryClick'].value) keyState['AuxiliaryMove'].set(true)
+    const touch = event.touches[0]
+    world.pointerState.position.set(
+      (touch.clientX / window.innerWidth) * 2 - 1,
```
speigg, line 119: I'm hesitant to make this kind of input reactive; this input is likely to change every frame, which means anyone who uses this state in their reactor will schedule reactor updates every frame. This seems like a dangerous pattern. State that changes every frame should really be handled in a system. Reactor updates are much more expensive than systems, and best reserved for reacting to complex discrete state changes.

speigg (changes requested):
I think we are getting a bit reactor heavy; two main concerns:

- Converting input events into reactive state seems generally dangerous, since state is mutable by the time we process that input state reactively, it may have already changed. 
- Scheduling reactive updates every frame is a potentially huge performance bottleneck, and is a pattern we should discourage by only making state reactive if it changes infrequently (reactors are best reserved for managing complex relationships between states)

packages/engine/src/input/systems/ClientInputSystem.ts
HexaField: Read again, this is primary click and primary move

packages/engine/src/input/systems/ClientInputSystem.ts
HexaField: This is not a state reactor, just a simple vector2

speigg: Got it, misunderstood :)

speigg (changes requested):
I'd like to propose a few simplifications:

First of all, we don't really need to support double click, and we honestly probably shouldn't. It's not isomorphic, and it adds a lot of complexity. 

Supporting an explicit double-click generally means delaying single-click events (this is what browsers do). Otherwise we risk creating an api that is very hard to use, and likely to have the same user input event handled multiple times (incorrectly). Browsers allow you to opt-out of this delay by disabling double click events in various ways (e.g., preventDefault(), or touch-action: manipulation css). 

In any case, I suggest we don't rely on double click events at all. 

Also, I'd like to avoid some of the indirection introduced with the current version of `createButtonListener`. I'd propose something like this instead:

```ts
const ButtonState = defineState({} as Record<ButtonType, {
	clicked: boolean,
	pressed: boolean,
	value: number
}>)

const primaryButton = getState(ButtonInputState).Primary

// on click
If (primaryButton.clicked) {}

// while button pressed
If (primaryButton.pressed) { }

```

^ this api should be flexible enough to be easily isomorphic in all the use cases we care about, and easy to use/reason about. What do you think?

It also maps very easily to the gamepad buttons api, with only a "clicked" property added.

speigg (approved):
Looks great, although this really PR has little to do with input sources. I assume we are going to do that in a separate PR. We should  rename this one to GamePad button api standardization. 

Also, we should probably have explicit mappings for button aliases on keyboard input, the same way we handle XR gamepads. 

Finally, “PrimaryClick” button should be renamed to “Primary” button. Same with “SecondaryClick”, etc.

packages/engine/src/xr/XRControllerSystem.ts
```diff
@@ -105,7 +105,13 @@ export function updateGamepadInput(world: World, source: XRInputSource) {
       for (let i = 0; i < buttons.length; i++) {
         const buttonMapping = mapping[i]
         const button = buttons[i]
-        world.buttons[buttonMapping] = button
+        if (!world.buttons[buttonMapping] && button.pressed) {
```
speigg, line 108: Having different shapes is problematic for optimization. All button instances should be created the same exact way.

speigg (changes requested):

PR #7286, opened 2022-11-29
Add setRigidBodyType

speigg:
## Summary

Add setRigidBodyType to keep ECS in sync w/ physics


packages/engine/src/physics/components/RigidBodyComponent.ts
```diff
@@ -84,3 +91,11 @@ export const getTagComponentForRigidBody = (type: RigidBodyType): RigidBodyTypes
       return RigidBodyKinematicVelocityBasedTagComponent
   }
 }
+
+export const setRigidBodyType = (entity: Entity, type: RigidBodyType) => {
+  const rigidbody = getComponent(entity, RigidBodyComponent)
+  const oldTypeTag = getTagComponentForRigidBody(rigidbody.body.bodyType())
+  removeComponent(entity, oldTypeTag)
+  const typeTag = getTagComponentForRigidBody(type)
```
HexaField, line 99: we need to actually set the rigidbody's type

HexaField (changes requested):

PR #7304, opened 2022-12-06
Kinematic lerp

dinomut1:
Adds interpolation for kinematic rigidbody transforms
depends on #7293


packages/engine/src/physics/systems/PhysicsSystem.ts
```diff
@@ -134,8 +150,31 @@ export default async function PhysicsSystem(world: World) {
...
+    const timestep = engineState.fixedDeltaSeconds.value / substeps
+    world.physicsWorld.timestep = timestep
+    const timestepMultiplier = 50
+    /*const maxDistance = 0.5
+    const maxDistanceSq = maxDistance * maxDistance*/
+    const kinematicBodies = [
+      ...kinematicPositionBodyQuery(),
+      ...kinematicVelocityBodyQuery()
+    ] /*.filter(kinematicBodyEntity => {
+      const transformComponent = getComponent(kinematicBodyEntity, TransformComponent)
```
speigg, line 162: Let's just delete this commented out code, I don't think we'll need it. We can just add a tag to opt in or out of kinematic smoothing later.

speigg (changes requested):

packages/engine/src/physics/systems/PhysicsSystem.ts
```diff
@@ -57,6 +59,15 @@ export const PhysicsPrefabs = {
   collider: 'collider' as const
 }
 
+export function smoothKinematicBody(entity: Entity, alpha: number) {
+  const rigidbodyComponent = getComponent(entity, RigidBodyComponent)
+  const transformComponent = getComponent(entity, TransformComponent)
+  rigidbodyComponent.position.lerp(transformComponent.position, alpha)
+  rigidbodyComponent.rotation.slerp(transformComponent.rotation, alpha)
```
HexaField, line 66: I believe we are using lerp elsewhere for rotations, and slerp is quite expensive. Perhaps when we need, we can add an option to slerp per rigidbody, but for most cases interpolating between frames a basic lerp should suffice.

packages/engine/src/physics/systems/PhysicsSystem.ts
```diff
@@ -134,8 +156,13 @@ export default async function PhysicsSystem(world: World) {
 
     // step physics world
     const substeps = engineState.physicsSubsteps.value
-    world.physicsWorld.timestep = engineState.fixedDeltaSeconds.value / substeps
+    const timestep = engineState.fixedDeltaSeconds.value / substeps
+    world.physicsWorld.timestep = timestep
+    const smoothAlpha = 50 * timestep
```
HexaField, line 161: Why 50? What does this mean?

packages/engine/src/physics/systems/PhysicsSystem.ts
```diff
@@ -134,8 +156,13 @@ export default async function PhysicsSystem(world: World) {
...
     for (let i = 0; i < substeps; i++) {
+      // smooth kinematic pose changes
+      for (const entity of kinematicPositionBodyQuery()) smoothKinematicBody(entity, smoothAlpha)
```
HexaField, line 164: Let's store the result of this query outside the substep loop to avoid multiple calls to bitecs.

packages/engine/src/transform/systems/TransformSystem.ts
```diff
@@ -275,7 +280,12 @@ export default async function TransformSystem(world: World) {
       if (makeDirty) world.dirtyTransforms[entity] = true
     }
 
-    const dirtyRigidbodyEntities = invCleanDynamicRigidbodyEntities.filter(isDirty)
+    const dirtyRigidbodyEntities = invCleanDynamicRigidbodyEntities.filter(isDirty).filter(
```
HexaField, line 283: Let's create a new filter function, rather filtering twice and creating an arrow function.

HexaField (changes requested):

packages/engine/src/physics/systems/PhysicsSystem.ts
speigg: It's a smoothness multiplier

packages/engine/src/physics/systems/PhysicsSystem.ts
speigg: Only gets called once

packages/engine/src/transform/systems/TransformSystem.ts
speigg: Sure

packages/engine/src/physics/systems/PhysicsSystem.ts
HexaField: let's try and avoid magic numbers and give this a name

HexaField (approved):

PR #7358, opened 2022-12-16
Minimum changes to make project compatible with Apple silicon

seuck:
Minimum changes to make project compatible with Apple silicon.

## Summary
- Installation process finalises successfully on Apple silicon
- Server starts and "location/default" is loaded on Chrome browser with some visual glitches

## QA Notes
- Tested using Node version 16.19
- Python installed manually and, after installation, the provided script “Install Certificates.command” has been executed.


speigg:
@barankyle can you verify that updating all these packages is fine?

HexaField:
> @barankyle can you verify that updating all these packages is fine?

We probably also don't need to add mediasoup to the root package json, @seuck if the intention is to update the 
mediasoup package, we only need to do this in the `instanceserver` package.

We would also appreciate a small writeup on the reason these packages need to be updated for Apple silicon if you have the time, thanks for this!

PR #7490, opened 2023-01-24
KTX2 Support

dinomut1:
Support KTX2 compression in Model export.
Add "Compress" option in file browser, attached to KTX2Encode feathers service


packages/server-core/src/assets/ktx2-encode/ktx2-encode.service.ts
```diff
@@ -0,0 +1,50 @@
...
+import util from 'util'
+
+import { KTX2EncodeArguments } from '@xrengine/engine/src/assets/constants/CompressionParms'
+
+import { Application } from '../../../declarations'
+
+declare module '@xrengine/common/declarations' {
+  interface ServiceTypes {
+    'ktx2-encode': {
+      create: any
```
HexaField, line 15: we should add the return type for this API

HexaField (changes requested):

HexaField (approved):

speigg (approved):

PR #7501, opened 2023-01-27
Static resource tables

barankyle:
## Summary

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/engine/src/scene/components/TypeComponent.ts
```diff
@@ -0,0 +1,36 @@
+import { hookstate, none } from '@xrengine/hyperflux'
+
+import { Entity } from '../../ecs/classes/Entity'
+import {defineComponent, getComponent} from '../../ecs/functions/ComponentFunctions'
+import {NameComponent} from "./NameComponent";
+
+export const TypeComponent = defineComponent({
+  name: 'TypeComponent',
+
+  onInit: () => '',
```
HexaField, line 10: should return true

HexaField (changes requested):

packages/common/src/dbmodels/Image.ts
```diff
@@ -0,0 +1,8 @@
+export interface ImageInterface {
+  id: string
+  name?: string
+  tags?: string[]
+  height: number
+  width: number
+  jpegStaticResourceId?: string
```
speigg, line 7: this should just be `staticResourceId`

packages/common/src/interfaces/AudioInterface.ts
```diff
@@ -0,0 +1,9 @@
+export interface AudioInterface {
+  id: string
+  name?: string
+  tags?: string[]
+  duration: number
+  mp3StaticResourceId?: string
+  oggStaticResourceId?: string
+  mpegStaticResourceId?: string
```
speigg, line 8: No reason for these to be separate static resources. An item in the audio table should point to a single static resource. That resource can be represented in a variety of formats / lods.

packages/common/src/interfaces/ImageInterface.ts
```diff
@@ -0,0 +1,18 @@
...
+  width: number
+  src: string
+  jpegStaticResourceId?: string
+  jpegStaticResource?: StaticResourceInterface
+  ktx2StaticResourceId?: string
+  ktx2StaticResource?: StaticResourceInterface
+  gifStaticResourceId?: string
+  gifStaticResource?: StaticResourceInterface
+  pngStaticResourceId?: string
+  pngStaticResource?: StaticResourceInterface
```
speigg, line 17: Likewise here; each item in the image table should point to a single static resource.

packages/common/src/interfaces/VideoInterface.ts
```diff
@@ -0,0 +1,16 @@
...
+  name?: string
+  tags?: string[]
+  height: number
+  width: number
+  duration: number
+  mp4StaticResourceId?: string
+  m3u8StaticResourceId?: string
+  mp4StaticResource?: StaticResourceInterface
+  m3u8StaticResource?: StaticResourceInterface
+  thumbnail?: string
```
speigg, line 14: Same; each video should point to a single static resource

packages/common/src/interfaces/VolumetricInterface.ts
```diff
@@ -0,0 +1,11 @@
+export interface VolumetricInterface {
+  id: string
+  name?: string
+  tags?: string[]
+  duration: number
+  src: string
+  thumbnail?: string
```
speigg, line 7: `thumbnailResourceId`

speigg (changes requested):
So, generally, the idea is that whenever files are _tightly coupled_ to one another (e.g., uvol geometry data in .drcs and texture data in .mp4 ), or _interchangeable variants_ of one another (e.g., different encodings of the same content), they should be contained in a single resource. 

Each **Image** in the image table should point to a single image `resourceId`. 

Each **Volumetric** should contain a `thumbnailId` and a `resourceId`. 

Each **Model** should contain a `resourceId` (for it's model various files) a `thumbnailId` , and maybe later (if we find good workflows / use cases) additional resource ids for overriding materials or textures or whatnot. 

Each **Avatar** should contain a `resourceId`, a `thumbnailId`, a `modelId`, and possibly additional model id slots for accessories (hats, gloves, etc.). 

All of the other tables don't really have to be done in a single PR. It would be preferable to have a clean vertical slice that we can validate with an end-to-end workflow. 

Let's also make sure each static resource can be linked somehow to each project.

speigg (approved):

PR #7506, opened 2023-01-27
Fix Loop Animation Component / GLTF Imported Animations

AidanCaruso:
Small changes. Any imported models now have animations set, and transforms are marked as dirty when animations are present. 

Also, changed `if (makeDirty) world.dirtyTransforms[entity] = true` to directly set the dirtyTransforms to the value of makeDirty, the extra logic just didn't make sense to me.

### DIFF
--- a/packages/engine/src/transform/systems/TransformSystem.ts
+++ b/packages/engine/src/transform/systems/TransformSystem.ts
@@ -303,8 +303,7 @@ export default async function TransformSystem(world: World) {
         world.dirtyTransforms[entity] ||
         world.dirtyTransforms[getOptionalComponent(entity, LocalTransformComponent)?.parentEntity ?? 0] ||
         world.dirtyTransforms[getOptionalComponent(entity, ComputedTransformComponent)?.referenceEntity ?? 0] ||
-        hasComponent(entity, ComputedTransformComponent) ||
-        (hasComponent(entity, AnimationComponent) && getComponent(entity, AnimationComponent).animations.length > 0)
+        hasComponent(entity, ComputedTransformComponent)
 
       world.dirtyTransforms[entity] = makeDirty
     }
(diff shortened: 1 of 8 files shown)


packages/engine/src/transform/systems/TransformSystem.ts
```diff
@@ -302,8 +303,10 @@ export default async function TransformSystem(world: World) {
         world.dirtyTransforms[entity] ||
         world.dirtyTransforms[getOptionalComponent(entity, LocalTransformComponent)?.parentEntity ?? 0] ||
         world.dirtyTransforms[getOptionalComponent(entity, ComputedTransformComponent)?.referenceEntity ?? 0] ||
-        hasComponent(entity, ComputedTransformComponent)
-      if (makeDirty) world.dirtyTransforms[entity] = true
+        hasComponent(entity, ComputedTransformComponent) ||
+        (hasComponent(entity, AnimationComponent) && getComponent(entity, AnimationComponent).animations.length > 0)
```
HexaField, line 307: For performance reasons, we don't want to add if statements to the transform system if possible. This can be done in another system, say the AnimationSystem, where we can set dirty upon updating the animation mixer.

HexaField (changes requested):

packages/engine/src/transform/systems/TransformSystem.ts
AidanCaruso: Yep that makes way more sense

packages/engine/src/transform/systems/TransformSystem.ts
```diff
@@ -303,7 +304,8 @@ export default async function TransformSystem(world: World) {
         world.dirtyTransforms[getOptionalComponent(entity, LocalTransformComponent)?.parentEntity ?? 0] ||
         world.dirtyTransforms[getOptionalComponent(entity, ComputedTransformComponent)?.referenceEntity ?? 0] ||
         hasComponent(entity, ComputedTransformComponent)
-      if (makeDirty) world.dirtyTransforms[entity] = true
+
+      world.dirtyTransforms[entity] = makeDirty
```
speigg, line 308: `world.dirtyTransfoms` is defined as `Record<Entity, true>`, and we remove items from that record via `delete world.dirtyTranforms[eid]`

If we are going to allow this record to contain `false`, we should also change it's definition, and change all the `delete` operations to instead set the value to `false`.

speigg (changes requested):

packages/engine/src/transform/systems/TransformSystem.ts
AidanCaruso: Hmm, maybe it'd be best to make that into its own PR rather than shoehorn it into a simple animation fix then, assuming this is worth it for performance.

HexaField (approved):

PR #7518, opened 2023-01-30
Autopilot

### LINKED ISSUE #3811: Autopilot Improvements
- [ ] Integrate Autopilot (Tap to Move) into our various scenes
- [ ] Render target waypoint graphic at tap/click location

AidanCaruso:
## Summary

- Adds right click activated click/tap-to-move autopilot
- Autopilot uses basic vector math and heuristics to move avatar toward selected point on a collider, without a navmesh
- Adds a visual object to show the autopilot's destination point

## References

closes #3811

## Checklist

## QA Steps


packages/engine/src/avatar/functions/autopilotFunctions.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Physics, RaycastArgs } from '../../physics/classes/Physics'
+import { CollisionGroups } from '../../physics/enums/CollisionGroups'
+import { getInteractionGroups } from '../../physics/functions/getInteractionGroups'
+import { SceneQueryType } from '../../physics/types/PhysicsTypes'
+
+export const autopilotGetPosition = (entity: Entity): Vector3 | undefined => {
+  const physicsWorld = Engine.instance.currentWorld.physicsWorld
+  const world = Engine.instance.currentWorld
+  const interactionGroups = getInteractionGroups(CollisionGroups.Default, CollisionGroups.Ground)
+  const raycastArgs = {
```
HexaField, line 14: this object can be moved out of the function and we update the properties that need to be updated

packages/engine/src/avatar/functions/moveAvatar.ts
```diff
@@ -128,9 +129,26 @@ export const updateReferenceSpaceFromAvatarMovement = (movement: Vector3) => {
...
+export const applyAutopilotInput = (entity: Entity) => {
+  const controller = getComponent(entity, AvatarControllerComponent)
+  if (!controller || controller.autopilotWalkpoint == undefined) return
+
+  const walkpoint = new Vector3()
+  walkpoint.set(controller.autopilotWalkpoint.x, controller.autopilotWalkpoint.y, controller.autopilotWalkpoint.z)
+  const avatarPos = getComponent(entity, TransformComponent).position
+  const moveDir = walkpoint.sub(avatarPos)
+
+  if (moveDir.length() > 0.5) updateLocalAvatarPosition(moveDir.normalize().multiplyScalar(0.1))
```
HexaField, line 144: we can optimize this to be `moveDir.lengthSq() > 0.5 * 0.5` and we should precompute 0.5 in a module scoped const and give it a name to avoid putting magic numbers in the code

HexaField (changes requested):

packages/engine/src/avatar/functions/moveAvatar.ts
AidanCaruso: Will do, using square length always seems to slip my mind.

packages/engine/src/avatar/AvatarInputSystem.ts
```diff
@@ -121,6 +123,11 @@ export default async function AvatarInputSystem(world: World) {
     )
   }
 
+  const canvas = EngineRenderer.instance.renderer.domElement
+  canvas.addEventListener('auxclick', () => {
```
HexaField, line 127: should use `world.buttons` instead of adding event listeners

packages/engine/src/avatar/functions/autopilotFunctions.ts
```diff
@@ -0,0 +1,97 @@
...
+  placeMarker(avatarControllerComponent, rayNormal)
+}
+
+export const ScaleFluctuate = (Marker: Object3D, sinOffset = 4, scaleMultiplier = 0.2, pulseSpeed = 10) => {
+  const scalePulse = scaleMultiplier * (sinOffset + Math.sin(pulseSpeed * Engine.instance.currentWorld.elapsedSeconds))
+  Marker.scale.set(scalePulse, 1, scalePulse)
+  Marker.updateMatrixWorld()
+}
+
+export class AutopilotMarker {
```
HexaField, line 51: rather than a class singleton, this can be a hyperflux state property

packages/engine/src/avatar/functions/autopilotFunctions.ts
```diff
@@ -0,0 +1,97 @@
...
+    controller.autopilotWalkpoint.y,
+    controller.autopilotWalkpoint.z
+  )
+  const newRotation = new Quaternion().setFromUnitVectors(V_010, rayNormal)
+  autopilotMarkerObject.rotation.setFromQuaternion(newRotation)
+
+  const waitForAutopilot = () => {
+    const reached = (resolve) => {
+      if (!controller.autopilotWalkpoint) resolve()
+      else setTimeout((_) => reached(resolve), 250)
```
HexaField, line 83: we want to avoid promises and setTimeout usage. any async logic will happen outside the animation frame and will reduce performance.

if the intention here is to just wait for the autopilot movement to be completed, we should use a state reactor to respond to state changes

packages/engine/src/avatar/functions/moveAvatar.ts
```diff
@@ -133,9 +135,46 @@ export const updateReferenceSpaceFromAvatarMovement = (movement: Vector3) => {
...
+  const controller = getComponent(entity, AvatarControllerComponent)
+  if (!controller || controller.autopilotWalkpoint == undefined) return
+
+  if (controller.gamepadLocalInput.lengthSq() > 0) {
+    controller.autopilotWalkpoint = undefined
+    return
+  }
+
+  const walkpoint = new Vector3()
+  walkpoint.set(controller.autopilotWalkpoint.x, controller.autopilotWalkpoint.y, controller.autopilotWalkpoint.z)
```
HexaField, line 154: can be `walkpoint.copy(controller.autopilotWalkpoint)` instead of .set

packages/engine/src/avatar/components/AvatarControllerComponent.ts
```diff
@@ -26,6 +26,8 @@ export type AvatarControllerComponentType = {
   /** @todo refactor animation system */
   speedVelocity: { value: number }
   translationApplied: Vector3
+  // Below are autopilot input values
+  autopilotWalkpoint: Vector3 | undefined
```
HexaField, line 30: let's call this autopilotTarget and have it stored in hyperflux rather than the controller component, as it will only ever be used by the local user's avatar

packages/engine/src/avatar/functions/moveAvatar.ts
```diff
@@ -133,9 +135,46 @@ export const updateReferenceSpaceFromAvatarMovement = (movement: Vector3) => {
...
+  const avatarMovementSettings = getState(AvatarMovementSettingsState).value
+  const legSpeed = controller.isWalking ? avatarMovementSettings.walkSpeed : avatarMovementSettings.runSpeed
+  const delta = 0.0175
+
+  if (distanceSquared > minimumDistanceSquared)
+    updateLocalAvatarPosition(
+      moveDirection
+        .normalize()
+        .multiplyScalar(delta * legSpeed)
+        .add(new Vector3(0, controller.verticalVelocity, 0))
```
HexaField, line 169: avoid creating new vectors in hot spots. this can be `.setY(` instead of `.add(`

HexaField (changes requested):

packages/engine/src/avatar/AvatarInputSystem.ts
AidanCaruso: I actually was using world.buttons originally but switched back to event listeners because I felt it'd be better than writing extra logic to detect the beginning of a click from the world.buttons mouse click.

HexaField: should be as easy as `if(world.buttons.Primary?.pressed)`

speigg (changes requested):
Let's do normal (left) click for click-to-move, please; otherwise, looking great :)

HexaField:
> Let's do normal (left) click for click-to-move, please; otherwise, looking great :)

How are we differentiating this from normal mouse grab look?

AidanCaruso:
Yup, that was my dilemma. Set it to right click just to avoid calling autopilot functions when dragging the camera.

packages/engine/src/avatar/functions/autopilotFunctions.ts
```diff
@@ -69,13 +63,20 @@ const SetupMarker = () => {
   markerState.merge({ markerObject: mesh })
 }
 
+export const ScaleFluctuate = (sinOffset = 4, scaleMultiplier = 0.2, pulseSpeed = 10) => {
```
HexaField, line 66: function names should be camelCase

packages/engine/src/avatar/functions/moveAvatar.ts
```diff
@@ -147,23 +147,22 @@ const _additionalMovement = new Vector3()
  */
 
 const minimumDistanceSquared = 0.5 * 0.5
-
+const walkPoint = new Vector3() as Vector3
```
HexaField: unnecessary `as Vector3` cast

packages/engine/src/avatar/AvatarMovementSystem.ts
```diff
@@ -1,9 +1,17 @@
...
+import { EngineActions } from '../ecs/classes/EngineState'
 import { World } from '../ecs/classes/World'
-import { applyGamepadInput } from './functions/moveAvatar'
+import { EngineRenderer } from '../renderer/WebGLRendererSystem'
+import { applyAutopilotInput, applyGamepadInput } from './functions/moveAvatar'
 
 export default async function AvatarMovementSystem(world: World) {
   const execute = () => {
     applyGamepadInput(world.localClientEntity)
+    applyAutopilotInput(world.localClientEntity)
```
HexaField, line 14: rather than adding to AvatarMovementSystem, we should keep this modular and make a copy of this file that handles applyAutopilotInput, such that we can  use each system in isolation

packages/engine/src/avatar/functions/autopilotFunctions.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const AutopilotMarker = defineState({
+  name: 'autopilotMarkerState',
+  initial: () => ({
+    markerObject: undefined as undefined | Object3D,
+    walkTarget: undefined as Vector3 | undefined
+  })
+})
+
+const SetupMarker = () => {
```
HexaField, line 56: function names should be camelCase

packages/engine/src/avatar/functions/autopilotFunctions.ts
```diff
@@ -0,0 +1,101 @@
...
+  const newRotation = new Quaternion().setFromUnitVectors(V_010, rayNormal)
+
+  marker.position.copy(markerState.walkTarget.value)
+  marker.quaternion.copy(newRotation)
+  marker.updateMatrixWorld()
+}
+
+const minDot = 0.45
+export const assessWalkability = (entity: Entity, rayNormal: Vector3): boolean => {
+  const flatEnough = rayNormal.dot(V_010) > minDot
```
HexaField, line 93: can just return the result rather than declaring a new const

packages/engine/src/avatar/functions/moveAvatar.ts
```diff
@@ -1,27 +1,29 @@
-import { QueryFilterFlags } from '@dimforge/rapier3d-compat'
-import { Euler, Matrix4, Quaternion, Vector3 } from 'three'
+import RAPIER, { QueryFilterFlags } from '@dimforge/rapier3d-compat'
```
HexaField, line 1: remove unnecessary top level import

packages/engine/src/avatar/functions/moveAvatar.ts
```diff
@@ -140,9 +142,48 @@ export const updateReferenceSpaceFromAvatarMovement = (movement: Vector3) => {
...
+ * Avatar movement via click/pointer position
+ */
+
+const minimumDistanceSquared = 0.5 * 0.5
+const walkPoint = new Vector3()
+export const applyAutopilotInput = (entity: Entity) => {
+  const markerState = getState(AutopilotMarker)
+
+  const controller = getComponent(entity, AvatarControllerComponent)
+  if (!controller || markerState.walkTarget.value == undefined) return
```
HexaField, line 155: can do `!markerState.walkTarget.value` instead of `== undefined`

HexaField (changes requested):

packages/engine/src/avatar/functions/moveAvatar.ts
AidanCaruso: No idea why I wrote the cast. But I added a commit specifically to remove that before I asked you to review so not sure why it still showed up for you.

packages/engine/src/avatar/functions/moveAvatar.ts
AidanCaruso: Ya, no idea why I used !controller but then did == undefined immediately after.

packages/engine/src/avatar/functions/moveAvatar.ts
```diff
@@ -117,7 +119,7 @@ export function updateLocalAvatarPosition(additionalMovement?: Vector3) {
     const hit = groundHits[0]
     const controllerOffset = controller.controller.offset()
     // controller.isInAir = !grounded
-    controller.isInAir = hit.distance > 1 + controllerOffset * 1.5
+    controller.isInAir = hit.distance > 1 + controllerOffset * 10
```
AidanCaruso, line 122: I cranked the controller offset multiplier value up to 10 experimentally. This prevents the value from erroneously returning true on the sky station stairs, causing the avatar to immediately fall down when using autopilot. In my testing I did not find any issues caused by this higher value.

HexaField: Please check on puttclub, this will likely cause unintended consequences

AidanCaruso: I tested location/puttclub/alpha on desktop by running/jumping around and noticed no major difference, actually an improvement I saw by setting the value to 10 was that isInAir is less frequently erroneously set to true when moving between minor obstacles, like the edges of golf courses. This means no sudden transition to mid air animations in my testing. That said, a common issue in both cases is that the avatar would get stuck on tiny barriers, when you would expect a sliding motion.

Any other specific testing you have in mind?

AidanCaruso: In fact, there are still some odd edge cases in which isInAir is set to true when it shouldn't be. Because autopilot has to return if we are no longer grounded, this can obviously cause some issues.

speigg (approved):

HexaField (approved):

PR #7535, opened 2023-02-03
Replace three-nebula with three.quarks

dinomut1:
https://www.npmjs.com/package/three.quarks

Re-implement particle system functionality using three.quarks


dinomut1:
closes #7542

speigg (approved):
Skimmed through the changes, looks good to me!

PR #7543, opened 2023-02-07
Drop Shadows

### LINKED ISSUE #6885: Fallback to soft sphere shadows on mobile
E.g., 

Something like this projected beneath all avatars and/or other models 

https://threejs.org/manual/?q=shadow#en/shadows

AidanCaruso:
## Summary

- Adds drop shadows
- Toggled via the ShadowComponent in-editor
- Entities with the DropShadowComponent are queried
- Right now based on bounding boxes and places via threejs raycast
- Uses threejs instanced meshes

## References

closes #6885

## Checklist

image

- [ ] Minor issue with how planes are projected. Currently planes will be drawn even if there is no surface under them, as long as the center raycast returns a hit.  This can be fixed by switching from one instanced mesh to an array of meshes based on decal geometry assuming it is performant enough.
image


packages/engine/src/avatar/functions/spawnAvatarReceptor.ts
```diff
@@ -121,7 +121,7 @@ export const spawnAvatarReceptor = (spawnAction: typeof WorldNetworkAction.spawn
     createAvatarCollider(entity)
   }
 
-  addComponent(entity, ShadowComponent, { receive: true, cast: true })
+  addComponent(entity, ShadowComponent, { receive: true, cast: true, castDropShadow: true })
```
HexaField, line 124: instead of a new property, lets just use `cast` and use drop shadows only when 3d shadows are turned off

packages/engine/src/scene/systems/ShadowSystem.ts
```diff
@@ -96,7 +116,73 @@ export default async function ShadowSystem(world: World) {
...
+  const texture = loader.load(
+    (process.env['VITE_FILE_SERVER'] ?? 'https://localhost:8642') + '/projects/default-project/public/drop-shadow.png'
+  )
+  const shadowMaterial = new MeshBasicMaterial({ map: texture, side: DoubleSide, transparent: true })
+  let dropShadows = new InstancedMesh(shadowGeometry, shadowMaterial, shadowComponentQuery().length)
+
+  let sceneObjects = Array.from(Engine.instance.currentWorld.objectLayerList[ObjectLayers.Camera] || [])
+  console.log(sceneObjects)
+
+  const CreateDropShadows = () => {
```
HexaField, line 133: this can be refactored into a query reactor, in which if the conditions of regular shadows disabled & casts shadows are met on an entity, we create a drop shadow

packages/engine/src/scene/systems/ShadowSystem.ts
```diff
@@ -96,7 +116,73 @@ export default async function ShadowSystem(world: World) {
...
+      const transform = getComponent(entity, TransformComponent)
+
+      const raycaster = new Raycaster()
+      raycaster.layers.disable(ObjectLayers.DropShadowCaster)
+      raycaster.firstHitOnly = true
+      raycaster.set(transform.position, shadowDirection)
+
+      const intersects = raycaster.intersectObjects(sceneObjects)
+      if (!intersects.length || !intersects[0].face) {
+        setDropShadowMatrix(new Matrix4())
```
HexaField, line 165: any math objects (Matrix4, Box3, Quaternion etc) that are not going to persist after calculations have been done should be made into scratch variables to minimize garbage collection

HexaField (changes requested):

packages/engine/src/scene/systems/ShadowSystem.ts
```diff
@@ -96,7 +116,73 @@ export default async function ShadowSystem(world: World) {
...
+
+      const setDropShadowMatrix = (matrix: Matrix4) => {
+        dropShadows.setMatrixAt(index, matrix)
+        index++
+      }
+
+      const transform = getComponent(entity, TransformComponent)
+
+      const raycaster = new Raycaster()
+      raycaster.layers.disable(ObjectLayers.DropShadowCaster)
```
HexaField, line 159: rather than a new object layer here, we can just ensure the instanced mesh does not have the camera layer enabled

HexaField (changes requested):

packages/engine/src/scene/systems/ShadowSystem.ts
```diff
@@ -96,7 +116,73 @@ export default async function ShadowSystem(world: World) {
...
+      dropShadows.matrixAutoUpdate = false
+      dropShadows.layers.set(ObjectLayers.DropShadowCaster)
+      world.scene.add(dropShadows)
+    }
+
+    let index = 0
+    sceneObjects = Array.from(Engine.instance.currentWorld.objectLayerList[ObjectLayers.Camera] || [])
+
+    for (const entity of shadowComponentQuery()) {
+      const group = getComponent(entity, GroupComponent)
```
HexaField, line 153: we can avoid this if statement by including GroupComponent in the query

packages/engine/src/avatar/functions/spawnAvatarReceptor.ts
AidanCaruso: This works, I just recall @speigg mentioning it would be ideal to be able to use drop shadows together with real shadows. Unless I recall incorrectly of course.

HexaField (changes requested):

packages/engine/src/scene/systems/ShadowSystem.ts
AidanCaruso: Oh wow I can't believe I didn't catch that one!!

speigg (approved):

PR #7559, opened 2023-02-11
Re-add castShadow property to light components and default to false

dinomut1:
For performance reasons we want to have lights not cast shadows by default

### DIFF
--- a/packages/engine/src/scene/components/PointLightComponent.ts
+++ b/packages/engine/src/scene/components/PointLightComponent.ts
@@ -11,19 +11,6 @@ import { ObjectLayers } from '../constants/ObjectLayers'
 import { setObjectLayers } from '../functions/setObjectLayers'
 import { addObjectToGroup, removeObjectFromGroup } from './GroupComponent'
 
-export type PointLightComponentType = {
-  color: Color
-  intensity: number
-  range: number
-  decay: number
-  castShadow: boolean
-  shadowMapResolution: number
-  shadowBias: number
-  shadowRadius: number
-  light: PointLight
-  helper: Object3D | null
-}
-
 export const PointLightComponent = defineComponent({
   name: 'PointLightComponent',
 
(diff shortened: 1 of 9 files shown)


packages/engine/src/scene/components/PointLightComponent.ts
```diff
@@ -11,6 +11,19 @@ import { ObjectLayers } from '../constants/ObjectLayers'
 import { setObjectLayers } from '../functions/setObjectLayers'
 import { addObjectToGroup, removeObjectFromGroup } from './GroupComponent'
 
+export type PointLightComponentType = {
```
HexaField, line 14: We can use ComponentType<typeof PointLightComponent> for these types

HexaField (changes requested):
Is there a reason we can't use a ShadowComponent with cast set to false? That was the intention of removing the property previously

packages/engine/src/scene/components/PointLightComponent.ts
dinomut1: Why not put the castShadow property with all of the other shadow properties that are in the light components?

dinomut1: Also, the lights need different default values for castShadow than meshes and whatnot (lights false, meshes true)

HexaField: > Why not put the castShadow property with all of the other shadow properties that are in the light components?

That's a good point, in this case we should remove remove castShadow from ShadowComponent and turn it into a tag component

speigg (approved):

packages/engine/src/scene/components/PointLightComponent.ts
speigg: @HexaField what about receiving shadows? You want a tag component for casting and receiving each?

HexaField: Sorry, yes, I mean put castShadow on light components, and have receiveShadow be set only via the presence of a tag ShadowComponent

PR #7577, opened 2023-02-14
Fixed some bugs with projects

barankyle:
## Summary

Some projects were not showing up as pushable even if the user had GH access to them. This was caused by not being the owner of the project and the project being saved with the '.git' suffix on the URL; the check for whether a user had push access solely through GH was not checking for projects that had the '.git' suffix on them. Made allowedRepos add a copy of the URL with the '.git' suffix before doing the findAll on projects.

Updated GITHUB_URL_REGEX to include a signature on the URL. This only really came up with projects that were present locally after a database reset, but it was breaking in that case. Since the signature is a capture group, everything that uses this regex now looks for data on index '2' from the exec instead of '1'.

Fixed an inconsistency in project patching return data. Client-side listener now expects params to be the project data, instead of an object with a 'project' child.

Fixed a bug with input text controlled by service state data. React's behavior is to set the selector to the end of an input if its value changes. This was causing some instances, like in project updating, to be very annoying to update through manual typing, since every character typed or deleted would jump to the end. Updated InputText to use a useEffect to reset the selector back to where it started.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/client-core/src/admin/services/ProjectUpdateService.ts
+++ b/packages/client-core/src/admin/services/ProjectUpdateService.ts
@@ -44,7 +44,7 @@ const initializeProjectUpdateReceptor = (action: typeof ProjectUpdateActions.ini
     selectedSHA: '',
     projectName: '',
     submitDisabled: true,
-    triggerSetDestination: false,
+    triggerSetDestination: '',
     updateType: 'none' as ProjectUpdateType,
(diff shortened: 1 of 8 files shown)


HexaField (approved):
Nice fixes.

PR #7579, opened 2023-02-14
Ecs drop shadows

AidanCaruso:
## Summary

Switch drop shadows from threejs instanced mesh object to ECS individual meshes. This allows for per-mesh materials.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/engine/src/scene/systems/ShadowSystem.ts
+++ b/packages/engine/src/scene/systems/ShadowSystem.ts
@@ -241,7 +241,7 @@ export default async function ShadowSystem(world: World) {
 
         const opacityBias = 1
         const shadowMaterial = (getComponent(dropShadowComponent.entity, GroupComponent)[0] as any).material as Material
-        shadowMaterial.opacity = Math.min(opacityBias / centerCorrectedDist, 1)
+        shadowMaterial.opacity = Math.max(Math.min(opacityBias / centerCorrectedDist, 1), 0)
 
         const sizeBias = 1
         const finalSize = dropShadowComponent.radius * Math.min(centerCorrectedDist * sizeBias, 2)
--- a/packages/engine/src/scene/systems/ShadowSystem.ts
+++ b/packages/engine/src/scene/systems/ShadowSystem.ts
@@ -1,12 +1,15 @@
 import { useEffect } from 'react'
 import {
   Box3,
+  BoxGeometry,
   DirectionalLight,
   DoubleSide,
   Group,
   InstancedMesh,
   Matrix4,
+  Mesh,
   MeshBasicMaterial,
+  Object3D,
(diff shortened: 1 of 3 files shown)


packages/engine/src/scene/systems/ShadowSystem.ts
```diff
@@ -223,34 +237,38 @@ export default async function ShadowSystem(world: World) {
...
 
         const intersected = raycaster.intersectObjects(sceneObjects)[0]
         if (!intersected || !intersected.face) {
-          setDropShadowMatrix(defaultShadowMatrix)
+          setDropShadowMatrix(defaultShadowMatrix, dropShadowComponent.entity, entity)
           continue
         }
 
+        const halfPoint = 0.5
+        const centerCorrectedDist = Math.max(intersected.distance - dropShadowComponent.center.y * halfPoint, 0)
```
AidanCaruso, line 256: True center would not be multiplied by a "halfpoint", this is a stopgap measure because casting from the true bottom of objects would give the avatar a tiny shadow. Going to fix possibly by adding the distance scaling to the start radius.

speigg (changes requested):
Just tested this branch w/ puttclub AR, and this change causes the ball to lose it's shadow. somehow.

AidanCaruso:
> Just tested this branch w/ puttclub AR, and this change causes the ball to lose it's shadow. somehow.

The shadow is there, it's just super tiny. Because I'm casting from the bottom of objects meaning the "distance" is zero and because the final size is the product of the distance and the radius and not the sum it's so small you can't see it. My mistake, time to fix!

speigg:
Yes, the final shadow radius when distance is zero should be close to the actual radius of the object.

AidanCaruso:
> Yes, the final shadow radius when distance is zero should be close to the actual radius of the object.

I felt suspicious taht I'd made some basic mistake there. Oh well, fixed now.

packages/engine/src/scene/systems/ShadowSystem.ts
```diff
@@ -223,34 +237,38 @@ export default async function ShadowSystem(world: World) {
       for (const entity of dropShadowComponentQuery()) {
         const dropShadowComponent = getComponent(entity, DropShadowComponent)
 
-        if (!dropShadowComponent.center) continue
-
-        const groupComponent = getComponent(entity, GroupComponent)
+        if (!entityExists(dropShadowComponent.entity)) {
+          removeComponent(entity, DropShadowComponent)
```
HexaField, line 241: shouldn't we add a `continue` statement here?

HexaField (changes requested):

speigg (approved):

PR #7602, opened 2023-02-19
Fix Instancing Component

dinomut1:
Fixes instancing component: 
-Grass is rendered correctly with the logarithmic depth buffer
-Node editor reimplemented with hyperflux
-Instancing component reimplemented with defineComponent schema

### DIFF
--- a/packages/engine/src/scene/functions/loaders/InstancingFunctions.ts
+++ b/packages/engine/src/scene/functions/loaders/InstancingFunctions.ts
@@ -15,13 +15,16 @@ import {
   PlaneGeometry,
   Quaternion,
   RawShaderMaterial,
+  ShaderChunk,
+  ShaderMaterial,
   Texture,
   Vector2,
   Vector3
 } from 'three'
 import matches from 'ts-matches'
 
-import { defineAction, dispatchAction } from '@xrengine/hyperflux'
+import { EntityUUID } from '@xrengine/common/src/interfaces/EntityUUID'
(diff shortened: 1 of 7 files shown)


packages/engine/src/scene/functions/loaders/InstancingFunctions.ts
```diff
@@ -72,8 +77,14 @@ export const GRASS_PROPERTIES_DEFAULT_VALUES: GrassProperties = {
   bladeHeight: { mu: 0.3, sigma: 0.05 },
   bladeWidth: { mu: 0.03, sigma: 0.01 },
   joints: 4,
-  grassTexture: 'https://resources-dev.etherealengine.com/assets/grass/blade_diffuse.jpg',
-  alphaMap: 'https://resources-dev.etherealengine.com/assets/grass/blade_alpha.jpg',
+  grassTexture: {
+    src: 'https://resources-dev.etherealengine.com/assets/grass/blade_diffuse.jpg',
```
HexaField, line 81: We should be sourcing these from the repo rather than the engine in case they are unreachable

HexaField (approved):
Nice work, let's consider doing this in a follow up

packages/engine/src/scene/functions/loaders/InstancingFunctions.ts
HexaField: https://github.com/XRFoundation/XREngine/issues/7605

PR #7604, opened 2023-02-19
Image Compression + Conversion

dinomut1:
Model Transform Improvements:
- add uastc compression option for basisu
- add reorder transform function

Image Convert Service:
- add feathers service for operating sharp (https://sharp.pixelplumbing.com/) to resize and convert images
- add "convert" context menu option in studio file browser panel

Reimplement file browser panel using hyperflux


speigg:
@dinomut1 let's fix the security vulnerabilities w/ shell commands that code scanning picked up

PR #7615, opened 2023-02-20
Fix component reactivity

speigg:
## Summary

This PR addresses superfluous state invalidations that were causing some reactors to re-run unnecessarily, including it's side-effects (after making the drop-shadow reactor reactive to changes in the GroupComponent to address the issue described in the note below, these superfluous state invalidations became evident).

More details on this hookstate issue here: https://github.com/avkonst/hookstate/issues/299

The solution involves placing all component instances in their own hookstate proxy, and reactively keeping track of component additions/removals in a stateful Component.existenceMap. With each component instance isolated in this way, changes to component data should only affect the reactors that are actually using that data. 

Note: 
This PR also cleans up some issues w/ the drop-shadow reactor, e.g., a race condition where it would generate a shadow too small for an avatar before the avatar had a chance to load it's model.


speigg:
gotta fix a failing test

hanzlamateen (commented):
1. Not sure if it is related, but I get following hookstate error when I change a scene from `Scene` from Apartment to Sky Station.
image

2. Similarly when I click on 'Portal - Sky Station Interior' & then 'Portal - Sky Station Exterior' in heirarchy panel then it fails due to hookstate error.
image
image

speigg:
@hanzlamateen  thank you that's very helpful

AidanCaruso:
In an apartment scene in which I have placed 6 cyberbots I've noticed low framerates after loading the models in, this was not an issue on dev, and I switched between the two to check, going to triple check now because I'm not immediately sure why this would be happening.
image
image

AidanCaruso:
Also getting errors similar to Hanzla when duplicating objects 
image

packages/engine/src/scene/systems/ShadowSystem.ts
```diff
@@ -171,108 +184,86 @@ export default async function ShadowSystem(world: World) {
...
         }
 
-        const centerCorrectedDist = Math.max(intersected.distance - dropShadowComponent.center.y, 0.0001)
+        const centerCorrectedDist = Math.max(intersected.distance - dropShadow.center.y, 0.0001)
 
         //arbitrary bias to make it a bit smaller
         const sizeBias = 0.3
-        const finalRadius =
-          sizeBias * dropShadowComponent.radius + dropShadowComponent.radius * centerCorrectedDist * 0.5
+        const finalRadius = sizeBias * dropShadow.radius + dropShadow.radius * centerCorrectedDist * 0.5
```
AidanCaruso, line 258: Super minor suggestion, but would it be helpful to use math.min to cap the final radius to some value, maybe 2*dropShadow.radius? This would prevent shadow planes from becoming enormous as distance increases.

speigg:
Low framerates when models are dropped in seems related to raycasts for drop shadows. We need to enable bounding volume hierarchies by default to speed up CPU raycasts. Currently the default is to have the disabled. I'll invert that.

HexaField:
> Low framerates when models are dropped in seems related to raycasts for drop shadows. We need to enable bounding volume hierarchies by default to speed up CPU raycasts. Currently the default is to have them disabled. I'll invert that.

We also should not do any drop shadows logic in the editor

speigg:
Gonna merge this in; 

In separate PRs:
- Will update default `generateBVH` property to true, and update our scenes
- Will update useComponent to suspend/resume react components (this will take care of some of the errors we see in the console)

AidanCaruso:
> > Low framerates when models are dropped in seems related to raycasts for drop shadows. We need to enable bounding volume hierarchies by default to speed up CPU raycasts. Currently the default is to have them disabled. I'll invert that.
> 
> We also should not do any drop shadows logic in the editor

This is actually a super good point.

AidanCaruso (approved):

PR #7627, opened 2023-02-24
Use React Devtools to introspect Hyperflux Reactors

speigg:
## Summary

You can now use React DevTools to introspect all running Hyperflux Reactors !


packages/engine/src/ecs/functions/ComponentFunctions.ts
```diff
@@ -201,6 +201,8 @@ export const setComponent = <C extends Component>(
     Component.mapState[entity].set(c)
     bitECS.addComponent(world, Component, entity, false) // don't clear data on-add
     if (Component.reactor) {
+      if (!Component.reactor.name || Component.reactor.name === 'reactor')
+        Object.defineProperty(Component.reactor, 'name', { value: `${Component.name}Reactor` })
```
HexaField, line 205: could we put the entity number in here too?

HexaField (approved):

PR #7715, opened 2023-03-09
Download project as ZIP in admin/projects

### LINKED ISSUE #7643: Download Project Button in Admin Panel
download-project-button

Add a button in the Project Panel to download a project from the server. This will help with cloud development and moving content between deployments

AidanCaruso:
## Summary

Adds a download project button to the admin/projects. Getting project directories on the file server recursively is done through an extension to the get function of the `file-browser` feathers service. This adds the `JSZip` package for zipping the returned URLs and the `save-as` package as a convenience for downloading. 
image

As a side note, testing with the putt club project, zipping takes roughly 15 seconds on my computer after pressing the download button. It might be worthwhile to add an indicator while this is happening.

## References

closes #7643

## Checklist

## QA Steps


packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -79,18 +80,32 @@ export class FileBrowserService implements ServiceMethods<any> {
   async get(directory: string, params?: UserParams): Promise<Paginated<FileContentType>> {
     if (!params) params = {}
     if (!params.query) params.query = {}
-    const { $skip, $limit, storageProviderName } = params.query
+    const { $recursive, $skip, $limit, storageProviderName } = params.query
 
     delete params.query.storageProviderName
     const skip = $skip ? $skip : 0
     const limit = $limit ? $limit : 100
+    const recursive = $recursive
```
HexaField, line 88: feathers pagination properties begin with $, we dont' (and shouldnt) follow this convention for other APIs. this can just be `recursive`

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -79,18 +80,32 @@ export class FileBrowserService implements ServiceMethods<any> {
...
 
     const storageProvider = getStorageProvider(storageProviderName)
     const isAdmin = params.user && params.user?.scopes?.find((scope) => scope.type === 'admin:admin')
     if (directory[0] === '/') directory = directory.slice(1) // remove leading slash
     if (params.provider && !isAdmin && directory !== '' && !/^projects/.test(directory))
       throw new Forbidden('Not allowed to access that directory')
     let result = await storageProvider.listFolderContent(directory)
+    let resultClone = [] as FileContentType[]
```
HexaField, line 96: this will only fetch results one more layer, projects can be nested deeoer.

instead replace L#95 with `let result = await storageProvider.listObjects(directory, recursive)`

HexaField (changes requested):

packages/server-core/src/media/file-browser/file-browser.class.ts
AidanCaruso: Got it, little experience with feathers so I was just following the pattern I saw.

packages/server-core/src/media/file-browser/file-browser.class.ts
AidanCaruso: Ah I didn't realize it was that simple. To get deeper layers I was using the loop below.

AidanCaruso: Wow. I didn't realize it was that simple. Here I wrote an entire loop to manually go through those directories and push them.

dinomut1 (dismissed):
This is recursively downloading every file in the project before zipping. This is going to result in many unnecessary feathers calls and we miss out on compressing network traffic. The project should be zipped serverside and then returned to the client with one feathers call

AidanCaruso:
> This is recursively downloading every file in the project before zipping. This is going to result in many unnecessary feathers calls and we miss out on compressing network traffic. The project should be zipped serverside and then returned to the client with one feathers call

Just to make sure I understand correctly, this would require creating a separate specific feathers service that creates a zip from the project's directories and rather than passing every directory to the client we just pass the zip file.

HexaField:
> Just to make sure I understand correctly, this would require creating a separate specific feathers service that creates a zip from the project's directories and rather than passing every directory to the client we just pass the zip file.

Yes.

speigg:
We should differentiate between download and update icons

mrhegemon (approved):

packages/client-core/src/common/services/FileBrowserService.ts
```diff
@@ -82,6 +82,7 @@ export const FileBrowserService = {
 
     const params = {
       query: {
+        recursive: false,
```
HexaField, line 85: please remove

HexaField (changes requested):
Thinking about this a bit more, we should NOT be putting the zip inside the project folder, as this will likely propagate into git repositories, which will also slow pushing/pulling a lot. Let's put it in a ~/temp directory in the storage provider and we can periodically clean it up.

AidanCaruso:
> Thinking about this a bit more, we should NOT be putting the zip inside the project folder, as this will likely propagate into git repositories, which will also slow pushing/pulling a lot. Let's put it in a ~/temp directory in the storage provider and we can periodically clean it up.

I agree, I can't believe I overlooked this. I am now dropping them into a directory labeled `/temp`.

HexaField (approved):

PR #7739, opened 2023-03-12
fix: do not remove last identity provider

### LINKED ISSUE #7575: [Bug]: Users can delete last identity-provider
People are able to delete their last identity-provider, which should not be allowed, since that will make it impossible to log into their account if they log out.

aditya-mitra:
## Summary

- Do not show the OAuth Identity Provider for removal if it is the only one present.
- Also, added the check in server-core along with corresponding tests.
- Refactored `identity-provider service` tests to use `userId` instead of `providers` array. ([assert should not be done inside a loop](https://stackoverflow.com/questions/41832594/should-you-do-an-assert-statement-inside-a-loop))

## References

Closes #7575 

## Checklist

## QA Steps

### DIFF
--- a/packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
+++ b/packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
@@ -93,6 +93,7 @@ const ProfileMenu = ({ hideLogin, allowAvatarChange, isPopover, changeActiveMenu
     (authState.linkedin && !oauthConnectedState.linkedin) ||
     (authState.twitter && !oauthConnectedState.twitter)
 
+  /**allow removing social logins if there are at least 2 social logins connected*/
   const removeSocial = Object.values(oauthConnectedState).filter((value) => value).length > 1
 
   // const loadCredentialHandler = async () => {
--- a/packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
+++ b/packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
@@ -93,13 +93,7 @@ const ProfileMenu = ({ hideLogin, allowAvatarChange, isPopover, changeActiveMenu
     (authState.linkedin && !oauthConnectedState.linkedin) ||
     (authState.twitter && !oauthConnectedState.twitter)
 
-  const removeSocial =
-    (authState?.discord && oauthConnectedState.discord) ||
-    (authState.facebook && oauthConnectedState.facebook) ||
-    (authState.github && oauthConnectedState.github) ||
-    (authState.google && oauthConnectedState.google) ||
-    (authState.linkedin && oauthConnectedState.linkedin) ||
-    (authState.twitter && oauthConnectedState.twitter)
+  const removeSocial = Object.values(oauthConnectedState).filter((value) => value).length > 1
 
   // const loadCredentialHandler = async () => {
   //   try {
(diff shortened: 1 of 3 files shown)


HexaField (changes requested):
This looks good, we should also ensure this in the backend service.

HexaField (approved):

packages/server-core/src/user/identity-provider/identity-provider.test.ts
```diff
@@ -69,37 +70,61 @@ describe('identity-provider service', () => {
...
-
-      assert.ok(item, 'Identity provider item is found')
-    }
+    const item = await app.service('identity-provider').find({
+      query: {
+        userId
+      }
+    })
+
+    assert.ok(item, 'Identity provider item is found')
```
barankyle, line 91: If we're going to make all of these providers on the same user in this test, we should be testing that there are three identity providers in what's returned from identity-provider.find, just to make sure it's correct.

packages/server-core/src/user/identity-provider/identity-provider.test.ts
```diff
@@ -69,37 +70,61 @@ describe('identity-provider service', () => {
...
   })
 
   it('should remove identity providers', async () => {
```
barankyle, line 94: Looking at this test, I would like it to be two tests. One should test removing one identity-provider by ID, and should check that that i-p is no longer present while the other are. The second would remove by userId as this test is, and does an explicit check that there are no more identity-providers for that user.

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
```diff
@@ -44,6 +44,18 @@ const checkIdentityProvider = (): any => {
   }
 }
 
+const checkOnlyIdentityProvider = () => {
```
barankyle, line 47: This doesn't work as intended if an identity provider is being deleted by ID; in that case, there is no query.userId, which causes a SQL query issue since it's trying to perform a find with a null value on userId. This hook should look up the provider, get the user's ID from that, and then use that in the .find() to determine how many identity-providers they have.

We actually shouldn't be supporting removing identity-providers in bulk, via query: { userId: <ID> }. I know that's how it was being done in the tests, but after thinking about it more, I don't think that's correct behavior, and the tests should be changed to delete identity-providers by ID. Currently, the only way identity-providers are removed in bulk is if a user deletes their account, and with that, we rely on SQL foreign key relationships to cascade that deletion to the identity-provider table. I don't think we ever want to delete multiple identity-providers at once.

packages/server-core/src/user/identity-provider/identity-provider.test.ts
aditya-mitra: Added this change!

packages/server-core/src/user/identity-provider/identity-provider.test.ts
aditya-mitra: Added this change!

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
aditya-mitra: > This doesn't work as intended if an identity provider is being deleted by ID; in that case, there is no query.userId, which causes a SQL query issue since it's trying to perform a find with a null value on userId. This hook should look up the provider, get the user's ID from that, and then use that in the .find() to determine how many identity-providers they have.

I have added this change!

> We actually shouldn't be supporting removing identity-providers in bulk, via query: { userId: }. I know that's how it was being done in the tests, but after thinking about it more, I don't think that's correct behavior, and the tests should be changed to delete identity-providers by ID. Currently, the only way identity-providers are removed in bulk is if a user deletes their account, and with that, we rely on SQL foreign key relationships to cascade that deletion to the identity-provider table. I don't think we ever want to delete multiple identity-providers at once.

Agreed! Multiple identity providers will only be internally deleted (that too in cascade), in cases if a user is deleting his/her account. If we are going to incorporate this change, maybe we should create another PR so that this and the current issue are both issues are resolved in their separate PRs.

barankyle: Having a separate issue sounds fine.

packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
```diff
@@ -93,13 +93,8 @@ const ProfileMenu = ({ hideLogin, allowAvatarChange, isPopover, changeActiveMenu
...
 
-  const removeSocial =
-    (authState?.discord && oauthConnectedState.discord) ||
-    (authState.facebook && oauthConnectedState.facebook) ||
-    (authState.github && oauthConnectedState.github) ||
-    (authState.google && oauthConnectedState.google) ||
-    (authState.linkedin && oauthConnectedState.linkedin) ||
-    (authState.twitter && oauthConnectedState.twitter)
+  /**allow removing social logins if there are at least 2 social logins connected*/
+  const removeSocial = Object.values(oauthConnectedState).filter((value) => value).length > 1
```
barankyle, line 97: Forgot to mention this the other day, this should be >= 1. The list of OAuth identity providers that you have connected is the list of identity providers you can remove; there's a check in the client code that will not even attempt to remove the IP if it's the only one. If this were just > 1, then a user with a single connected OAuth IP wouldn't see the IP that they had connected.

aditya-mitra: Right. I have added this change!

barankyle (approved):

Issue #7645, opened 2023-02-28
[EPIC] Upgrade to feathers 5.0.0


speigg:
https://github.com/feathersjs/feathers/releases/tag/v5.0.0

- [x] Upgrade all packages to 5.0.0. Test current code base.
- [x] Removed unused hooks
- [x] #7726
- [x] https://github.com/EtherealEngine/etherealengine/issues/7802 https://feathersjs.com/api/koa.html
- [x] #8871
- [x] Create migrations to ensure tables creation and ensure seeders work fine.
- [x] Checklist of database based services:
  - [x] analytics
  - [x] authentication-setting
  - [x] avatar
  - [x] aws-setting
  - [x] bot #8573
  - [x] bot-command
  - [x] build-status
  - [x] channel #8665
  - [x] channel-user #8481 
  - [x] chargebee-setting
  - [x] client-setting
  - [x] coil-setting
  - [x] email-setting
  - [x] github-repo-access #8286
  - [x] helm-setting
  - [x] identity-provider #8423
  - [x] instance #8615
  - [x] instance-server-setting
  - [x] instance-attendance #8399
  - [x] instance-authorized-user #8400
  - [x] invite #8497
  - [x] invite-type #8421
  - [x] location #8284
  - [x] location-settings #8284
  - [x] location-type #8284
  - [x] location-admin #8392
  - [x] location-authorized-user #8386
  - [x] location-ban #8381
  - [x] login-token #8452
  - [x] match-instance
  - [x] match-user
  - [x] message #8572
  - [x] project #8507 
  - [x] project-permission-type #8363
  - [x] project-permission #8456
  - [x] recording #8424
  - [x] recording-resource #8410
  - [x] redis-setting
  - [x] route
  - [x] scope #8480
  - [x] scope-type #8299
  - [x] server-setting
  - [x] static_resource #8457
  - [x] task-server-setting
  - [x] user #8368
  - [x] user-api-key #8338
  - [x] user-kick #8446
  - [x] user-relationship #8447
  - [x] user-relationship-type #8448
  - [x] user-setting #8449
- [x] Checklist of non-database based services:
  - [x] accept-invite #8663
  - [x] asset-library #8680
  - [x] auth-management #8683
  - [x] email #8684
  - [x] file-browser #8737
  - [x] instance-provision #8765
  - [x] login #8701
  - [x] magic-link #8724
  - [x] model-transform #8729
  - [x] project-setting #8742
  - [x] scene #8743
[...]

PR #7744, opened 2023-03-13
Changes to introduces configs for feather sql instead of sequelize

hanzlamateen:
## Summary

This PR includes changes for moving to feathers 5.0 and using typebox based schemas, resolvers, validators and knex instead of sequelize. 

The PR mainly targets `route`, `analytics`, `chargebee-setting`, `task-server-setting` & `coil-setting` service. It also updates the table creation of this service from sequelize to knex based migration. 

Also the seeding of this table is updated to knex based seed, although we are calling it through our code instead of knex seed command. We are not using knex seed command because it requires us to specify file names in the form of string array which is not preferred at the moment. Another way is to have all seed files inside a folder and then knex seeder will call them in alphabetical order which is again not desired.

### Create Migration File
To create migration file, you will need to `cd packages/server-core` & then run `npm run migrate:make -- [NAME]` i.e. `npm run migrate:make -- test`
It will generate file i.e. `20230407060205_test.ts` in root of `server-core` which can then be manually copied into respective folder.

## References

#7645

## Checklist

## QA Steps


speigg:
@cla-bot check

packages/server-core/knexfile.ts
```diff
@@ -0,0 +1,46 @@
+import fs from 'fs'
+import type { Knex } from 'knex'
+import path from 'path'
+
+import appConfig from '@etherealengine/server-core/src/appconfig'
+
+const migrationsDirectories = ['./migrations']
```
HexaField, line 7: could we replace this with a "find recursive" such that we can put each migration file along with it's service in a migrations folder, otherwise this folder will get big very quickly

HexaField (changes requested):

packages/server-core/knexfile.ts
HexaField: discussed this, not a good idea considering it will be far easier to see all the changes if they are in one place

HexaField (approved):

mrhegemon (approved):
Works great from a fresh database. We should probably test on a migration.

hanzlamateen:
> Works great from a fresh database. We should probably test on a migration.

I have tested it on a local database initialized using through dev branch and then using this branch. Works fine.

Issue #7726, opened 2023-03-11
Research how feathers does it's tests and present and implement it in the user service


HexaField:
Since we depend on the database for our server tests, it makes it essentially impossible to do proper unit testing on our feathers services. We want to be able to clear the database inbetween each test. Feathers likely has a tool & pattern for this already, so we should research what that is and implement it to establish the pattern. The `user` table is a good table to do this in.

PR #7746, opened 2023-03-13
feat: bulk insert in seeder

aditya-mitra:
## Summary

Using multiple insert, featherjs do will a bulk insert of all the values into MySQL. This will be faster than sequential insertion of templates in loop.

## References

Refs #7726 

## Checklist

## QA Steps


packages/server-core/src/seeder.ts
```diff
@@ -7,49 +8,18 @@ import config from './appconfig'
...
-      }
-    }
+  for (const config of seederConfig) {
+    if (!config.path || !config.templates) return
+
+    const templates = config.templates
+
+    const service = app.service(config.path as any) as FeathersService
+
+    await service.create(templates)
```
HexaField, line 21: This seems like it removes the check to see if the table has already been seeded (`isSeeded = result.total > 0`) which must run when `prepareDb` is true, as the `Prepare Database` mode is a migration tool for seeding new tables upon updating a deployment. Please add back that logic. We should also be pushing `service.create(templates)` into an array of promises that is then all awaited together with `await Promise.all()` to parallelise and speed up.

HexaField (changes requested):

packages/server-core/src/seeder.ts
aditya-mitra: Update: I have found a way to implement these changes, however I am facing a little problems. Please allow me time till tomorrow 12 pm IST to complete it.

aditya-mitra: I have reintroduced the `prepareDb` changes (checking if the record exists before insert) which were missing. I have also used `Promise.all()` for bulk inserts and single inserts.
Some of _.seed.ts_. files faced problems with bulk inserts, so I added an option `insertSingle` in some _.seed.ts_ files which will insert the templates one by one instead of bulk insert.

HexaField (approved):

HexaField:
@cla-bot check

PR #7747, opened 2023-03-13
feat: reseed db before tests

### LINKED ISSUE #7726: Research how feathers does it's tests and present and implement it in the user service

aditya-mitra:
## Summary

Using an app setting `testModeForceRefresh = true`, the database can be seed in between tests.
Here, this setting has been used in _user.test.ts_

## References

Closes #7726 

## Checklist

## QA Steps


packages/server-core/src/sequelize.ts
```diff
@@ -33,7 +33,9 @@ export default (app: Application): void => {
       promiseReject = reject
     })
 
-    app.setup = async function (...args: any) {
+    app.setup = async function (...args) {
+      const testModeForceRefresh = app.get('testModeForceRefresh')
```
HexaField, line 37: rather than setting this manually, we can rely on `appConfig.testEnabled`, such that it is automatically applied every time `createFeathersExpressApp` is run

HexaField (changes requested):
replicating this locally, I get `"message": "Table 'etherealengine.analytics' doesn't exist",`

My guess is this doesn't work for multiple tests.

packages/server-core/src/sequelize.ts
aditya-mitra: I tried to do this initially but encountered some problems. DB is also reseeded before tests (in `npm run pretest`) which requires [`process.exit(0)` here](https://github.com/EtherealEngine/etherealengine/blob/dev/packages/server-core/src/sequelize.ts#L104). Using `appConfig.testEnabled` will have some problems.

aditya-mitra:
> replicating this locally, I get `"message": "Table 'etherealengine.analytics' doesn't exist",`
> 
> My guess is this doesn't work for multiple tests.

@HexaField Can you please tell me the steps you followed to replicate this?

This is because I am testing this on a fresh database and not getting any error.
I also added this _reseed_ in a few more tests in packages/server-core to check if it works across different test cases. It seemed to be working.

image

Can you please recheck it once?

packages/server-core/src/sequelize.ts
HexaField: Ahh I see. Let's use the appConfig instead of feathers internal state. We can add a new configuration option with the desired functionality.

HexaField:
> @HexaField Can you please tell me the steps you followed to replicate this?
> 
> This is because I am testing this on a fresh database and not getting any error. I also added this _reseed_ in a few more tests in packages/server-core to check if it works across different test cases. It seemed to be working.
> 
> image
> 
> Can you please recheck it once?

I have copied your code exactly, and added 
```ts


This is the result:
```
[1678916225094] ERROR: Duplicate key name 'identity_provider_id'
    component: "server-core:sequelize"
    err: {
      "type": "DatabaseError",
      "message": "Duplicate key name 'identity_provider_id'",
```

aditya-mitra:
@HexaField I found the error (it was a problem with a test) and have fixed it.

> ```ts
> ```ts
>   if (appConfig.testEnabled) 
>     app.set('testModeForceRefresh', true)
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> to createFeathersExpressApp in line 146 to ensure it reseeds before every test
> ```

I have also added the setting to reseed db in `appConfig` itself using above method. **All tests will now reseed db before running their tests.**

packages/server-core/src/sequelize.ts
aditya-mitra: Added this change!

HexaField:
> @HexaField I found the error (it was a problem with a test) and have fixed it.
> 
> I have also added the setting to reseed db in `appConfig` itself using above method. **All tests will now reseed db before running their tests.**

Fantastic, let's replace the usage of `testModeForceRefresh` with `appConfig.testEnabled` directly and this should be good to go

aditya-mitra:
> Fantastic, let's replace the usage of testModeForceRefresh with appConfig.testEnabled directly and this should be good to go

Done with these changes!

mrhegemon (approved):

HexaField (changes requested):
in server `start.ts` L27 and server-core `seeder.tsx` L63 we should add `&& !config.testEnabled` to both conditionals

HexaField (approved):

aditya-mitra:
@cla-bot check

PR #7782, opened 2023-03-21
Instanced LOD System

### LINKED ISSUE #7785: : Create InstancingRegistry data structure.

### LINKED ISSUE #7786: Create InstancedLOD component

### LINKED ISSUE #7787: Create InstancedLODSystem

dinomut1:
BabylonJS has a more robust instanced LOD system that we will use as a basis. The difference between Babylon's instancing and the "thin" threejs instancing is detailed here: 
https://discourse.threejs.org/t/lod-instancing/20524/2

We will implement the more robust instancing and LOD system using a LODComponent and LODSystem that work together to keep track of each instance's distance from the camera, and switch different instanced LODs on and off based on said distance. These components should be generated automatically when a model with instanced geometry is loaded into a scene. 

The components should be configurable in the editor with some menu interface to add, remove, and edit LOD levels for every mesh in the scene, instanced or not. This UX should be embedded in the properties of the ModelComponent that contains the lod instances

We also need a GLTF extension for defining instanced LODs in Blender, modifying and saving LODs in engine, and successfully passing instanced LODs through Model Transforms without loss of information.

The same goes for a USD plugin that can define instanced LODs with equivalent functionality to the GLTF extension

This pull request closes #7785 
closes #7786 
closes #7787


packages/engine/src/scene/functions/loaders/LODFunctions.ts
```diff
@@ -97,7 +98,12 @@ export function processLoadedLODLevel(entity: Entity, index: number, model: Mesh
         component.instanceLevels.set(new InstancedBufferAttribute(new Uint8Array([index]), 1))
       } else {
         //if the lodComponent does have a matrix defined, set the loaded model's matrix to it
-        model.matrix.fromArray(component.instanceMatrix.value.array)
+        if (loadedModel) {
+          loadedModel.parent?.add(model)
+          model.matrix.copy(loadedModel.matrix)
+          model.updateMatrixWorld(true)
```
HexaField, line 104: `model.updateMatrixWorld(true)` will call `model.updateMatrix()` internally

HexaField (changes requested):

packages/engine/src/scene/SceneClientModule.ts
```diff
@@ -30,6 +31,11 @@ export function SceneClientModule() {
       type: SystemUpdateType.UPDATE_LATE,
       systemLoader: () => Promise.resolve({ default: LightSystem })
     },
+    {
+      uuid: 'ee.engine.LODSystem',
+      type: SystemUpdateType.PRE_RENDER,
```
HexaField, line 36: can be POST_RENDER

HexaField (changes requested):

2dels (approved):
Great job!

PR #7830, opened 2023-03-29
Capture Styling Updates

jerknose:
New tailwind theming styling and layout. Using Daisy Ui components.


speigg (approved):

speigg:
As long as we have a plan to integrate this with our current theme system, looks good to me!

Issue #5997, opened 2022-05-09
[Epic] Unit tests for all server-core services


HexaField:
- [ ] analytics
- [ ] analytics-setting
- [ ] asset-library
- [ ] authManagement
- [ ] authentication
- [ ] authentication-setting
- [ ] avatar
- [ ] aws-setting
- [ ] bot
- [ ] bot-command
- [ ] channel
- [ ] channel-type
- [ ] chargebee-setting
- [ ] client-setting
- [ ] discord-bot-auth
- [ ] email
- [ ] email-setting
- [x] file-browser
- [ ] file-browser/upload
- [ ] game-server-setting
- [ ] gameserver-load
- [ ] gameserver-provision
- [ ] gameserver-subdomain-provision
- [ ] github-app
- [ ] group
- [ ] group-user
- [ ] group-user-rank
- [x] identity-provider
- [x] instance
- [ ] instance-attendance
- [ ] instance-authorized-user
- [ ] instance-provision
- [x] invite
- [ ] invite-type
- [ ] location
- [ ] location-admin
- [ ] location-ban
- [ ] location-settings
- [ ] location-type
- [ ] login
- [ ] login-token
- [ ] magic-link
- [ ] match-instance
- [ ] match-ticket
- [ ] match-ticket-assignment
- [ ] match-user
- [ ] message
- [ ] message-status
- [ ] party
- [ ] party-user
- [ ] portal
- [ ] project
- [ ] project-build
- [ ] project-invalidate
- [ ] project-setting
- [ ] redis-setting
- [ ] route
- [ ] route-activate
- [ ] routes-installed
- [ ] rtc-ports
- [x] #7740
- [x] scenes
- [ ] scope
- [ ] scope-type
- [ ] server-setting
- [ ] sms
- [ ] static-resource
- [ ] static-resource-type
- [ ] testbot
- [ ] upload-asset
- [x] user
- [ ] user-api-key
- [ ] user-relationship
- [ ] user-relationship-type
- [ ] user-role
- [ ] user-settings
- [ ] visualisation

PR #7837, opened 2023-03-29
tests: unit tests for route

aditya-mitra:
## Summary

Unit Tests for server-core's route service

## References

Refs #5997 

## Checklist

## QA Steps


packages/server-core/src/route/route/route.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  it('should not be activated by default (the installed project)', async () => {
+    const route = await app.service('route').find({ query: { project: newProject } })
+    assert.equal(route.total, 0)
+  })
+
+  it('should activate a route', async () => {
+    const activateResult = await app
+      .service('route-activate')
+      .create({ project: newProject, route: newProject, activate: true }, params)
```
aditya-mitra, line 55: This does not seem to create a route because [routeToActivate here](https://github.com/EtherealEngine/etherealengine/blob/dev/packages/server-core/src/route/route/route.service.ts#L68) seems to be be false. 

How can I add the routes in the generated project's `xrengine.config.ts` such that this value becomes true?

packages/server-core/src/route/route/route.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    await app.service('project').Model.destroy({ where: { name: projectName } })
+  } catch (e) {}
+}
+
+const updateXREngineConfigForTest = (newProject: string) => {
+  const testXREngineConfig = `
+  import type { ProjectConfigInterface } from '@etherealengine/projects/ProjectConfigInterface'
+
+  const config: ProjectConfigInterface = {
+    onEvent: undefined,
```
HexaField, line 28: we can make this leaner by removing all properties except `routes`

HexaField (changes requested):

packages/server-core/src/route/route/route.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+const cleanup = async (app: Application, projectName: string) => {
+  const projectDir = path.resolve(appRootPath.path, `packages/projects/projects/${projectName}/`)
+  deleteFolderRecursive(projectDir)
+  try {
+    await app.service('project').Model.destroy({ where: { name: projectName } })
+  } catch (e) {}
+}
+
+const updateXREngineConfigForTest = (newProject: string) => {
```
HexaField, line 23: this argument should be `customRoute`

packages/server-core/src/route/route/route.test.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  after(async () => {
+    await destroyEngine()
+    await cleanup(app, newProject)
+  })
+
+  it('should find the installed project routes', async () => {
+    newProject = `test-project-${uuid()}`
+    await app.service('project').create({ name: newProject }, params)
+    updateXREngineConfigForTest(newProject)
```
HexaField, line 70: let's define a different value rather than reusing newProject here

HexaField (changes requested):

packages/server-core/src/route/route/route.test.ts
aditya-mitra: I have used `testRoute` now.

packages/server-core/src/route/route/route.test.ts
aditya-mitra: added this change!

packages/server-core/src/route/route/route.test.ts
aditya-mitra: yes, not all properties are required by the `xrengine.config`.
added this change!

packages/server-core/src/route/route/route.test.ts
HexaField: We want two different variables, one for the project name and one for the route

aditya-mitra: Understood. I have used 2 variables - `testProject` and `testRoute` having different values and testing them.

HexaField:
Please resolve typescript errors

aditya-mitra:
I have removed the ts error.
There seems to be some DB connection error in the `branch-build/compile-codebase` job.

speigg (approved):
Awesome!

HexaField (changes requested):
My only concern here is whether or not this will affect vscode nested repo searching, which is quite brittle at the moment. Let's add a gitignore for the generated project folder.

aditya-mitra:
> My only concern here is whether or not this will affect vscode nested repo searching, which is quite brittle at the moment. Let's add a gitignore for the generated project folder.

image

I think it is automatically added to gitignore (it seems to be defined in the root gitignore)

speigg (approved):

speigg (approved):

HexaField (approved):

PR #7896, opened 2023-04-14
add support for trimesh collider

dinomut1:


packages/engine/src/physics/classes/Physics.ts
```diff
@@ -151,7 +151,7 @@ function createColliderDesc(
   // if (!colliderDescOptions.shapeType && mesh.geometry.type === 'CylinderGeometry')
   //   colliderDescOptions.shapeType = ShapeType.Cylinder
 
-  if (typeof colliderDescOptions.shapeType === 'undefined') return
+  if (!overrideShapeType && typeof colliderDescOptions.shapeType === 'undefined') return
```
HexaField, line 154: is this necessary if we are guaranteed to be passing shapeType in?

HexaField (changes requested):

packages/engine/src/scene/components/ColliderComponent.ts
```diff
@@ -108,8 +108,9 @@ export const ColliderComponent = defineComponent({
     const isLoadedFromGLTF = useOptionalComponent(entity, GLTFLoadedComponent)
     const groupComponent = useOptionalComponent(entity, GroupComponent)
 
+    const isMeshCollider = [ShapeType.TriMesh, ShapeType.ConvexPolyhedron].includes(colliderComponent.shapeType.value)
```
HexaField, line 111: shouldn't we also be checking for ModelComponent here?

HexaField (changes requested):

packages/engine/src/physics/classes/Physics.ts
dinomut1: We are not guaranteed to be passing shapeType in. ColliderDescOptions defines shapeType as possibly undefined

packages/engine/src/scene/components/ColliderComponent.ts
dinomut1: We don't use ModelComponent at any point, just the GroupComponent

dinomut1: We should leave the door open to allow any kind of component to supply the mesh that is used

PR #7904, opened 2023-04-17
Expand Particle System Component UX

dinomut1:
## Summary
three quarks allows dynamic behavior to be configured using bezier curves and gradients. This PR adds UI components that let the user configure beziers and gradients


speigg (approved):
Nice!

PR #7908, opened 2023-04-18
Optimize Query Reactors

speigg:
A lot of low hanging fruit for React.memo() here. Significantly faster reactor updates, engine-wide.


packages/engine/src/ecs/functions/SystemFunctions.tsx
jerknose, line 1: We should be using destructured imports to help with tree shaking. import React, {memo, FC} from 'react'

mrhegemon (approved):
Memo!

HexaField (approved):

PR #7909, opened 2023-04-19
Cron job file cleanup

usaurabh02:
## Summary

Created cronjob for temporary file cleanup in task server. Code for Issue scheduled cronjob to clean up /temp folder in storage provider #7771 

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/taskserver/src/cron-tempfile-deletion.ts
+++ b/packages/taskserver/src/cron-tempfile-deletion.ts
@@ -1,0 +1,46 @@
+import { ProjectInterface } from '@etherealengine/common/src/interfaces/ProjectInterface'
+
+export const getCronJobBody = (project: ProjectInterface, image: string): object => {
+  return {
+    apiVersion: 'cronjob-2',
+    kind: 'CronJob-tempfile-deletion',
+    metadata: {
+      name: `${process.env.RELEASE_NAME}`,
+      labels: {
+        'etherealengine/release': process.env.RELEASE_NAME
+      }
+    },
+    spec: {
+      schedule: '0 0 * * *', // This will run cron job at 12:00 AM everyday.
+      concurrencyPolicy: 'Replace',
+      successfulJobsHistoryLimit: 1,
+      failedJobsHistoryLimit: 2,
+      jobTemplate: {
+        spec: {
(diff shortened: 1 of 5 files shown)


packages/taskserver/src/cron-tempfile-deletion.ts
```diff
@@ -0,0 +1,46 @@
+import { ProjectInterface } from '@etherealengine/common/src/interfaces/ProjectInterface'
+
+export const getCronJobBody = (project: ProjectInterface, image: string): object => {
+  return {
+    apiVersion: 'cronjob-2',
+    kind: 'CronJob-tempfile-deletion',
```
barankyle, line 6: Are the apiVersion and kind fields a) necessary for this body and b) the correct format if so? The project-update cronjob getCronJobBody didn't seem to need those fields, likely because the k8BatchClient call with it is filling them in. But also, I don't recognize either of those field values. apiVersion is usually something like `v1` or `batch/v1beta1`, and kind for CronJobs is usually 'CronJob'. If those are values you filled in to distinguish it, know that they have very specific purposes and accepted values.

usaurabh02: Not necessarily needed here. I will remove those fields as I can see they are in the k8BatchClient. Thanks.

scripts/delete-tempfile-cron.ts
```diff
@@ -0,0 +1,83 @@
...
+            updateType: 'tag'
+          }
+        ]
+      }
+    })
+    const k8BatchClient = getState(ServerState).k8BatchClient
+    if (k8BatchClient)
+      for (const project of UpdateTempPruner.data) {
+        try {
+          await k8BatchClient.patchNamespacedCronJob(
```
barankyle, line 59: I think this should use listNamespacedCronJob to check if the cronjob exists, and then either patch it if it does, or create it if it does not.

barankyle (changes requested):
I don't see where delete-tempfile-cron.ts is ever called. It needs to be called somewhere in order to create/patch the cronjob. Taskserver could run this on startup in its start.ts.

PR #7988, opened 2023-05-17
Fixed and improved Docker build caching.

barankyle:
## Summary

Builder build process caching was not working because it was a multi-stage build.
The default inline cache only caches the final stage of a multi-stage build, which
almost always results in cache misses on the earlier stages. All builds in the builder
were not being cached at all.

Converted builds to explicitly use buildx. buildx build has support for automatically
pushing tags, so separate steps for tagging and pushing images have been removed.

In order to cache multi-stage builds, buildx has an option to cache everything. This
has to be done with a cache target of something other than inline cache (which includes
cache in the image). There is an option to use a Docker repository, but currently this
is not supported by ECR. For now, using S3 as the cache target (another built-in option)
for service images, and GH Actions cache for builder image.
An issue has been opened to switch to the repo for the service images once it's supported.

Commented out running npm install when each project is installed. This appeared to be causing
issues where some packages, particularly semver, were not fully installed properly by the
end, leading to errors when those packages were referenced. There is now a single npm
install after installing all projects.

## References

closes #_insert number here_

## Checklist

## QA Steps


scripts/run-builder.sh
```diff
@@ -19,6 +19,9 @@ npm run create-build-status
 BUILDER_RUN=$(tail -1 builder-run.txt)
 npm run install-projects >project-install-build-logs.txt 2>project-install-build-error.txt || npm run record-build-error -- --service=project-install
 test -s project-install-build-error.txt && npm run record-build-error -- --service=project-install
+cd packages/projects
```
HexaField, line 22: localProjectDirectory in the script that has been disabled points to packages/projects/projects, which im a bit confused about how it achieves `npm i` successfully for all the projects anyway, but wondering why this is not the same

.github/workflows/prod-deploy.yml
```diff
@@ -52,35 +52,29 @@ jobs:
...
+      - name: move package.json
+        run: mv package.json package.jsonmoved
       - name: npm-install 'cli' and 'aws-sdk'
         run: npm install cli aws-sdk
-      - name: Publish to Elastic Container Registry
-        run: bash scripts/publish_ecr_builder.sh prod $GITHUB_SHA $DOCKER_LABEL $PRIVATE_ECR $AWS_REGION
+      - name: restore package.json
+        run: mv package.jsonmoved package.json
+      - name: Expose GitHub Runtime
+        uses: crazy-max/ghaction-github-runtime@v2
```
hanzlamateen, line 62: This is used to access STORAGE_AWS_ACCESS_KEYs?

hanzlamateen (approved):
Looks good

.github/workflows/prod-deploy.yml
barankyle: It's actually to access some GitHub Actions environment variables used for caching Docker cache within GH Actions.

PR #8004, opened 2023-05-24
Peer Notification Cleanup

HexaField:
## Summary

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/client-core/src/networking/NetworkInstanceProvisioning.tsx
```diff
@@ -236,14 +240,30 @@ export const PartyInstanceProvisioning = () => {
...
+      NotificationService.dispatchNotify(`${props.name} ${t('common:toast.left')}`, { variant: 'default' })
+    }
+  }, [])
+
+  return <></>
+}
+
 export const InstanceProvisioning = () => {
   const networkConfigState = useHookstate(getMutableState(NetworkState).config)
+  const userNames = useHookstate(getMutableState(WorldState).userNames)
```
speigg, line 257: Usernames are not the same as connected peers or users; can we use a more appropriate source of truth?

speigg (changes requested):

packages/client-core/src/networking/NetworkInstanceProvisioning.tsx
speigg: We should be getting a list of the connected user ids, and mapping those into WorldNetworkUser reactors

packages/client-core/src/networking/NetworkInstanceProvisioning.tsx
```diff
@@ -236,14 +240,30 @@ export const PartyInstanceProvisioning = () => {
   return null
 }
 
+export const WorldNetworkPeer = (props: { userID: UserId; name: string }) => {
```
speigg, line 243: Reactively grab username here instead of passing down as prop, and only dispatch the notification in the effect when the userid has a username. Can add a boolean state to see if the user has not already been announced, and otherwise dispatch a notification that they changed their name.

packages/client-core/src/networking/NetworkInstanceProvisioning.tsx
HexaField: WorldState is derived from connected peers, it is a mapping of connected users to their names, this is the source of truth for this information.

speigg: The usernames map should be a source of truth for mapping user ids to a user name, and nothing more. Otherwise we are overloading it to represent networking state, and that's prone to error.

speigg: Alternatively, we can put the username map elsewhere, and just have connectedUserIds in the NetworkState

PR #8020, opened 2023-05-28
Jitter buffer

speigg:
Implements a jitter buffer with a default network delay of 100 ms, or about 6 simulation frames. Cleans up timing data.

### DIFF
--- a/packages/engine/src/ecs/classes/EngineState.ts
+++ b/packages/engine/src/ecs/classes/EngineState.ts
@@ -7,7 +7,7 @@ import { Entity } from './Entity'
 // TODO: #6016 Refactor EngineState into multiple state objects: timer, scene, world, xr, etc.
 export const EngineState = defineState({
   name: 'EngineState',
-  initial: {
+  initial: () => ({
     simulationTimestep: 1000 / 60,
 
     frameTime: Date.now(),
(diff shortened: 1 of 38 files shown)


packages/engine/src/ecs/classes/EngineState.ts
```diff
@@ -8,13 +8,16 @@ import { Entity } from './Entity'
 export const EngineState = defineState({
   name: 'EngineState',
   initial: {
```
HexaField, line 10: This should be an arrow function not an object, such that Date.now() are called every time EngineState is created, rather than cloning the Date.now() of when the module was loaded

packages/engine/src/physics/systems/PhysicsSystem.ts
```diff
@@ -154,7 +154,7 @@ const execute = () => {
 
   // step physics world
   const substeps = engineState.physicsSubsteps
-  const timestep = engineState.fixedDeltaSeconds / substeps
+  const timestep = engineState.simulationTimestep / 1000 / substeps
```
HexaField, line 157: can we make this a bit clearer by enclosing in `( )` brackets?

HexaField (changes requested):

speigg:
Made the changes you suggested. Merging now ;)

PR #8027, opened 2023-05-29
LOD Improvements

dinomut1:


packages/engine/src/scene/systems/LODSystem.tsx
```diff
@@ -45,8 +50,16 @@ function updateLOD(
   }
 }
 
+let engineState: typeof EngineState._TYPE | null
+function reactor() {
+  useEffect(() => {
+    engineState = getState(EngineState)
```
HexaField, line 56: This is unnecessary, you can just put `const engineState = getState(EngineState)` at the start of `execute`

HexaField (changes requested):

HexaField (approved):

PR #8049, opened 2023-06-02
fix media bugs with multiple peers to same user

HexaField:
## Summary

PeerMediaChannelState made assumptions about and was determinded by an algorithm to select specific peers to be shown in the user media windows. PeerMediaChannelState should hold an entry for every peer, as logic is dependent upon it to handle consumers & producers.

Moving the assumption to the media window component itself cleans up the logic a lot, and fixes the issues.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/client-core/src/components/UserMediaWindows/index.tsx
+++ b/packages/client-core/src/components/UserMediaWindows/index.tsx
@@ -2,51 +2,70 @@ import { useHookstate } from '@hookstate/core'
 import React from 'react'
 
 import { PeerID } from '@etherealengine/common/src/interfaces/PeerID'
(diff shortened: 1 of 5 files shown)


barankyle (approved):
When I tested it locally, I saw that it won't display user bubbles until they have some sort of media transmitting. Is that what we want, or do we want to keep it where it shows up in non-immersive mode no matter what?

HexaField:
> When I tested it locally, I saw that it won't display user bubbles until they have some sort of media transmitting. Is that what we want, or do we want to keep it where it shows up in non-immersive mode no matter what?

Was this for other users or just other peers? We want to show all users, but only other peers if they have media.

PR #8056, opened 2023-06-06
Update vite, json-tree, lazy load, manual chunks

jerknose:


barankyle:
FYI I resolved a couple of package.json merge conflicts after I merged my dependency cleanup PR in. It basically amounted to removing tailwindcss altogether from the root package.json, as AFAICT it's only needed in the client and ui packages (which have it in their deps), and removing vite from engine and matchmaking dependencies since neither of those are built with vite.

barankyle (approved):

PR #8060, opened 2023-06-07
Serve all client files from storage provider

### LINKED ISSUE #7990: Add ability to serve client entirely from Storage Provider
Currently, when we serve the client from the storage provider, it's not 100% served from there. dev.etherealengine.com points to the Kubernetes cluster, so the initial HTML is coming from the cluster; the vite build is just pointing every other built file to come from resources-dev.etherealengine.com. It would be more efficient in some ways to have the "serve client from storage provider" option point dev.etherealengine.com to the storage provider wholesale, and then not need a `resources-<stage>` subdomain for files, but rather have all resources come from the client subdomain.

To summarize, this would have CloudFront, etc. be configured to host as `<stage>.<domain>`, and not use a resources subdomain at all.

barankyle:
## Summary

Serving the client from the storage provider now serves everything from there. Previously,
the root index.html file was coming from the EKS cluster (api pod or client pod), while every
other client file was built to and served from the Cloudfront domain. This was inefficient
in multiple ways, both in building and storing client files on Docker image and pods just
to serve that one file, and because it requires maintaining multiple domains for client files.

Client build will now use an alternate Dockerfile that doesn't do anything besides push the
files to S3. build_and_publish_package.sh will not push this image to ECR since it will not
be served by Kubernetes.

All static files will come from the client subdomain, which should be attached to the Cloudfront
distribution instead of the `resources.<domain>` subdomain (a certificate for the client subdomain
is needed in us-east-1 in this case). The only exception is if STORAGE_S3_ENDPOINT is defined,
as this serves from the S3 URL directly instead of through Cloudfront.

If both SERVE_CLIENT_FROM_API and SERVE_CLIENT_FROM_STORAGE_PROVIDER are true, it will serve
from the storage provider, and will not build the client files into the API.

## References

closes #7990 

## Checklist

## QA Steps


scripts/update-cloudfront-function.ts
```diff
@@ -0,0 +1,55 @@
...
+import {InstalledRoutesInterface} from "@etherealengine/common/src/interfaces/Route";
+import {Application} from "@feathersjs/feathers";
+
+cli.enable('status')
+
+const PAGE_SIZE = 1
+
+
+const getAllRoutes = async (app: Application, routes: InstalledRoutesInterface[], skip: number) => {
+    const routePage = await app.service('route').find({
```
hanzlamateen, line 18: Instead of 'route' we can use `routePath` variable that is exported from route.scheme.ts

> We are trying to use single point of reference variable for service name, db name, etc.

hanzlamateen (approved):

PR #8103, opened 2023-06-20
Changes for location services to use schema, knex & feathers 5

### LINKED ISSUE #8284: Migrate location services to feathers 5

hanzlamateen:
## Summary

This PR Covers following services migration to feathers 5:
- location
- location-type
- location-setting

## References

#7645, closes #8284

## Checklist

## QA Steps


barankyle (changes requested):
I've noticed that selecting a location type when creating/editing a location in /admin/locations does not work ATM; no options are populating the selector, so it cannot be used. Since I see this on dev as well, it seems this PR didn't break it, but please fix it since it fits with everything else this PR is doing.

I also noticed that if a user goes to a location they are not authorized for, the Not Authorized notification is not shown. From the logs it looks like that action it getting dispatched, but it's not being picked up, or something. This can be tested by making a private location, then opening a new browser profile and trying to go to that location as a guest.

hanzlamateen:
> I've noticed that selecting a location type when creating/editing a location in /admin/locations does not work ATM; no options are populating the selector, so it cannot be used. Since I see this on dev as well, it seems this PR didn't break it, but please fix it since it fits with everything else this PR is doing.
> 
> I also noticed that if a user goes to a location they are not authorized for, the Not Authorized notification is not shown. From the logs it looks like that action it getting dispatched, but it's not being picked up, or something. This can be tested by making a private location, then opening a new browser profile and trying to go to that location as a guest.

I have fixed the location type not appearing in location drawer issue.

Though for Not Authorized notification, I am not sure why its not appearing. I looked into the code and its been called properly. It seems like `WarningUISystem` is having some problem internally. Its not showing up the ui despite the `WarningUIService.openWarning` being called for above scenario. Its just something in xrui and partically around call https://github.com/EtherealEngine/etherealengine/blob/9b477608c821efe42737f6b974f5125a845b24ae/packages/client-core/src/systems/WarningUISystem.tsx#L208 which is causing it not to appear.

barankyle (approved):

PR #8112, opened 2023-06-21
VariantComponent

### LINKED ISSUE #8109: LODComponent Refactor to VariationComponent
The LODComponent needs to be refactored to decouple it from the model component. to allow for variations of 

- [ ] Images (ImageComponent)
- [ ] Videos (VideoComponent)
- [ ] Arbitrary ECS Data (PrefabComponent)
- [ ] 3D Models (ModelComponent) + Instanced Models (InstancingComponent)
- [ ] Materials (MaterialComponent)
- [ ] Geometry (---???)

The component itself needs to have a generalized set of heuristics that are run to determine when to switch between different variations. Currently we have a Manual, Camera Distance, and Device heuristics which should be implemented for all applicable resources

### LINKED ISSUE #8116: Implement Variants for 3D Models (ModelComponent)

dinomut1:
## Summary

Abstract LODComponent into more generalized version which handles changing src values on different assets in the scene

## References

closes #8116


packages/engine/src/scene/systems/VariantSystem.tsx
```diff
@@ -31,28 +31,25 @@ import { getState } from '@etherealengine/hyperflux'
...
 let lastUpdate = 0
 
+const modelVariantQuery = defineQuery([VariantComponent, ModelComponent, TransformComponent])
+
 function execute() {
   const engineState = getState(EngineState)
   if (engineState.elapsedSeconds - lastUpdate < updateFrequency) return
   lastUpdate = engineState.elapsedSeconds
+
+  modelVariantQuery().forEach(setModelVariant)
```
HexaField, line 49: `for of` is fastest, and is what is standard amongst the code base

packages/engine/src/scene/components/VariantComponent.tsx
```diff
@@ -107,6 +111,18 @@ const VariantLevelReactor = React.memo(({ entity, level }: { level: number; enti
...
 
+  useEffect(() => {
+    //if the variant heuristic is set to Distance, add the DistanceFromCameraComponent
+    if (variantComponent.heuristic.value === 'DISTANCE') {
+      setComponent(entity, DistanceFromCameraComponent)
+      variantLevel.metadata['minDistance'].value === undefined && variantLevel.metadata['minDistance'].set(0)
+      variantLevel.metadata['maxDistance'].value === undefined && variantLevel.metadata['maxDistance'].set(0)
+    } else {
+      //otherwise, remove the DistanceFromCameraComponent
+      hasComponent(entity, DistanceFromCameraComponent) && removeComponent(entity, DistanceFromCameraComponent)
```
HexaField, line 122: why are we removing the DistanceFromCameraComponent when not in distance mode? it's needed for frustum culling to work properly in SceneObjectSystem

HexaField (changes requested):

packages/engine/src/scene/components/VariantComponent.tsx
dinomut1: Currently DistanceFromCameraComponent is only ever assigned to avatars on import. We're not using it for any Model assets, and probably shouldn't if it can be avoided

PR #8211, opened 2023-06-30
added mobile jump

SYBIOTE:
## Summary

added on avatar tap jump for touch enabled devices

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/engine/src/avatar/AvatarInputSystem.ts
```diff
@@ -275,7 +273,7 @@ const execute = () => {
         if (primaryClickCount > 0) {
           primaryClickTimer += getState(EngineState).deltaSeconds
           if (primaryClickTimer <= primaryClickTimeout) {
-            if (buttons.PrimaryClick?.up) {
+            if (buttons.PrimaryClick?.up && isAvatarClicked()) {
```
HexaField, line 276: let's store the result of `buttons.PrimaryClick?.up && isAvatarClicked()` so we don't have to recompute it a few lines later

HexaField (changes requested):

packages/engine/src/avatar/AvatarInputSystem.ts
SYBIOTE: please wait, i am testing it

packages/engine/src/avatar/AvatarInputSystem.ts
```diff
@@ -270,10 +270,12 @@ const execute = () => {
           reDoubleClickTimer = 0
         }
       } else {
+        const avatarClicked = isAvatarClicked()
```
HexaField, line 273: we should compute this as `const avatarClicked = buttons.PrimaryClick?.up && isAvatarClicked()` to avoid running the physics raycast when `buttons.PrimaryClick?.up` is not true

HexaField (changes requested):

packages/engine/src/avatar/AvatarInputSystem.ts
SYBIOTE: its done, and i optimized it quite a bit as well

packages/engine/src/avatar/AvatarInputSystem.ts
SYBIOTE: Yup it's done

packages/engine/src/avatar/AvatarInputSystem.ts
```diff
@@ -144,11 +160,62 @@ const onKeyP = () => {
...
+  const hits = Physics.castRayFromCamera(
+    Engine.instance.camera,
+    Engine.instance.pointerState.position,
+    Engine.instance.physicsWorld,
+    raycastComponentData
+  )
+  if (hits.length) {
+    const hit = hits[0]
+    const hitEntity = (hit.body?.userData as any)?.entity as Entity
+    console.log(hitEntity, Engine.instance.localClientEntity)
```
HexaField, line 173: Remove log

HexaField (changes requested):
Looking good

packages/engine/src/avatar/AvatarInputSystem.ts
SYBIOTE: missed this,  fixed it

packages/engine/src/avatar/AvatarInputSystem.ts
```diff
@@ -144,11 +160,61 @@ const onKeyP = () => {
...
+let primaryClickCount = 0
+const primaryClickTimeout = 2
+let primaryClickTimer = 0
+const reDoubleClickTimeout = 0.2
+let reDoubleClickTimer = 0
+const handleTouchJump = (buttons): boolean => {
+  if (!isTouchAvailable) return false
+  if (buttons.PrimaryClick?.up) {
+    //for single frame raycast
+    if (!isAvatarClicked()) return false
```
HexaField, line 189: if the avatar is not clicked, we should reset all timers to 0

packages/engine/src/avatar/AvatarInputSystem.ts
```diff
@@ -144,11 +160,61 @@ const onKeyP = () => {
...
+    }
+  }
+  return false
+}
+
+let primaryClickCount = 0
+const primaryClickTimeout = 2
+let primaryClickTimer = 0
+const reDoubleClickTimeout = 0.2
+let reDoubleClickTimer = 0
```
HexaField, line 184: we should name this `doubleClickTimer` and `doubleClickTimeout`

HexaField (changes requested):
only a small problem i'm seeing. it seems to be sticky with double clicks, often jumping when i expect it not to (after holding down for a while or not tapping for a while) which i think is just not resetting the timers under certain conditions.

SYBIOTE:
i understood the issue the timeout is 
`const doubleClickTimeout = 2`
its too big
its waiting 2 seconds before timing out between click of course it feels sticky, ill shorten it to .6 seconds

packages/engine/src/avatar/AvatarInputSystem.ts
SYBIOTE: yup renamed the variables as well

SYBIOTE: i think we should make a function in client system , which allows us to detect double tap for any  button in the system

SYBIOTE: we would just have to paste this function

SYBIOTE:
merged with latest dev , i think we can merge this in now , no more conflicts

HexaField (approved):

PR #8225, opened 2023-07-05
Updated client to prefer lowest-res video when video is small.

barankyle:
## Summary

Rendering 1080p video for every user with their camera on can be taxing on a user's
device when several people have their cameras on. Additionally, it's very hard to
tell the difference in quality when the video is so small. Made consumers prefer the
lowest spatial layer (resolution) unless PiP is on, at which point it switches to
the highest spatial layer, or if immersive media is enabled, which will use either
the top layer-1 if there are more than 2, or 0 if there are not.

Updated user window order so that the user's own screen/cam come at the top.

Replaced tcp-port-detect with detect-port, as the former was not detecting certain
processes that were actively running.

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/client-core/src/components/UserMediaWindow/index.tsx
```diff
@@ -434,6 +435,13 @@ export const UserMediaWindow = ({ peerID, type }: Props): JSX.Element => {
     document.getElementById(peerID + '-' + type + '-audio-container')!.append(audioElement)
   }, [])
 
+  useEffect(() => {
+    if (!videoStream) return
+    const mediaNetwork = Engine.instance.mediaNetwork as SocketWebRTCClientNetwork
+    if (isPiP) setPreferredConsumerLayer(mediaNetwork, videoStream as ConsumerExtension, 3)
```
HexaField, line 441: Are these layer numbers belonging to an enum? If so, we should use them directly, and if not, we should define one rather than using magic numbers.

HexaField (changes requested):

packages/client-core/src/components/UserMediaWindow/index.tsx
barankyle: I changed things some after seeing this comment. I was actually not using the lowest layer available; the layers are 0-indexed, and the defaults have been changed to reflect this. Instead of hard-coding max to '3', it's now checking how many layers there are from scalabilityMode, and then making some calculations from there.

HexaField (approved):

PR #8247, opened 2023-07-07
Changes for authentication service to use schema, knex & feathers 5

hurairahmateen:
## Summary

## References

#7645 

## Checklist

## QA Steps


packages/server-core/src/setting/authentication-setting/authentication-setting.class.ts
```diff
@@ -98,27 +99,21 @@ export class Authentication<T = AdminAuthSettingDataType> extends Service<T> {
     }
   }
 
-  async patch(id: string, data: any, params?: Params): Promise<T[] | T> {
-    const authSettings = await this.app.service('authentication-setting').get(id)
-    let existingCallback = JSON.parse(authSettings.callback as any)
-    if (typeof existingCallback === 'string') existingCallback = JSON.parse(existingCallback)
+  async patch(id: Id, data: AuthenticationSettingPatch, params?: AuthenticationSettingParams) {
```
barankyle, line 102: Patching auth settings isn't working for me, and it seems to be because of the need to JSON parse strings and then turn them back into strings before saving. Please test this from /admin/settings#authentication.

barankyle (changes requested):

packages/server-core/src/setting/authentication-setting/authentication-setting.class.ts
barankyle: data.oauth and data.callback are coming into this handler as strings, but they need to be parsed into objects before manipulation of them can be done properly. They might need to be stringified before super._patch is called.

packages/server-core/src/setting/authentication-setting/authentication-setting.class.ts
```diff
@@ -23,71 +23,72 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
-  constructor(options: Partial<SequelizeServiceOptions>, app: Application) {
+  constructor(options: KnexAdapterOptions, app: Application) {
     super(options)
     this.app = app
   }
 
-  async find(params?: UserParams): Promise<T[] | Paginated<T>> {
-    const auth = (await super.find()) as any
-    const loggedInUser = params!.user as UserInterface
+  async find(params?: AuthenticationSettingParams) {
```
barankyle, line 70: I'm encountering issues with authentication-setting.find with non-admin users. It's returning `[{"message":"Cannot convert undefined or null to object"}]`. It might be because calls with the admin scope are returning data that calls without it do not. Could it be that certain fields are defined as required that shouldn't be?

packages/server-core/src/setting/authentication-setting/authentication-setting.class.ts
hurairahmateen: I pushed a fixed. Kindly review.

packages/server-core/src/setting/authentication-setting/authentication-setting.class.ts
hurairahmateen: I pushed a fixed. Kindly review.

packages/server-core/src/setting/authentication-setting/authentication-setting.class.ts
barankyle: My local copy, and I believe our deployed copies as well, still have callback saved in the database as a double-stringified string, i.e. it seems like it's been JSON.stringified twice. Calling JSON.parse() just once still has it as a string, and so 'data.callback[key] = `${config.client.url}/auth/oauth/${key}`' is throwing errors about trying to assign values to a string. Please add a check after line 120 and call JSON.parse again if it's still a string. 

I don't think any existing copies of this in any databases would need a third call of JSON.parse.

hurairahmateen: Fixed it. Should be working now.

barankyle (approved):

PR #8283, opened 2023-07-14
Event Sourcing for Grabbables

HexaField:
## Summary

- [x] Fixes netcode for authority transfer
- [x] Implements Event Sourcing for Grabbables
- [x] Renames equippables to grabbables
- [x] Fixes testing configuration problems

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/engine/src/interaction/components/GrabbableComponent.ts
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * GrabberComponent
+ * - Allows an entity to grab a GrabbableComponent
+ */
+export const GrabberComponent = defineComponent({
+  name: 'GrabberComponent',
+
+  onInit(entity) {
+    return {
+      grabbedEntity: null as Entity | null
```
speigg, line 93: Does this mean only one object can be grabbed at a time? Why limit it in this way?

speigg (changes requested):

speigg:
When a new client joins, the grabbable does not appear in the current position.

HexaField:
> When a new client joins, the grabbable does not appear in the current position.

you'll need to reinit your db, this is due to it spawning as a fixed rigidbody and not dynamic. the problem with this then though, is it falls through the floor because the skybase loads after it

packages/engine/src/interaction/components/GrabbableComponent.ts
HexaField: for now, yeah. this PR is just refactoring this functionality, not extending it

PR #8403, opened 2023-07-31
Add More Compression Options to Model Transform

dinomut1:
Add "compression level", and "max codebooks" option for etc1 and "UASTC level" for uastc ktx2 compression.


CITIZENDOT:
I think Compression Level only applies to `ETC1S` files: [basis_wrappers.cpp#L1791](https://github.com/BinomialLLC/basis_universal/blob/master/webgl/transcoder/basis_wrappers.cpp#L1791)

dinomut1:
It does, we only allow configuration for comp_level when ETC1S is selected

PR #8420, opened 2023-08-03
new: introduce actions panel in debug menu

### LINKED ISSUE #8406: Debug shows actions
The debug menu should have an infinite scroll with search on the engine's local action cache

aditya-mitra:
## Summary

Introduces History and Cached Actions in Debug menu

Screenshot_20230803_102322

## References

closes #8406 

## Checklist

## QA Steps

### DIFF
--- a/packages/client-core/src/components/Debug/index.tsx
+++ b/packages/client-core/src/components/Debug/index.tsx
@@ -43,13 +43,13 @@ import { RootSystemGroup, SimulationSystemGroup } from '@etherealengine/engine/s
 import { entityExists } from '@etherealengine/engine/src/ecs/functions/EntityFunctions'
 import { EntityTreeComponent } from '@etherealengine/engine/src/ecs/functions/EntityTree'
 import { System, SystemDefinitions, SystemUUID } from '@etherealengine/engine/src/ecs/functions/SystemFunctions'
-import { NetworkState } from '@etherealengine/engine/src/networking/NetworkState'
 import { RendererState } from '@etherealengine/engine/src/renderer/RendererState'
 import { NameComponent } from '@etherealengine/engine/src/scene/components/NameComponent'
 import { UUIDComponent } from '@etherealengine/engine/src/scene/components/UUIDComponent'
-import { getMutableState, useHookstate } from '@etherealengine/hyperflux'
+import { NO_PROXY, getMutableState, useHookstate } from '@etherealengine/hyperflux'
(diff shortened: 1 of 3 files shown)


packages/client-core/src/components/Debug/index.tsx
```diff
@@ -105,18 +105,18 @@ const convertSystemTypeToDesiredType = (system: System): DesiredType => {
   return desired
 }
 
-export const Debug = ({ showingStateRef }) => {
-  useHookstate(getMutableState(EngineState).frameTime).value
+export const Debug = ({ showingStateRef }: { showingStateRef: React.MutableRefObject<boolean> }) => {
   const rendererState = useHookstate(getMutableState(RendererState))
   const engineState = useHookstate(getMutableState(EngineState))
+
+  engineState.frameTime.value // make Engine.instance data reactive in the render tree
```
HexaField, line 112: calling this does nothing and is unnecessary

HexaField (changes requested):

packages/client-core/src/components/Debug/index.tsx
aditya-mitra: This is necessary for re-rendering the JSONTree's when anything changes. For example, if a new entity is added to the scene, the `namedEntities` data might not show it.

HexaField:
Screenshot from 2023-08-11 09-14-06

As we can see here, the actions are concatenating together. We want this to be a time ordered log, where each action appears in the order it was sent. I think we should have a separate JSON tree for each action in the cached action / action history arrays, instead of one JSON tree for each array.

aditya-mitra:
image

We are now showing 3 lists: an ordered list showing the actions history as ordered (from last to first), actions history grouped by their types, and cached history grouped by their types.

HexaField:
> We are now showing 3 lists: an ordered list showing the actions history as ordered (from last to first), actions history grouped by their types, and cached history grouped by their types.

This is not what I asked.

We still only want 2 sections (cahced & history) but rather than 1 JSON tree for each category, we want a JSON tree for each action.

Actions are an implementation of an event log. This event log is in ordered by time naturally, and that must be reflected in the debug menu. The point of displaying this menu visually is to easily inspect the action history, thus the debug menu should give you the information you need in the fewest steps necessary.

What we need here, is a list of actions (with infinite scroll) in which the label of each action when hidden is the `type` field of the action, and the action can be opened into a JSON tree view.

aditya-mitra:
Requested changes implemented in #8509

PR #8436, opened 2023-08-05
Fixed paginate: false for routes

barankyle:
## Summary

If the client passes paginate: false on params, it will not make it to the API server's handler for security reasons. Changed the find call to accept paginate: false as a query param, and to move it to the top of the params object if it's false.

Had to make changes to route's typing to support this. Made some changes to RouterService to handle the fact that the route.find call now returns an array rather than a paginated response. Cast several calls to route.find as either paginated or unpaginated based on what they are requesting.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/server-core/src/route/route/route.class.ts
+++ b/packages/server-core/src/route/route/route.class.ts
@@ -23,18 +23,51 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
-import type { Params } from '@feathersjs/feathers'
-import type { KnexAdapterParams } from '@feathersjs/knex'
-import { KnexService } from '@feathersjs/knex'
+import type { NullableId, Params } from '@feathersjs/feathers'
+import type { KnexAdapterOptions, KnexAdapterParams } from '@feathersjs/knex'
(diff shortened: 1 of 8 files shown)


packages/server-core/src/route/route/route.class.ts
```diff
@@ -23,18 +23,51 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
+> {
+  app: Application
+
+  constructor(options: KnexAdapterOptions, app: Application) {
+    super(options)
+    this.app = app
+  }
+
+  async find(params?: RouteParams) {
+    if (params?.query?.paginate === false) {
```
HexaField, line 51: Should this be `typeof params?.query?.paginate !== undefined)` to catch if it's true or something else and delete it?

HexaField (changes requested):

packages/server-core/src/route/route/route.class.ts
barankyle: I made it always remove the query param. There's still a check if it's false before setting params.paginate to it since params.paginate cannot be true, or anything other than false or some sort of explicit pagination param that I doubt we'll ever pass.

HexaField (approved):

PR #8439, opened 2023-08-07
feat(TransformProperty): local transform component mode

aditya-mitra:
## Summary

Added a boolean input to use `LocalTranformComponent` when the input is checked.

image

## References

closes https://github.com/EtherealEngine/etherealengine/issues/8432

## Checklist

## QA Steps


HexaField (changes requested):
this works fine, but we cannot update transform properties on the TransformComponent the same way we do with LocalTransformComponent (via the gizmo), we need to compute and update the TransformComponent directly, or it has incorrect transformations applied and sometimes does not work, snapping back to the original position. We still need to update the gizmo accordingly

HexaField (changes requested):
Using the vector3 scrubbers in the properties panel when using non local transform resets all the values to 0

speigg:
IMO we should always be editing the local transform by default, and have the option to view the world transform (by modifying the local transform behind the scenes).

HexaField (changes requested):
this needs to be rewritten to work with our new snapshots concept

dinomut1 (approved):
Looks like this works as intended after resolving merge conflicts

PR #8470, opened 2023-08-12
Fixed a few issues with admin avatar drawer.

### LINKED ISSUE #8454: [Bug]: Adding avatars from /admin/avatars does not return model/thumbnail, thumbnail is never shown in create drawer
I encountered a couple of issues with avatar creation/editing from /admin/avatars.

First, avatar.create is not returning the thumbnail and model resources. When the new avatar is rendered in the list, it has no thumbnail, and if you go to edit it, the URLs for those resources are not present. This makes it look like something went wrong with creation. If you refresh the page, avatar.find is returning the thumbnail and model resources.

Second, the thumbnail is not displayed when the drawer is in create mode, so you can't see what the uploaded or generated thumbnail looks like. It is shown in Edit mode.

barankyle:
## Summary

Newly-created avatars did not have thumbnails/models show up until a page refresh or another avatar was created. This was due to the switch to FeathersHooks; the avatar is created but did not have the model or thumbnail yet, then it uploads the model and thumbnail, but useFind() was only being triggered by the creation, not the resources getting added. Added a patched handler on avatar.ts that will notify people with that avatar and the user who made the request. Passing params to avatar.patch in uploadAvatarStaticResource to make this happen.

Thumbnail preview was not showing in AvatarDrawer in create mode because it was set to only show in edit mode. Removed this restriction, as it is helpful to show the thumbnail during creation too.

## References

closes #8454 

## Checklist

## QA Steps

### DIFF
--- a/packages/server-core/src/user/avatar/avatar.ts
+++ b/packages/server-core/src/user/avatar/avatar.ts
@@ -23,9 +23,12 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
-import { avatarMethods, avatarPath } from '@etherealengine/engine/src/schemas/user/avatar.schema'
+import { avatarMethods, avatarPath, AvatarType } from '@etherealengine/engine/src/schemas/user/avatar.schema'
 
+import { UserID, userPath, UserType } from '@etherealengine/engine/src/schemas/user/user.schema'
+import { Paginated } from '@feathersjs/feathers'
 import { Application } from '../../../declarations'
+import logger from '../../ServerLogger'
 import { AvatarService } from './avatar.class'
 import avatarDocs from './avatar.docs'
 import hooks from './avatar.hooks'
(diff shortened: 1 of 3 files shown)


hanzlamateen (approved):

packages/server-core/src/user/avatar/avatar.ts
```diff
@@ -54,4 +57,23 @@ export default (app: Application): void => {
 
   const service = app.service(avatarPath)
   service.hooks(hooks)
+
+  service.publish('patched', async (data: AvatarType, context) => {
+    try {
+      const { params } = context
+      let targetIds = [params.user?.id]
+      const usersWithAvatar = (
+        (await app.service(userPath).find({
```
hanzlamateen, line 66: We can use `_find` here inorder to avoid running resolvers as all we need is user ids from this API call.

hanzlamateen (changes requested):

packages/server-core/src/user/avatar/avatar.ts
barankyle: Done

hanzlamateen (approved):

PR #8483, opened 2023-08-17
Changes for channel-user service to use schema, knex & feathers 5

### LINKED ISSUE #8481: Migrate channel_user service to feathers 5

MoizAdnan:
## Summary

## References

#7645, closes #8481 

## Checklist

## QA Steps


packages/server-core/src/social/channel-user/migrations/20230817133703_channel-user.ts
```diff
@@ -0,0 +1,86 @@
...
+  if (oldNamedTableExists) {
+    // In case sequelize creates the new table before we migrate the old table
+    if (tableExists) await trx.schema.dropTable(channelUserPath)
+    await trx.schema.renameTable(oldTableName, channelUserPath)
+  }
+
+  tableExists = await trx.schema.hasTable(channelUserPath)
+
+  if (!tableExists && !oldNamedTableExists) {
+    await knex.schema.createTable(channelUserPath, (table) => {
```
hanzlamateen, line 53: It should us `trx` here instead of `knex`

hanzlamateen (changes requested):

MoizAdnan:
@barankyle please approve this PR

packages/server-core/src/social/channel-user/channel-user.resolvers.ts
```diff
@@ -0,0 +1,56 @@
...
+import { v4 } from 'uuid'
+
+import { ChannelUserQuery, ChannelUserType } from '@etherealengine/engine/src/schemas/social/channel-user.schema'
+import { userPath } from '@etherealengine/engine/src/schemas/user/user.schema'
+import type { HookContext } from '@etherealengine/server-core/declarations'
+import { getDateTimeSql } from '../../util/get-datetime-sql'
+
+export const channelUserResolver = resolve<ChannelUserType, HookContext>({
+  user: virtual(async (channelUser, context) => {
+    const user = await context.app.service(userPath)._get(channelUser.userId)
```
hanzlamateen, line 37: There should be a null check here for `channelUser.userId` because migration file lists userId as defaults to null

hanzlamateen (changes requested):

barankyle (changes requested):
I'm encountering errors when loading an existing channel and its messages. It seems that channel_users is not being populated on the channel when being returned from channel.find. The error technically isn't there, but a few front-end components like Message.tsx and FriendsMenu.tsx try to use a value from one of the channel users and fails because it's not there.

You can test this by opening two different Chrome profiles, going to the same location, then making those two users friends. Try to open the chat for both of them, it looks like the one who immediately accepted the friend request doesn't have this problem, but the one who sent the friend request does, and if either of them reload, they definitely can't open it because of this issue.

barankyle (changes requested):
One minor thing I'm seeing now. When a chat between friends is open, the chat window is not immediately updating with new messages that others have sent. If I have two users as friends and they're sending messages to each other, the act of sending a message causes a re-fetch of the message history, but if the other person is sending messages, those new messages don't show up until the first user sends a message. The other person's messages should be filling in automatically as they're sent.

I do not see this issue on dev branch, so it seems this is a problem in this branch.

barankyle (approved):

hanzlamateen (approved):

PR #8495, opened 2023-08-18
Allow Behave Graph "play" field to be serialized to True

dinomut1:
Allows designating Behave Graphs to auto-execute on scene load. We prevent execution in the editor with a EngineState check in the reactor rather than always serializing "play" to false.


SYBIOTE:
this is good, lets merge it in

SYBIOTE (approved):

PR #8541, opened 2023-08-21
Animation Speed Reactivity

dinomut1:


HexaField (changes requested):
We should move `animationSpeed` to the `LoopAnimationComponent` since avatars do not utilize this property and the loop animation is saved, not the Animation Component

PR #8577, opened 2023-08-23
Updated uploadToRepo to use git LFS

barankyle:
## Summary

Existing code did not use git LFS, which made it unusable with projects that had files too big to upload directly to GitHub, as the blob uploads would be rejected. createBlobForFile now checks the size of the file and, if it's over 30 MB, which is low enough to not come close to GitHub's stated 50 MB cutoff, it instead uploads the file to GitHub's LFS and makes the file a pointer to it instead (which is standard LFS practice). The flow also adds the files to the .gitattributes file, creating and adding it to the git history if that doesn't already exist.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/server-core/src/media/recursive-archiver/archiver.class.ts
+++ b/packages/server-core/src/media/recursive-archiver/archiver.class.ts
@@ -24,21 +24,91 @@ Ethereal Engine. All Rights Reserved.
 */
 
 import { ServiceMethods } from '@feathersjs/feathers/lib/declarations'
-import appRootPath from 'app-root-path'
 import JSZip from 'jszip'
 import fetch from 'node-fetch'
-import path from 'path/posix'
 
+import { BadRequest } from '@feathersjs/errors'
 import { Application } from '../../../declarations'
 import logger from '../../ServerLogger'
 import { UserParams } from '../../api/root-params'
+import config from '../../appconfig'
+import { createExecutorJob, getDirectoryArchiveJobBody } from '../../projects/project/project-helper'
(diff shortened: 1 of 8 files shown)


packages/server-core/src/media/recursive-archiver/archiver.class.ts
```diff
@@ -24,87 +24,112 @@ Ethereal Engine. All Rights Reserved.
...
 
-    return zipOutputDirectory
+  constructor(app: Application) {
+    this.app = app
+  }
+
+  async setup(app: Application, path: string) {}
+
+  async get(directory: string, params?: ArchiveParams): Promise<string> {
+    if (!config.kubernetes.enabled || params?.isJob) return archive(directory, params)
```
hanzlamateen, line 117: Having `!config.kubernetes.enabled` means for local development we wouldnt be using Git LFS. Is it because of assumption that in local environment file will be uploaded to local file server?

I am just wandering if this will impact once we move to MinIO?

hanzlamateen (approved):

packages/server-core/src/media/recursive-archiver/archiver.class.ts
barankyle: Unless I'm missing something, Git LFS is being used in all situations. I converted archiver, project-github-push, and project.update to either make a job or run the function they originally did. Locally, we never want to make a job, so those functions will just call their original logic, which is now in a separate function. That separate function, or sub-functions of it, will use LFS. In a deployed environment, the default is to make an executor job for it, with the job calling the function with isJob: true, which will run the intended function.

PR #8595, opened 2023-08-28
fix scrubber

SYBIOTE:
## Summary

fix improper passing of props to scrubber

## References

closes #_insert number here_

## Checklist

## QA Steps


packages/editor/src/components/inputs/Scrubber.tsx
```diff
@@ -37,18 +37,20 @@ import { useHookstate } from '@etherealengine/hyperflux'
 import Portal from '../layout/Portal'
 
 type ScrubberContainerProps = {
-  tag?: keyof JSX.IntrinsicElements
+  tag?: any
```
HexaField, line 40: why did we remove the type here? can we specify a type that is actually correct here instead of removing the typing

packages/editor/src/components/inputs/Scrubber.tsx
```diff
@@ -153,9 +156,9 @@ const Scrubber = ({
     state.delta.set(0)
     state.mouseX.set(event.clientX)
     state.mouseY.set(event.clientY)
-
+    console.log('DEBUG', scrubberEl, scrubberEl?.current)
```
HexaField, line 159: remove console logs

HexaField (changes requested):

packages/editor/src/components/inputs/Scrubber.tsx
SYBIOTE: yup

packages/editor/src/components/inputs/Scrubber.tsx
SYBIOTE: i was getting two typescript errors otherwise 
Types of property 'ref' are incompatible. 
Type 'Ref<HTMLElement>' is not assignable to type 'LegacyRef<SVGSymbolElement> | undefined'
and
Expression produces a union type that is too complex to represent.ts(2590)

on line 48

HexaField (approved):

PR #8596, opened 2023-08-28
tests(invite service): add tests for empty string search

### LINKED ISSUE #8491: [Bug] Invite page crashed admin panel

aditya-mitra:
## References

closes #8491 

## Checklist

## QA Steps


packages/server-core/src/social/invite/invite.class.ts
```diff
@@ -199,6 +199,8 @@ export const inviteAll = async (inviteService: Invite, query: Query, user: UserT
   }
   if (!query.existenceCheck) delete query.userId
   delete query.existenceCheck
+  delete query.search
```
hanzlamateen, line 202: Maybe its good to have a check here like

```
if (query.search) {
  delete query.search
}
```

hanzlamateen (changes requested):

packages/server-core/src/social/invite/invite.class.ts
HexaField: yes

aditya-mitra: Actually it was the reason for the bug
image

hanzlamateen: Yes, it can be removed. But just put that delete in an if condition. I think as you merged the PR with latest dev that change is no longer required?

aditya-mitra: Yes, you are right
That code is now present in latest dev
Maybe we could add the tests in this PR only

aditya-mitra: Yes, you are right
That code is now present in latest dev
Maybe we could add the tests in this PR only

hanzlamateen (approved):

Issue #8361, opened 2023-07-24
Add Link Component


dinomut1:
Use InputComponent to create a simple interactable that will open a hyperlink in a new tab

PR #8622, opened 2023-08-29
Hyper Link Component

DanielBelmes:
## Summary

Added a link component that can be added to entities that can take inputs. When hovering over the parent entitiy it will indicate it is clickable and highlight. When clicked it will open a configured hyperlink in a new tab.

## References

#8361 

## Checklist

## QA Steps

Throw in a model and add the link to its properties. You can click within studio.


packages/engine/src/scene/components/LinkComponent.ts
```diff
@@ -0,0 +1,79 @@
...
+    return {
+      url: 'https://www.etherealengine.org',
+    }
+  },
+
+  onSet: (entity, component, json) => {
+    console.log("ONSET")
+    console.log(json)
+    if (!json) return
+    try {
```
HexaField, line 47: is the try/catch for when the url passed into `new URL` is invalid? If so, we should be handling these cases in the reactor, and adding an entity error (see MediaComponent for reference) if the URL is invalid. `onSet` should set the data if the data type is valid, not specifically the data content.

packages/engine/src/scene/systems/LinkSystem.ts
```diff
@@ -0,0 +1,57 @@
...
+    const inputComponent = getComponent(entity, InputComponent)
+    const inputSourceEntity = inputComponent?.inputSources[0]
+
+    if (inputSourceEntity) {
+      const inputSource = getOptionalComponent(inputSourceEntity, InputSourceComponent)
+      const buttons = inputSource?.buttons
+
+      if (buttons?.PrimaryClick?.touched) {
+        if(buttons.PrimaryClick.up) {
+          window.open(linkComponent.url,"_blank")
```
HexaField, line 47: We should ensure `window` is defined in case this system accidentally is run on the server

packages/editor/src/components/properties/LinkNodeEditor.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const linkComponent = useComponent(props.entity, LinkComponent)
+
+  return (
+    <NodeEditor
+      {...props}
+      name={t('editor:properties.link.name')}
+      description={t('editor:properties.link.description')}
+    >
+      <InputGroup name="Link Url" label={t('editor:properties.image.lbl-imgURL')}>
+        <ImageInput value={linkComponent.url.value ?? ''} onChange={updateProperty(LinkComponent, 'url')} />
```
HexaField, line 56: `linkComponent.url.value` should always be defined and of type string, so ` ?? ''` is unnecessary

packages/engine/src/scene/components/LinkComponent.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  onInit: (entity) => {
+    setComponent(entity, InputComponent)
+    return {
+      url: 'https://www.etherealengine.org',
+    }
+  },
+
+  onSet: (entity, component, json) => {
+    console.log("ONSET")
```
HexaField, line 44: cleanup logs

packages/engine/src/scene/components/LinkComponent.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.log(json)
+    if (!json) return
+    try {
+      if(typeof json?.url === 'string') {
```
HexaField, line 48: `json?.url` conditional existence check is unnecessary as we already escape `if (!json) return` earlier

packages/engine/src/scene/components/LinkComponent.ts
```diff
@@ -0,0 +1,79 @@
...
+      url: component.url.value
+    }
+  },
+
+  reactor: function () {
+    const entity = useEntityContext()
+    const input = useComponent(entity, InputComponent)
+
+    useEffect(() => {
+      const canvas = document.getElementById('engine-renderer-canvas')!
```
HexaField, line 68: we can use `EngineRenderer.instance.renderer.domElement` instead of document.getElementById to avoid relying on a magic string here

HexaField (changes requested):
Looks good, just some minor changes required. Please ensure all files are formatted correctly too.

DanielBelmes:
So currently this is working but the UX very weak. Stemming from this [bug](https://github.com/EtherealEngine/etherealengine/issues/8650)  

Since it's a bug on 3.js side. I need to find a way to add some User Experience to hovering over interactables in WebXR.

packages/engine/src/renderer/components/HighlightComponent.ts
```diff
@@ -39,29 +40,52 @@ export const HighlightComponent = defineComponent({
...
-const addToSelection = (obj: Object3D) => {
+const animateScale = (obj: Mesh, v: Vector3, xrSession) => {
+  let start = 0.0
+  function animate(timeStamp) {
+    if (start === 0.0) start = timeStamp
+    const elapsed = timeStamp - start
+    obj.scale.lerp(v, MathUtils.damp(0, 1, 0.01, elapsed))
+
+    if (obj.scale.x !== v.x) {
+      !xrSession ? requestAnimationFrame(animate) : xrSession?.requestAnimationFrame(animate)
```
HexaField, line 75: We don't want to be creating new animation frame requests, this should be handled in a system.

HexaField (changes requested):

packages/engine/src/renderer/components/HighlightComponent.ts
DanielBelmes: for the sake of time. I've reverted the changes in case this is needed. Will still be working on the animation for UX reasons.

DanielBelmes: Alright resolved with the TweenComponent setup. Was stuck for too long tweening between 1-0 for lerp 😢

packages/engine/src/scene/components/LinkComponent.ts
```diff
@@ -37,24 +37,31 @@ export const LinkComponent = defineComponent({
   jsonID: 'link',
 
   onInit: (entity) => {
-    setComponent(entity, InputComponent)
+    try {
+      setComponent(entity, InputComponent) // TODO implement schemas for componenets to use in behave graph
```
HexaField, line 41: let's move this to a reactor useEffect with no deps, initializers should be kept to simply specifying the schema

packages/engine/src/scene/systems/LinkSystem.ts
```diff
@@ -23,6 +23,7 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
+import { RouterService } from '@etherealengine/client-core/src/common/services/RouterService'
```
HexaField: the engine package cannot import client-core, the engine runs isomorphic across the client and server. if this requires manipulating client state, we should move the system to client-core (see how AvatarUISystem is started)

HexaField (changes requested):

packages/engine/src/scene/systems/LinkSystem.ts
```diff
@@ -0,0 +1,74 @@
...
+    const inputSourceEntity = inputComponent?.inputSources[0]
+
+    if (inputSourceEntity) {
+      const inputSource = getOptionalComponent(inputSourceEntity, InputSourceComponent)
+      const buttons = inputSource?.buttons
+
+      if (buttons)
+        if (buttons.PrimaryClick?.touched) {
+          if (buttons.PrimaryClick.up) {
+            if (!linkComponent.sceneNav) {
```
HexaField, line 52: we have duplicated code here for both conditions, we should either combine these clauses or create a function to unify.

HexaField (changes requested):

packages/engine/src/scene/systems/LinkSystem.ts
DanielBelmes: yup sounds good.

speigg (approved):

PR #8664, opened 2023-09-06
fix: make scene `save as` visible

### LINKED ISSUE #8656: [BUG] "save as" is broken in the Studio
Save As UI is broken.  No button to submit and proportions don't match the input fields 

Image

aditya-mitra:
## Summary

- save scene as png
- make the component fields visible
- change pixel ratio before resizing window

image

## References

closes #8656 

## Checklist

## QA Steps


packages/editor/src/functions/takeScreenshot.ts
```diff
@@ -200,29 +200,29 @@ export async function takeScreenshot(
...
+    EngineRenderer.instance.effectComposer.setSize(width, height, false)
+    EngineRenderer.instance.renderer.setPixelRatio(1)
+
     const interval = setInterval(() => {
       const viewport = EngineRenderer.instance.renderContext.getParameter(
         EngineRenderer.instance.renderContext.VIEWPORT
       )
-      // todo - scrolling in and out sometimes causes weird pixel ratios that can cause this to fail
-      if (viewport[2] === Math.round(width * pixelRatio) && viewport[3] === Math.round(height * pixelRatio)) {
+      if (viewport[2] === Math.round(width) && viewport[3] === Math.round(height)) {
```
HexaField, line 212: Pixel ratio must be preserved here

packages/editor/src/functions/takeScreenshot.ts
```diff
@@ -200,29 +200,29 @@ export async function takeScreenshot(
...
     }, 10)
+
     const timeout = setTimeout(() => {
       console.warn('Could not resize viewport in time')
       clearTimeout(timeout)
       clearInterval(interval)
       reject()
     }, 10000)
-
-    // set up effect composer
```
HexaField, line 222: Why was this moved earlier up?

HexaField (changes requested):

packages/editor/src/functions/takeScreenshot.ts
aditya-mitra: After testing this a lot of times, the timer functions are called before the resizing of the scene. So, the `if` condition above does not ever get satisfied and an error is thrown.
Synchronously calling them does not cause that error.

packages/editor/src/functions/takeScreenshot.ts
aditya-mitra: We are changing the pixel ratio to 1.
Also tried this a few times, `viewport` comes out to be like `0 0 512 320` and `height` and `width` are also `320` and `512` respectively but `pixelRatio` is `2`. So, the condition is not fulfilled. Also,`pixelRatio` is defined outside the promise so it does not change even though the promise inside changes the pixelRatio to 1.

HexaField (approved):

PR #8692, opened 2023-09-07
Fade to Video

dinomut1:
Adds ability to fade to a video instead of just to black


packages/engine/src/camera/systems/CameraFadeBlackEffectSystem.ts
```diff
@@ -57,9 +80,15 @@ const CameraFadeBlackEffectSystemState = defineState({
       depthTest: false,
       uniforms: {
         color: { value: new Color('black') },
-        intensity: { value: 0 }
+        intensity: { value: 0 },
+        resolution: { value: new Vector2(window.outerWidth, window.outerHeight) }
       }
     })
+
+    window.addEventListener('resize', function () {
```
HexaField, line 88: let's move this to a reactor so we can clean it up

HexaField (changes requested):

PR #8707, opened 2023-09-08
Prevent asset load errors from blocking scene loading

speigg:
When an asset fails to load, this PR allows the scene to continue loading in spite of the missing asset. An error is printed in the console.


HexaField (approved):
nice!

PR #8733, opened 2023-09-12
Skinned mesh ECS bounding box frustum culling

AidanCaruso:
## Summary

Adds bounding box based frustum culling for skinned meshes based on the frustum culling and bounding box components.

## References

closes #_insert number here_

## Checklist

## QA Steps

### DIFF
--- a/packages/engine/src/transform/systems/TransformSystem.ts
+++ b/packages/engine/src/transform/systems/TransformSystem.ts
@@ -402,13 +402,13 @@ const execute = () => {
   _projScreenMatrix.multiplyMatrices(camera.projectionMatrix, camera.matrixWorldInverse)
   _frustum.setFromProjectionMatrix(_projScreenMatrix)
 
-  for (const entity of frustumCulledQuery())
-    FrustumCullCameraComponent.isCulled[entity] = _frustum.containsPoint(
-      getComponent(entity, TransformComponent).position
-    )
-      ? 0
-      : 1
-
+  for (const entity of frustumCulledQuery()) {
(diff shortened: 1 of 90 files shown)


packages/engine/src/transform/systems/TransformSystem.ts
```diff
@@ -402,13 +402,13 @@ const execute = () => {
...
   _frustum.setFromProjectionMatrix(_projScreenMatrix)
 
-  for (const entity of frustumCulledQuery())
-    FrustumCullCameraComponent.isCulled[entity] = _frustum.containsPoint(
-      getComponent(entity, TransformComponent).position
-    )
-      ? 0
-      : 1
-
+  for (const entity of frustumCulledQuery()) {
```
HexaField, line 405: we might want to benchmark if having two queries here, one with bbox and one without gives better results, since we have a few extra conditions when they are combined

HexaField (changes requested):

packages/engine/src/transform/systems/TransformSystem.ts
AidanCaruso: Yes! I was thinking about this earlier, so I've been taking performance profiles for a while and I have seen little change, roughly ~3.5ms regardless of the extra bbox condition, and no discernable difference in fps.
image

AidanCaruso: image

Looks like it does tack on some ms to get the bounding box component and then perform tests. In this case it looks like an extra 3ms for 128 avatars. I'm not a fan.

HexaField (approved):

PR #8754, opened 2023-09-12
Animation race condition fixes

speigg:
## Summary
Expose AnimationAction properties in LoopAnimationComponent, and update BehaveGraph nodes


dinomut1 (changes requested):
Change xre.pause to xre.stop to avoid confusion on its effect

speigg:
> Change xre.pause to xre.stop to avoid confusion on its effect

Should do a proper pause now

dinomut1 (approved):

Issue #8771, opened 2023-09-13
[Bug]: Optimizations for Engine and nice features


DanielBelmes:
Chrome Dev tools mobile emulator doesn't trigger mobile code.
No onload event for software to listen too.
Even when CSM is disabled shadowmap steps are running from threejs
Foreach used in executeSystem which can slow us down slightly.
DropShadow tracking incorrectly

PR #8772, opened 2023-09-13
Optimize, Fix, and improve

DanielBelmes:
## Summary

- Added load event on window for e2e testing
- Fixed Drop Shadow raycasting
- Disabling shadows now disables shadowmap in threejs saving a couple draw calls
- Optimized system execute saving a couple ms

## References

#8771

## Checklist

### DIFF
--- a/packages/engine/src/scene/systems/ShadowSystem.tsx
+++ b/packages/engine/src/scene/systems/ShadowSystem.tsx
@@ -66,10 +66,11 @@ import { createEntity, removeEntity, useEntityContext } from '../../ecs/function
 import { createQueryReactor, defineSystem } from '../../ecs/functions/SystemFunctions'
 import { BoundingBoxComponent } from '../../interaction/components/BoundingBoxComponents'
 import { RendererState } from '../../renderer/RendererState'
-import { RenderSettingsState } from '../../renderer/WebGLRendererSystem'
+import { EngineRenderer, RenderSettingsState } from '../../renderer/WebGLRendererSystem'
 import { getShadowsEnabled, useShadowsEnabled } from '../../renderer/functions/RenderSettingsFunction'
 import { TransformComponent } from '../../transform/components/TransformComponent'
 import { XRLightProbeState } from '../../xr/XRLightProbeSystem'
+import { XRState } from '../../xr/XRState'
 import { DirectionalLightComponent } from '../components/DirectionalLightComponent'
 import { DropShadowComponent } from '../components/DropShadowComponent'
 import { GroupComponent, addObjectToGroup } from '../components/GroupComponent'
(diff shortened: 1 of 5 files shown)


packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -236,7 +237,7 @@ const DropShadowReactor = createQueryReactor([ShadowComponent], function DropSha
     if (sphere.radius > maxRadius) return
 
     const radius = Math.max(sphere.radius * 2, minRadius)
-    const center = groupComponent.value[0].worldToLocal(sphere.center)
+    const center = groupComponent.value[0].position
```
HexaField, line 240: this is a regression, as the point of using the bounding circle is for objects whose centers are significantly distant from the entity's transform. if the world/local space change fixes them appearing in the wrong location, then we can probably just use `sphere.center` directly here instead of `groupComponent.value[0].position` (equivalent to `TransformComponent position`)

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -302,6 +302,15 @@ const execute = () => {
 }
 
 const reactor = () => {
+  const rendererState = getMutableState(RendererState)
+  const useShadows = useHookstate(rendererState.useShadows)
+  const renderMode = useHookstate(rendererState.renderMode)
+  const isEditor = useHookstate(getMutableState(EngineState).isEditor)
+  const sessionState = useHookstate(getMutableState(XRState).sessionActive)
+  useEffect(() => {
+    EngineRenderer.instance.renderer.shadowMap.enabled = getShadowsEnabled()
```
HexaField, line 311: we can remove the redundant `getShadowsEnabled()` here and combine these set operations

packages/engine/src/xr/XRState.ts
```diff
@@ -161,4 +161,5 @@ export const isMobileXRHeadset =
       navigator.userAgent.includes('VR') ||
       navigator.userAgent.includes('AR') ||
       navigator.userAgent.includes('Reality') ||
-      navigator.userAgent.includes('Wolvic')
+      navigator.userAgent.includes('Wolvic') ||
+      navigator.userAgent.includes('Mobile')
```
HexaField, line 165: isMobileXRHeadset should only be true for mobile headsets (ie non Desktop-VR), not headsets AND mobile devices.

Will this trigger for any non-headset mobile devices?

HexaField (changes requested):

packages/engine/src/xr/XRState.ts
DanielBelmes: Ah understood. Removing.

packages/engine/src/scene/systems/ShadowSystem.tsx
DanielBelmes: I thought about it further and also think this doesn't have to be reactive. You're either a mobile XR device, IOS, editor, or not. I don't believe i've seen us swapping shadow render modes.

packages/engine/src/scene/systems/ShadowSystem.tsx
DanielBelmes: This was quite trouble some as it was relying on the groupComponents position updates. So instead of using worldToLocal I changed it to translate group position by that calculated bounding sphere center.

HexaField (approved):

PR #8861, opened 2023-09-20
Model Transform in Cloud

dinomut1:
Refactors Model Transform service to run as a job managed by the API server


packages/server-core/src/assets/ModelTransformLoader.ts
```diff
@@ -67,11 +67,13 @@ export default async function ModelTransformLoader() {
     MOZLightmapExtension,
     EEMaterialExtension
   ])
+  const dracoDecoder = await draco3d.createDecoderModule()
```
HexaField, line 70: parallel promise?

HexaField (changes requested):

speigg (approved):

PR #8874, opened 2023-09-22
Transform system fixes

### LINKED ISSUE #8631: [Bug]: Unparenting an object changes its global transform
If you unparent an object in the scene, it will lose the transform that it currently has. We need to be able to unparent an object and have its transform stay as it is, and its local transform updates to keep its position, rotation and scale constant

### LINKED ISSUE #8847: [Bug]: Parent Transforms Getting Negated in Children
To reproduce:
In any scene, add an image or gltf model to the scene. Offset it slightly, or leave it at origin. the transform of the child element is irrelevant to this bug:

Screenshot from 2023-09-19 12-33-39

Next, add another element to the scene and add the model/image as a child of that element. Adjust the transform on the parent in any way:

Screenshot from 2023-09-19 12-33-27

Next save the scene and refresh. When the scene loads again, the child element will have its transform modified to negate the parent transform:

Screenshot from 2023-09-19 12-34-37

dinomut1:
Changes serialization and deserialization of transform data to create LocalTransformComponents instead of TransformComponents. 

Closes #8847 
Closes #8631 

Dependents:
puttclub: https://github.com/puttclub/puttclub/pull/128
eepro-ecommerce-tools: https://github.com/EtherealEnginePro/eepro-ecommerce-tools/pull/46
ee-development-test-suite: https://github.com/EtherealEngine/ee-development-test-suite/pull/49


packages/engine/src/ecs/functions/EntityTree.ts
```diff
@@ -220,10 +214,9 @@ export function addEntityNodeChild(entity: Entity, parentEntity: Entity, childIn
   const childTransform = getComponent(entity, TransformComponent)
   getMutableState(EngineState).transformsNeedSorting.set(true)
   if (parentTransform && childTransform) {
-    computeTransformMatrix(parentEntity)
-    computeTransformMatrix(entity)
+    // computeTransformMatrix(parentEntity)
```
HexaField, line 217: is this intentionally left commented? if this is unneeded, we should hyst remove it

packages/client-core/src/systems/WidgetUISystem.ts
```diff
@@ -158,7 +159,7 @@ const execute = () => {
   }
   for (const action of registerWidgetQueue()) {
     const widget = RegisteredWidgets.get(action.id)!
-    setLocalTransformComponent(widget.ui.entity, widgetMenuUI.entity)
+    setComponent(widget.ui.entity, EntityTreeComponent, { parentEntity: widgetMenuUI.entity })
```
HexaField, line 162: widgets are not working after changing this

HexaField (changes requested):

packages/client-core/src/systems/WidgetUISystem.ts
dinomut1: added a setComponent(widget.ui.entity, LocalTransformComponent) to here which should fix that

packages/engine/src/ecs/functions/EntityTree.ts
dinomut1: removed

PR #9022, opened 2023-10-09
Mocap Low-Pass Filter

speigg:
Implements a simple low-pass filter on incoming landmark data to cleanup noisy data. 

TODO: use a more advanced predictive filter


speigg:
Just realized, this filter is only applied on every video frame; should be applied on every animation frame

DanielBelmes (approved):

PR #9031, opened 2023-10-10
`parentEntity` is null when asset is dropped into the scene.

CITIZENDOT:
Replace it with SceneEntity if that's the case.

## Summary
copilot:summary

## References
closes #_insert number here_

## Explanation
copilot:walkthrough
copilot:poem

## QA Steps

## Checklist
- If this PR is still a WIP, convert to a draft
- When this PR is ready, mark it as "Ready for review"
- [ensure all checks pass](https://github.com/etherealengine/etherealengine/wiki/Testing-&-Contributing)
- Changes have been manually QA'd 
- Changes reviewed by at least 2 approved reviewers

### DIFF
--- a/packages/editor/src/functions/EditorControlFunctions.ts
+++ b/packages/editor/src/functions/EditorControlFunctions.ts
@@ -221,6 +221,7 @@ const createObjectFromSceneElement = (
   beforeEntity = null as Entity | null,
   updateSelection = true
 ) => {
+  parentEntity = parentEntity ?? getState(SceneState).sceneEntity
   cancelGrabOrPlacement()
 
   const newEntity = createEntity()


packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -221,6 +221,7 @@ const createObjectFromSceneElement = (
   beforeEntity = null as Entity | null,
   updateSelection = true
 ) => {
+  parentEntity = parentEntity ?? getState(SceneState).sceneEntity
```
HexaField, line 224: Why isn't the default parameter catching this?

HexaField (changes requested):

packages/editor/src/functions/EditorControlFunctions.ts
CITIZENDOT: In case `addMediaNode()`, `parent` is explicitly passed as `null`: https://github.com/EtherealEngine/etherealengine/blob/53e4d7199156c0cbc700e96bf6b69e89803c02b0/packages/editor/src/functions/addMediaNode.ts#L61-L70

HexaField: ok, wherever we are passing as null we should pass in as undefined instead, to avoid this redundant double nullish coalesce

HexaField (approved):

PR #9038, opened 2023-10-11
Feat: ✨Updated aws sdk fileUpload

AmadeusMoon:
## Summary
- Update a function that was uploading files to amazon from sdk v2 to sdk v3.

## Explanation
- Changed imports from require syntax to es6
- Changed the import path of `staticResourcePath` to `'../../engine/src/schemas/media/static-resource.schema'`
- Added `Dotenv.config()` as per es6 standard
- Added ! to the 2 env on lines 43 and 44 to rule out undefined
- Changed the syntax of `fileUpload` to sdk3 instead of sdk2

## QA Steps
You can check amazon resources : 
- https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/preview/client/s3/command/PutObjectCommand/
- https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/preview/client/s3/command/HeadObjectCommand/


AmadeusMoon:
@cla-bot check

HexaField (changes requested):
This script has changed a lot since we last used it. We now have an `avatar` table which stores available avatars. We should use the same upload method that the admin panel exposes, found here https://github.com/EtherealEngine/etherealengine/blob/485bab55dadb963639d4476c0814ff7c32b8b0cc/packages/client-core/src/user/services/AvatarService.ts#L51

Although this will only work from the client, this script will need to use a modified call that uploads to the storage provider as shown here https://github.com/EtherealEngine/etherealengine/blob/bdfa29f56be1482c3cc2b7f4083476e29fc615f4/packages/server-core/src/user/avatar/avatar-helper.ts#L168

`avatar-helper.ts` has a lot of good implementations to reference

AmadeusMoon:
Ill check it in a little while, sorry I was busy redoing some stuff because I backtracked and didn't realize what you said. Also I apologize if this was better of being a draft, I did not expect to add changes to it to be honest.

PR #9058, opened 2023-10-15
Feat✨ upload-avatar.ts

AmadeusMoon:
## Summary
Refactored the function that uploads to s3

## Explanation
In `avatar-helper.ts`:
- Created `avatarUploadData` to outscope some functions from `installAvatarsFromProject`  then made it into an export as well and rereferenced the previously hardcoded functions in `installAvatarsFromProject` as the new `avatarUploadData`
- Turned the functions inside `avatarUploadData` to async await because fs functions are blocking
- Added the use of uuidv4 instead of math.round function for `Avatar-id`

In `upload-avatars.ts`:
- Added comments
- Changed imports from require syntax to es6
- Changed the import path of `staticResourcePath`
- Added ! to the 2 env on lines 43 and 44 to rule out undefined
- Changed the syntax of `fileUpload` to sdk3 instead of sdk2
- Moved functions and constants to be in close proximity to their dependencies
- Used uuidv4 for knex id generation
- Renamed `uploadFile` to `uploadToS3`
- Changed the .then syntax to try catch await
- Removed 1 comment since headObject performs check before upload
- Imported `avatarUploadData` and changed dotenv adress to correctly point to local env, also updated address of `avatarPath` to `projects/default-project/assets/avatars`
- Redesigned the `uploadToS3`function :
    - The input is now the same as `avatarUploadData` respectively ( model: Buffer, thumbnail: Buffer, name: string, extension: string, dependencies?: string[])
    - Check for object in db before upload if not found or force it uploads
    - Thumbnail is uploaded if buffer is not 0
    - Dependencies uploaded in metadata model as JSON.stringify if they exist, also if dependencies too big commented out separate dependencies upload
    - `processFile` renamed to `processFiles` and simplified it now calls `uploadToS3` and `saveToDB`, removed overhead.
    - Named the last Promise.all to `avatarData` which now uses a Promise.all to map through the array provided by `avatarUploadData` then calls
`processFiles` for each of them


ozgurhasar (approved):

HexaField (changes requested):
The upload-avatar.ts script can be simplified a lot by using `uploadAvatarStaticResource`  in avatar-helper.ts

AmadeusMoon:
> The upload-avatar.ts script can be simplified a lot by using `uploadAvatarStaticResource` in avatar-helper.ts

@HexaField 

Sorry so let me ask just to be sure : 
- Replace the outscoped function back in place
- Import `uploadAvatarStaticResource` instead

Is this what you intend me to do ? I still do not know how you want things done in your application, I have considered this but I was unsure if you wanted your static resources meddled with before an upload to s3.

Also if the case is what I asked, then should I include metadata about avatar such as what is contained in the data argument inside `uploadAvatarStaticResource` ?

ozgurhasar (approved):

Issue #7094, opened 2022-10-18
[Story] Stricter types


HexaField:
add OpaqueType typings & rename params for the following:

- [ ] `instanceId`: `WorldInstanceServerId`
- [ ] `channelId`: `MediaInstanceServerId`
- [ ] `roomCode`: `RoomCode`
- [ ] `sceneId`: `SceneId`
- [ ] `locationId`: `LocationId`
- [ ] `uuid`: `EntityUUID`
- [ ] `avatarId`: `AvatarId`

AmadeusMoon:
All of the mentioned types are enforced please add more and tick the current ones.

PR #9099, opened 2023-10-20
ChannelId

AmadeusMoon:
## Summary
- [x] ChannelID was already set everywhere except 1 single instance.

## References
#7094


AmadeusMoon:
I didn't know if I should just push this 1 irrelevant change to the other branch but in the end I decided to simply make it a tinny PR. Also the test commit was because I was fixing some type errors that showed up during the npm run test.
It seemed that I was fixing type errors in featherjs and I did not realize I was trying to commit changes to node_module. 😢

PR #9163, opened 2023-10-27
Material plugin prototype

JT00y:
add material plugin prototype,
next step:
how to do apply and remove the plugin prototype
assign internal state


dinomut1:
This is getting implemented differently now, we were close!

PR #9173, opened 2023-10-30
Basic Kit Snapping System

dinomut1:
Creates the AttachmentPointComponent, which is placed at different points on a kit piece to designate contextual snap points. 

Snapping occurs in the AttachmentPointSystem, which keeps track of all AttachmentPointComponents in a scene and snaps two together if they get close enough to one another.

Some things to keep in mind:
- This basic setup of the system iterates through every AttachmentPoint and checks distance between each point every frame. This is O(n^2) complexity which will quickly cause problems as more and more kit pieces are in a scene. Try to restrict the checks to only kit pieces that are currently selected in the studio, or have not been attached to anything yet.
- When the snap occurs, the selected kit piece should snap to the position of the non-selected kit piece. So if there is a floor kit piece in the scene and we are moving a wall piece around, the wall should snap to the position of the floor, and not the other way around. 
- To access the list of currently selected entities, use `const selectionState = useHookstate(getMutableState(SelectionState))`. This will give you access to the selected entities in the studio. However, the AttachmentPointComponent is actually being placed on a child entity to the selected entity. You will need to find the child entities of the selected entities using the `EntityTreeComponent` which keeps track of entity child and parent information.

You will also need to create the AttachmentPoints on the kit pieces by using the Custom Properties in Blender to edit the kit piece .glbs. You'll need to create a new Empty object in Blender for each AttachmentPoint: 
image

On these empty objects, you will need to add custom properties to specify that an entity should be attached to this object, and that an AttachmentPointComponent should be attached to the entity. Here is the configuration of Custom Properties that will do this: 
image
The `xrengine.entity` field just determines that an entity should be created with the provided name. The `xrengine.attachment-point` field declares that an AttachmentPointComponent should be created on this entity.
To create multiple attachment points, duplicate the empty object and position it elsewhere on the 3d model. be sure to name them differently


speigg:
@cla-bot check

PR #9200, opened 2023-11-02
Fix Deleting Nodes in Studio

### LINKED ISSUE #9168: [Bug]: Deleting Node in Studio Causes UI to Crash
### Operating System

Any

### Browser (if applicable)

_No response_

### What happened?

In the studio, open any scene, select a node, and delete it. The UI will disappear and throw a hookstate 111 error: 
image

### Steps to reproduce

```shell
See above
```

### Relevant log output

_No response_

dinomut1:
closes #9168 
Adds checks in HierarchyPanel to prevent rendering nodes after they've been deleted


SYBIOTE (approved):
this works, as additional security, we should investigate why the use effect was not working sometime, i suspect all of this will be revamped anyways so it might not be as relevant.

PR #9207, opened 2023-11-03
Refactor onquery node and add defaults and asserts

SYBIOTE:
This pull request simplifies the onquery node, adds defaults and asserts for nodes, and improves onquery iteration. The changes include:

- Added Assert to getSpline node to ensure a spline entity is selected

- Added default values to videoFit and audioMode nodes

- Renamed getAvatarAnimations to getAnimationPack

- Added default values to setAnimationAction node

- Added default values to addEntity node for componentName and parentEntity

- Improved deleteEntity node by adding a dropdown menu for entity selection

No issues were fixed in this pull request.


anselm (approved):
this seems to work

gheric suggests we find a way to remove the delay loop - but that can happen later

i do think the delay could be abstracted away from here in some way - but whatever works

Issue #9201, opened 2023-11-02
Studio must be obvious about when things are loading or has failed loading


HexaField:
Studio
- [ ] Inspector should  have loading icons for assets/components loading
- [ ] errors too

Client
- [ ] stand in loading icon or 3d gliph
- [ ] error system in 2d/3d too

Immersive scene loading screen looks nice

aditya-mitra:
Error and loading working as intended in the components and in the studio.

PR #9225, opened 2023-11-07
loading and error enhancements in studio

aditya-mitra:
## Summary

Show a loading indicator beside the entity whose components are being loaded.
Also fix the error tooltip which was not properly showing the detailed errors.

_Note:_ I replaced `hasComponent(entity, myComponent) set/removeComponent` with just `set/removeComponent` as both [check for entity before setting/removing](https://github.com/EtherealEngine/etherealengine/blob/e3909f1ce447e609376370c15b00c176fe56e6ce/packages/engine/src/ecs/functions/ComponentFunctions.ts/#L340)

## References
closes part of #9201 

## QA Steps

https://github.com/EtherealEngine/etherealengine/assets/55396651/525a66a7-4785-464b-88a8-c1b3c42d2089

Screenshot 2023-11-10 at 7 16 02 PM


AidanCaruso (approved):
Looks really good, definitely improves the loading/error tips. However, I feel the need to point out that this is hampered a bit by the model component's error handling logic generally being wrong. Nonexistant models returning 404s seemingly aren't always caught if they are of a supported asset type, and invalid URLs get flagged as invalid asset types. This can be fixed in another PR down the line.

PR #9237, opened 2023-11-08
file browser configurable top limit

### LINKED ISSUE #9220: File browser should have top level limit
We do not want admins or studio users being able to get above the `projects` folder
Screenshot from 2023-11-07 16-04-08

in both the studio and the admin panel, we need to present the file browser via the project being accessed. 

example: in the studio or admin panel, if i open the file browser view, it should open to the project that is active (via the active project in the studio, or in the projects page "View Project Files" button in the admin panel)

### LINKED ISSUE #9221: File browser should not be restricted to `projects` folder

aditya-mitra:
## Summary

Do not allow to explore all directories in the file browser panels.
Suppress going back when exploring particular project.
Also allow [configuring the top limit](https://github.com/EtherealEngine/etherealengine/pull/9237/files#diff-51087161442f68c5e789e41eda8110c8d47f29b9b6776133f73607960f40c0fa)

## References
closes #9220
closes #9221 

## QA Steps

Back button not present when in the selected project
image

Back button is shown when we go atleast one folder deep
image

Same for studio
image


barankyle (changes requested):
There's an issue with clicking on 'projects' in the breadcrumb trail, the part of the file browser that shows the path separated by '>' and lets you click up the folder structure. If you click on 'projects', the API is returning the new error about not being allowed to access that folder, but the frontend doesn't handle that error in any way, and the file browser gets stuck at 'Loading files'. 

Ideally, the breadcrumb trail would not show you a folder level that you're not allowed to select - the first item would be <projectName>. That way, users aren't even aware that there's a folder they can't access. If for some reason that proves overly difficult to implement, then the file browser should catch the new error that the API is returning, remove the loading message, and probably make a notification that that folder is inaccessible.

aditya-mitra:
> There's an issue with clicking on 'projects' in the breadcrumb trail, the part of the file browser that shows the path separated by '>' and lets you click up the folder structure. If you click on 'projects', the API is returning the new error about not being allowed to access that folder, but the frontend doesn't handle that error in any way, and the file browser gets stuck at 'Loading files'.
> 
> Ideally, the breadcrumb trail would not show you a folder level that you're not allowed to select - the first item would be . That way, users aren't even aware that there's a folder they can't access. If for some reason that proves overly difficult to implement, then the file browser should catch the new error that the API is returning, remove the loading message, and probably make a notification that that folder is inaccessible.

Implemented!

Screenshot 2023-11-16 at 6 11 01 PM

barankyle (approved):

PR #9243, opened 2023-11-09
System Insertion Order Explicit in Definitions

HexaField:
## References
closes #_insert number here_


AidanCaruso (changes requested):
Nice, most everything I tried works great. I'm a big fan of these changes, also noticed moving some side effects outside of state. Sadly though it looks like the capture page hangs on loading. I assume this change may have something to do with the avatar not spawning on the capture route (which is great, just likely needs a logic change somewhere). If the avatar were to spawn, it looks like networked animation actions and movement would still be an issue (unless I missed something), so it may still be worthwhile to perhaps think about disabling that functionality by reacting to some data that is only set by the capture page. Just a thought.

PR #9245, opened 2023-11-09
Use PBR material when appropriate

CITIZENDOT:
## Summary
When `excludeNormals` is enabled in the manifest, UVOL uses `MeshBasicMaterial`'s shader. Otherwise `MeshStandardMaterial`'s shader. Ofcourse they aren't used directly though.

## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/UVOL2Component.ts
+++ b/packages/engine/src/scene/components/UVOL2Component.ts
@@ -30,11 +30,15 @@ import {
   CompressedTexture,
   Group,
   InterleavedBufferAttribute,
+  Matrix3,
   Mesh,
   MeshBasicMaterial,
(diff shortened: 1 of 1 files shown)


speigg:
This should be a runtime feature, not part of the manifest imo. We want to be able to switch between PBR material, and baked textures depending on the device, and/or depending on quality settings.

CITIZENDOT:
When we have normal attributes in the geometry, there is no reason to not use them right? If they're stripped away at the encoder, we must use unlit material.

speigg:
We want to be able to switch between PBR (e.g., albedo, normal, roughness, metalness, emissive), and baked (just albedo, with lighting baked in).

speigg (approved):

PR #9255, opened 2023-11-10
Mocap pose based interactions

AidanCaruso:
## Summary
Lays the groundwork for an extensible pose based interaction system. The first mocap pose based event included in this PR is seat detection via interaction actions recieved by mount point components. Sitting with full body motion capture next to a seat will run the mount point component's seated logic. Standing up should also cause the avatar to leave the seat.

The interaction action/onInteract function has been replaced in favor of directly accessing input sources through the input source component, this also required some code changes in the grabbable system. In a similar vein, the local pose component is used to manage mocap pose inputs after pose detection. For now, the only integrated poses are sitting and standing, though pose types can be extended to any theoretically detectable pose. Pose detection for the seated/standing position is done through a reusable function that compares upper leg angles relative to the hips.

## References
closes #_insert number here_

## QA Steps

- Go to studio/default-project/apartment
- Add an entity with a mount point component
- Visit location/apartment
- Stand near the mount point component in the location
- Open the capture page
- Sit down in view of the camera
- Avatar should snap to the mount
- Stand up
- Avatar should dismount


packages/engine/src/mocap/poseToInput.ts
```diff
@@ -0,0 +1,76 @@
...
+    let metTargetStateAngle =
+      rig.rightUpperLeg.node.quaternion.angleTo(rig.spine.node.quaternion) < minSeatedAngle &&
+      rig.leftUpperLeg.node.quaternion.angleTo(rig.spine.node.quaternion) < minSeatedAngle
+    metTargetStateAngle = toPose == 'sitting' ? !metTargetStateAngle : metTargetStateAngle
+
+    if (!metTargetStateAngle || pose[toPose].value) return false
+
+    poseHoldTimer += deltaSeconds
+    if (poseHoldTimer > poseHoldTime) {
+      console.log(toPose)
```
DanielBelmes, line 59: 👀

packages/engine/src/mocap/MotionCaptureInputSystem.ts
```diff
@@ -0,0 +1,45 @@
...
+import { hasComponent, setComponent } from '../ecs/functions/ComponentFunctions'
+import { InputSystemGroup } from '../ecs/functions/EngineFunctions'
+import { defineSystem } from '../ecs/functions/SystemFunctions'
+import { MotionCapturePoseComponent } from './MotionCapturePoseComponent'
+import { MotionCaptureRigComponent } from './MotionCaptureRigComponent'
+import { evaluatePose } from './poseToInput'
+
+export const execute = () => {
+  const entity = Engine.instance.localClientEntity
+  if (!hasComponent(entity, MotionCaptureRigComponent)) return
```
DanielBelmes, line 36: Wanted to say this probably covers the client only check. But it wouldn't hurt to explicitly check if we're in client.

DanielBelmes (approved):
Pointed some things out. I approve. I think we need @HexaField networking eyes as i have not downloaded that info into my brain.

packages/engine/src/mocap/poseToInput.ts
AidanCaruso: Oops 🔢

HexaField (approved):

PR #9261, opened 2023-11-10
Update bug template

DanielBelmes:
## Summary

- Updates bug issue labels so that they work
- Adds version/commit field
- Splits up what happened into two logical text areas
- Makes log area a general area


DanielBelmes:
CLA BOT OMG

PR #9271, opened 2023-11-12
ScopeType enforced

AmadeusMoon:
## Summary
- Changed `interface ScopeType` to `interface ScopeTypeInterface`
- Added `ScopeType` opaque type to `scope.schema.ts`
- [x] scopeType is now ScopeType

## References
#7094 

## Checklist
✔️ Ran npm run check-errors, no related problems
✔️ Ran npm run check, no related problems


packages/engine/src/schemas/scope/scope-type.schema.ts
```diff
@@ -35,7 +37,9 @@ export const scopeTypeMethods = ['find', 'get'] as const
 // Main data model schema
 export const scopeTypeSchema = Type.Object(
   {
-    type: Type.String(),
+    type: TypedString<ScopeType>({
+      format: 'uuid'
```
hanzlamateen, line 41: @HexaField this shouldnt be uuid?

HexaField: Oh, perhaps. Does that affect the database?

hanzlamateen: It doesn't but maybe at some time in future it will. I have made the fix in https://github.com/EtherealEngine/etherealengine/pull/9331

PR #9275, opened 2023-11-13
CustomRoute fix

AmadeusMoon:
## Summary
`Public.tsx` was redone and customRoutes were handled only with the base route and not including the aditions.


AidanCaruso (approved):

AidanCaruso:
Could you use this PR to look into related issues with css not loading?

PR #9294, opened 2023-11-14
loadEngineInjection

anselm:
## Summary

load engine injection

## References
closes #_insert number here_


speigg (changes requested):
Typescript error:
```ts
src/behave-graph/nodes/Profiles/Engine/Events/onCollision.ts(105,5): error TS2304: Cannot find name 'startSystem'.
```

packages/engine/src/scene/systems/TriggerSystem.ts
```diff
@@ -35,7 +35,7 @@ import { UUIDComponent } from '../components/UUIDComponent'
 
 export const triggerEnter = (entity: Entity, triggerEntity: Entity, hit: ColliderHitEvent) => {
   const triggerComponent = getComponent(triggerEntity, ColliderComponent)
-  if (!Array.isArray(triggerComponent.triggers)) return
+  if (!triggerComponent || !Array.isArray(triggerComponent.triggers)) return
```
HexaField, line 38: rather than checking for the trigger component, we should add it to the `collisionQuery`, we can then ensure it exists

HexaField (changes requested):

packages/engine/src/scene/systems/TriggerSystem.ts
anselm: There is no concept TriggerComponent to query against. It is just a mode... and sometimes it is null. In fact avatars don't even have a ColliderComponent apparently?

anselm: It feels like that code has some kind of defect conceptually. Imagine a pairwise collision between a trigger volume and an avatar. The avatar in this case will be the triggerEntity (the entity that triggered the collision). This code is then looking through the avatar for triggers ...

Maybe the intention was to fetch the list of triggers from the trigger volume - so:

  const triggerComponent = getComponent(triggerEntity, ColliderComponent)
  
Should be: 
 
  const triggerComponent = getComponent(entity, ColliderComponent)

?

HexaField: oh i see, i think this can be cleaned up and work as i said, now that we have CollisionComponent being generated when there is data and removed when not

anselm: Um, ok, but there are conceptual issues still:

 - things entering that code don't always have a trigger -> either because avatar doesn't have one directly, OR because it is the receiving side of a triggerable interaction? which shouldn't need a trigger
 
 - that code may just be broken / may never have worked / may never have been tested? the test I added to block the crash may be actually due to testing the wrong object for a trigger?

anselm: I think it's best if I revert TriggerSystem.ts and treat that as a separate issue rather than trying to merge that in with this commit since this commit is languishing. The issue doesn't come up for me now due to fixes elsewhere, and examining TriggerSystem can be done as a separate commit.

HexaField (approved):

PR #9296, opened 2023-11-14
show specific errors for model component

### LINKED ISSUE #9272: [IMPROVE] : Model Component's error handling
References: https://github.com/EtherealEngine/etherealengine/pull/9225#pullrequestreview-1725322528

aditya-mitra:
## Summary

Show an `INVALID_SOURCE` error if the asset loader failed to fetch the source. 

## References
closes #9272 

## QA Steps

Screenshot 2023-11-14 at 5 17 15 PM


packages/engine/src/scene/functions/ErrorFunctions.ts
```diff
@@ -45,7 +45,7 @@ export const addError = <C extends Component>(
   message?: string
 ) => {
   console.error('[addError]:', entity, Component.name, error, message)
-  if (!hasComponent(entity, ErrorComponent)) addComponent(entity, ErrorComponent)
+  if (!hasComponent(entity, ErrorComponent)) setComponent(entity, ErrorComponent)
```
AidanCaruso, line 48: We also need logic to remove the error component. At the moment the last error set remains even when the asset loader does its thing successfully.

AidanCaruso (commented):
Looks good, just needs a minor change / logic to remove the error component.

AidanCaruso (changes requested):
Needs minor changes referenced above

packages/engine/src/scene/components/ModelComponent.ts
```diff
@@ -170,6 +169,7 @@ function ModelReactor() {
             (loadedAsset) => {
               loadedAsset.scene.animations = loadedAsset.animations
               if (!entityExists(entity)) return
+              removeError(entity, ModelComponent, 'INVALID_SOURCE')
```
AidanCaruso, line 172: I'm still seeing lingering error messages and error component after changing the source path to an invalid extension/url and then back to the correct path. 
image
This is probably because of the above return condition preventing the asset load logic below in this case. We should probably attempt to remove the errors before returning there.

aditya-mitra: [abc559fb-1717-4ad1-bd15-b8055260da8a.webm](https://github.com/EtherealEngine/etherealengine/assets/55396651/6bc1871d-a609-4105-9cb3-cdb0f32675a2)

Fixed this!

packages/engine/src/scene/components/ModelComponent.ts
```diff
@@ -143,7 +143,11 @@ function ModelReactor() {
 
   // update src
   useEffect(() => {
-    if (model.src === model.scene?.userData?.src) return
+    if (model.src === model.scene?.userData?.src) {
```
HexaField, line 142: this is not correct. if the source is the same, it should be a noop as it is currently

HexaField (changes requested):

packages/engine/src/scene/components/ModelComponent.ts
aditya-mitra: Actually this was the reason why the errors where persisting when the original model URL was used.

Using another URL source removes the error when the asset is loaded in [line 170](https://github.com/EtherealEngine/etherealengine/pull/9296/files#diff-daae2524b717be2ba24820c9515621d8714efd6e8d8bed8493853a425dc305cdR170). 
But, when the same source is used the errors are not removed because of the early return.

HexaField (approved):

Issue #8851, opened 2023-09-20
Send client logs to server before closing


hanzlamateen:
We should add an unload/close event handler engine's `logger.ts` to send client logs to the server if the browser window/tab was closed.
Currently logger just send the logs from client to server after 15 seconds scheduled interval.

#8022

PR #9297, opened 2023-11-14
Added log sending on early refresh/close

AmadeusMoon:
## Summary
- [x] `logger.ts` now sends logs if window is refreshed or tab is closed or minimized. ( Not guaranteed to function properly if the whole browser is closed )

## References
#8851 

## QA Steps
✔️ Ran `npm run check-errors`, no related errors found
✔️ Ran `npm run dev`, no errors. 
 

Edit: beforeunload event was changed to visibity event for better compatibility on phones.


packages/engine/src/common/functions/logger.ts
```diff
@@ -48,6 +48,12 @@ const engineCache = new NodeCache()
 // Schedule the data push at a certain time (e.g., every hour)
 schedule.scheduleJob('*/15 * * * * *', pushToEngine)
 
+// Client-side early close/refres send logs
+if (typeof window !== 'undefined') {
+  console.log('Window unload event triggered. Sending logs...')
```
DanielBelmes, line 53: This console log will fire when this file is loaded on the client. I think you want to add this console log into the event Listner function. So maybe wrap pushToEngine in a lambda/anonymousfunc and add the console log there.

DanielBelmes (changes requested):

packages/engine/src/common/functions/logger.ts
AmadeusMoon: Oh ok thank you, I didn't notice, I assumed the log appearing at the top was simply a side effect of how logs are displayed.

AmadeusMoon: You're right that the console fires when the if statement is met, I have specified how to test the function in the last commit message and this is what I have done: 

- Tried modifying placement of log, still does not work
- Log does not show in vscode console due to its client sided nature
- I have removed it all together, 

If you want a log you : 

- Can send a post request to server and fire a log function in response
- Create a sort of notification on unload asking if the user wants to leave ?

DanielBelmes:
Sorry I've been on a mini vacation since Friday. Have you seen the `beforeunload` event? https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event

Maybe it will fix the race condition you're seeing

PR #9300, opened 2023-11-15
Locations refer to scene by path rather than name

### LINKED ISSUE #9263: Locations refer to scene by path rather than name
- replace `api.service(scenePath).get()` with `fetch(location.scenePath)`
- migrate existing scenes from `projectName/sceneName` to url path
- `/studio/projectName/sceneName` becomes `/studio/projecName` and load scene without redirecting
- places that use the SceneID can remain the same, the sceneId just becomes the scene url path

MoizAdnan:
## Summary

- Location.sceneId now contains the path to scene file instead of 'projectName/sceneName'. 
- Scenes load without redirecting
- Changes in scene class for multitenancy

## References
closes #9263


speigg:
This is great, much simpler.

speigg (approved):

PR #9301, opened 2023-11-15
Changes to introduce user-avatar table

hanzlamateen:
## Summary
Introduced `user-avatar` table to store user's avatar. Previously, this information was stored in user table. Also added migration to move avatarId from user table to user-avatar table and remove avatarId column from user table.

## References
closes #_insert number here_


barankyle:
One minor thing, when I change avatars in a location, the displayed thumbnail doesn't update, either in the user settings modal or in any UserMediaWindows (both for myself and for other clients).

hanzlamateen:
> One minor thing, when I change avatars in a location, the displayed thumbnail doesn't update, either in the user settings modal or in any UserMediaWindows (both for myself and for other clients).

Must be fixed now.

barankyle (approved):

PR #9303, opened 2023-11-15
Check for ShadowComponent in UVOL

CITIZENDOT:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/UVOL1Component.ts
+++ b/packages/engine/src/scene/components/UVOL1Component.ts
@@ -44,11 +44,11 @@ import { EngineState } from '../../ecs/classes/EngineState'
 import {
   defineComponent,
   getMutableComponent,
-  getOptionalMutableComponent,
   hasComponent,
   removeComponent,
   setComponent,
-  useComponent
+  useComponent,
+  useOptionalComponent
 } from '../../ecs/functions/ComponentFunctions'
 import { AnimationSystemGroup } from '../../ecs/functions/EngineFunctions'
 import { useEntityContext } from '../../ecs/functions/EntityFunctions'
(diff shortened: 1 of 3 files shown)


packages/engine/src/scene/components/UVOL1Component.ts
```diff
@@ -160,9 +161,11 @@ function UVOL1Reactor() {
     if (volumetric.useLoadingEffect.value) {
       setComponent(entity, UVOLDissolveComponent)
     }
-    const shadow = getMutableComponent(entity, ShadowComponent)
-    shadow.cast.set(true)
-    shadow.receive.set(true)
+    const shadow = getOptionalMutableComponent(entity, ShadowComponent)
```
HexaField, line 164: we should put these in other useEffects
```ts
const shadow = useOptionalComponent(entity, ShadowComponent)
useEffect(() => {
    if (shadow) {
      shadow.cast.set(true)
      shadow.receive.set(true)
    }
    }, [shadow])
```

HexaField (changes requested):

packages/engine/src/scene/components/UVOL1Component.ts
CITIZENDOT: Done

packages/engine/src/scene/components/UVOL2Component.ts
```diff
@@ -460,6 +452,18 @@ transformed.z += mix(keyframeA.z, keyframeB.z, mixRatio);
     }
   }, [])
 
+  useEffect(() => {
+    if (!shadow) return
+    if (manifest.current.type === UVOL_TYPE.UNIFORM_SOLVE_WITH_COMPRESSED_TEXTURE && shadow) {
```
HexaField, line 457: these second `shadow` checks are unnecessary

HexaField (changes requested):

packages/engine/src/scene/components/UVOL2Component.ts
CITIZENDOT: Fixed.

HexaField (approved):

PR #9304, opened 2023-11-16
@Swc/core update to newer version

### LINKED ISSUE #9116: Pin @swc/core to something newer than 1.3.41
It seems like some recent changes to @swc/core don't play nicely with ts-node; running the latest usually throws errors about experimental flags being used that ts-node won't run without more explicit flags to do so

```
Error: @swc/core threw an error when attempting to validate swc compiler options.
You may be using an old version of swc which does not support the options used by ts-node.
Try upgrading to the latest version of swc.
Error message from swc:
Failed to deserialize buffer as swc::config::Options
JSON: {"sourceMaps":true,"module":{"noInterop":false,"type":"commonjs","strictMode":true,"ignoreDynamic":false},"swcrc":false,"jsc":{"parser":{"syntax":"typescript","tsx":false,"decorators":true,"dynamicImport":true,"importAssertions":true},"target":"es2022","transform":{"decoratorMetadata":true,"legacyDecorator":true,"react":{"throwIfNamespace":false,"useBuiltins":false}},"keepClassNames":true,"experimental":{"keepImportAssertions":true}}}

Caused by:
    unknown field `keepImportAssertions`, expected one of `plugins`, `keepImportAttributes`, `emitAssertForImportAttributes`, `cacheRoot`, `disableBuiltinTransformsForInternalTesting` at line 1 column 438
    at createVariant (/app/node_modules/ts-node/src/transpilers/swc.ts:262:13)
    at createSwcOptions (/app/node_modules/ts-node/src/transpilers/swc.ts:211:25)
    at create (/app/node_modules/ts-node/src/transpilers/swc.ts:56:41)
```

The solution seemed to have been to pin this at the version it was at, 1.3.41, but it's currently at something like 1.3.88. If we can't resolve the flag issues, we could at least find the most recent version that it will run at and pin it to that.

AmadeusMoon:
## Summary

- [x] Updated `@swc/core` from 1.3.41 to 1.3.96
- Added `@swc/counter: ^0.1.2`
- And `@swc/types: ^0.1.5`
- Removal of empty lines in package.json files due to the custom script I ran to modify the
`@swc/core` version

```
ee-bot and ee-develoment-test-suite also require @swc/core version update
```

## References
closes #9116 

## QA Steps
✔️ Ran `npm run check-errors`, no issues found
✔️ Ran `npm run check`, also no issues
✔️ Ran dev, everything seems functional


swc.js
AidanCaruso, line 1: Looks like you accidentally included an automation script

AmadeusMoon: Oh, sorry, I commited it by accident together with the changes, thanks for letting me know.

PR #9311, opened 2023-11-20
Portals Work Again

HexaField:
## Summary

Portals are working again. Largely just cleaned up scope of responsibility of various aspects, as well as migrating the hyperspace system into the hyperspace component reactor for better clarity and binding of data.

They still need a lot of cosmetic upgrade, but I will leave that for another PR.

## References
closes #_insert number here_


packages/engine/src/scene/components/HyperspaceTagComponent.ts
```diff
@@ -23,6 +23,151 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
+  onRemove(entity, component) {
+    removeEntity(component.ambientLightEntity.value)
+    removeEntity(component.hyperspaceEffectEntity.value)
+  },
+
+  reactor: () => {
+    const entity = useEntityContext()
+    const { transition, hyperspaceEffectEntity } = getComponent(entity, HyperspaceTagComponent)
+    const hyperspaceEffect = getComponent(hyperspaceEffectEntity, GroupComponent)[0] as any as PortalEffect
+    const cameraTransform = getComponent(Engine.instance.cameraEntity, TransformComponent)
```
speigg, line 105: Shouldn't these be useComponent ?

speigg (changes requested):

packages/engine/src/scene/components/HyperspaceTagComponent.ts
speigg: I suppose these are intentionally non-reactive. Should be fine.

speigg (approved):

PR #9322, opened 2023-11-21
Updates for allowing domain-related hooks and logic

barankyle:
## Summary

Updated code to support header population on internal calls.
    
Internal listeners that send messages to interested clients need
to persist headers on calls, as original headers are not preserved.

Fixed issues with email/SMS generation.

Made some resolver get's internal _get's, in cases where headers
may not be available at that point. All instances are cases where
headers might actively interfere with the process, e.g. looking
up static resources that might not have a field populated from
headers.

Converetd uploadMediaStaticResource and uploadRecordingStaticResource
into a single new service. uploadRecordingStaticResource technically
still exists, but it just uses its internal reference to the API
to call the new service. The code in both places was nearly identical,
and having it as a single service allows for header usage from outside
projects.

## References
closes #_insert number here_


packages/server-core/src/recording/recording-resource-upload/recording-resource-upload.class.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const recordingResource = await this.app.service(recordingResourcePath).create({
+      recordingId: recordingID,
+      staticResourceId: staticResource.id
+    })
+
+    await uploadPromise
+
+    await this.app.service(recordingResourcePath).patch(recordingResource.id, {
+      updatedAt: new Date().toISOString().slice(0, 19).replace('T', ' ')
```
hanzlamateen, line 80: You can use `getDateTimeSql` method here.

hanzlamateen (commented):
I have gone through the code and it looks good. Though havn't got a chance to test the changes locally.

hanzlamateen:
@barankyle used this branch of EE, standalone. Without any project, etc. Just `default-project`. Created a recording on capture page and the getting various errors on recordings page in admin panel.

https://github.com/EtherealEngine/etherealengine/assets/10975502/04fcda7e-4bae-478c-a5c4-2ae4a4e5478d

HexaField (changes requested):
I don't like putting headers in the actions. It's a security risk and can be done cleaner.

hanzlamateen:
Do we need these changes after https://github.com/EtherealEngine/etherealengine/pull/9371

packages/server-core/src/media/upload-asset/upload-asset.hooks.ts
```diff
@@ -36,7 +36,7 @@ export default {
     all: [logRequest()],
     find: [disallow()],
     get: [],
-    create: [setLoggedInUser('userId')],
+    create: [(context) => { console.log('upload-asset.create', context.params.headers) }, setLoggedInUser('userId')],
```
hanzlamateen, line 39: I think this can be removed?

packages/server-core/src/recording/recording-resource-upload/recording-resource-upload.hooks.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import { disallow } from 'feathers-hooks-common'
+import { SYNC } from 'feathers-sync'
+
+export default {
+  before: {
+    all: [],
+    find: [disallow()],
+    get: [disallow()],
+    create: [
```
hanzlamateen, line 34: Anyone can upload recording or doe we need some verifyScope check here?

hanzlamateen (changes requested):
`check-errors` seems to be failing.

packages/server-core/src/recording/recording-resource-upload/recording-resource-upload.hooks.ts
barankyle: I made this an internal-only service.

packages/server-core/src/media/upload-asset/upload-asset.hooks.ts
barankyle: Yup, done.

hanzlamateen (approved):

speigg (approved):

PR #9339, opened 2023-11-27
show thumbnail of scene on scene save as

aditya-mitra:
## Summary

Show the thumbnail of scene when the scene is saved as. Also, make the _saveasscene dialog_'s styling better.

#### Earliier

image

#### Now

image


packages/editor/src/components/EditorContainer.tsx
```diff
@@ -184,11 +184,12 @@ const onSaveAs = async () => {
   try {
     if (sceneName || editorState.sceneModified.value) {
       const blob = await takeScreenshot(512, 320, 'ktx2')
+      const thumbnailBlob = await takeScreenshot(256, 160, 'jpeg')
```
HexaField, line 187: we dont want to be taking two screenshots, we are able to display ktx2s as images

HexaField (changes requested):
We should also include the thumbnail in the regular save menu.

mrhegemon (approved):

mrhegemon (approved):

packages/editor/src/functions/takeScreenshot.ts
```diff
@@ -162,10 +162,10 @@ export async function previewScreenshot(
 export async function takeScreenshot(
   width: number,
   height: number,
-  format = 'ktx2' as 'png' | 'ktx2' | 'jpeg',
+  format = 'jpeg' as 'jpeg' | 'png',
```
mrhegemon, line 165: Huh? Why these formats?

mrhegemon (changes requested):

aditya-mitra:
> We should also include the thumbnail in the regular save menu.

Implemented!

image

HexaField (approved):

PR #9345, opened 2023-11-28
Updated S3 storage provider to make most project files private

barankyle:
When adding files to S3, any project files not in a top-level `public` folder will have a private ACL so as not to expose code. Anything within the top-level `public` folder will remain as public-read, no matter how deep, but `public` folders in other project folders will not be made public.

## References
closes #_insert number here_

### DIFF
--- a/packages/common/src/constants/ProjectKeyConstants.ts
+++ b/packages/common/src/constants/ProjectKeyConstants.ts
@@ -1,0 +1,27 @@
+/*
+CPAL-1.0 License
+
+The contents of this file are subject to the Common Public Attribution License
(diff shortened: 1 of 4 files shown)


hanzlamateen:
Is this PR ready for review? As its still in draft.

barankyle:
> Is this PR ready for review? As its still in draft.

It's ready for review. I left it as a draft so that it doesn't get accidentally merged in before we're ready for it. We'll need to change all of our existing projects to have their assets in /public so that they're accessible, otherwise the engine gets very broken since key files end up being private.

HexaField (approved):

PR #9348, opened 2023-11-29
Fix flicker on new entity select

SYBIOTE:
This pull request fixes the issue of flickering when selecting a new entity in the CoreNodeEditor component. The flickering was caused by an unnecessary re-rendering of the component due to an unused selectionState variable. This PR removes the unused variable and optimizes the rendering process.


dinomut1 (approved):
great jaerb

PR #9353, opened 2023-11-29
Position interpolation passes for avatar and mocap solve degradation

AidanCaruso:
## Summary

Stacked on mocap-pose-interaction, adds the InterpolatePosition component for use with mount points, and replaces the boolean lower body solve property with a float factor for smooth interpolation between animations and solved poses.

## References
closes #_insert number here_


packages/engine/src/mocap/MotionCaptureSystem.ts
```diff
@@ -138,16 +138,24 @@ const execute = () => {
...
-          boneName == VRMHumanBoneName.RightFoot
+      if (
+        boneName == VRMHumanBoneName.LeftUpperLeg ||
+        boneName == VRMHumanBoneName.RightUpperLeg ||
+        boneName == VRMHumanBoneName.LeftLowerLeg ||
+        boneName == VRMHumanBoneName.RightLowerLeg ||
+        boneName == VRMHumanBoneName.LeftFoot ||
+        boneName == VRMHumanBoneName.RightFoot
+      ) {
+        localbone.quaternion.slerp(
```
HexaField, line 151: we have fastSlerp injected onto the quaternion class which is preferable to slerp

HexaField (changes requested):

PR #9366, opened 2023-12-02
entity component error debugger, various bug fixes

HexaField:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/renderer/components/AnimateScaleComponent.ts
+++ b/packages/engine/src/renderer/components/AnimateScaleComponent.ts
@@ -35,7 +35,7 @@ import {
   setComponent,
   useComponent
 } from '../../ecs/functions/ComponentFunctions'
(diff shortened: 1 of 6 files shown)


packages/engine/src/renderer/components/AnimateScaleComponent.ts
```diff
@@ -63,18 +63,19 @@ export const AnimateScaleComponent = defineComponent({
...
+        animateScale(entity, originalScale)
       }
     }, [])
 
     return null
   }
 })
 
-const animateScale = (entity: Entity, newScale: Vector3, grow: boolean) => {
+const animateScale = (entity: Entity, newScale: Vector3) => {
```
DanielBelmes, line 78: wonder when we lost `grow`. Haven't tested this is a minute. Hopefully it still shrinks.

DanielBelmes (approved):

PR #9367, opened 2023-12-02
Resource Tagging

dinomut1:
## Summary
In preparation for the Assets Panel, we need to tag all assets in every project with appropriate categories.
This PR introduces a UI for manually adding tags, as well as crude auto-tagging based on asset class.
Also, static resources is now automatically seeded with every valid project asset on project upload.


HexaField (changes requested):

packages/server-core/src/projects/project/project.class.ts
```diff
@@ -164,7 +164,6 @@ export class ProjectService<T = ProjectType, ServiceParams extends Params = Proj
       promises.push(uploadLocalProjectToProvider(this.app, projectName))
     }
 
-    await Promise.all(promises)
```
barankyle, line 167: Need to restore this line too.

PR #9374, opened 2023-12-04
Refactor transform gizmo component to ECS

SYBIOTE:
This pull request refactors the transform gizmo component to use the Entity Component System (ECS) architecture. It includes several commits that convert the existing code to ECS, streamline variables and variable names, add comments, and fix issues with filtered selection. The changes also include the addition of new transform axis constants and the implementation of a reactor function for handling selection and gizmo updates. This refactor improves the maintainability and extensibility of the transform gizmo component.


HexaField (changes requested):
This is a good start.

We can improve this a few ways.

1. We want a `local` mode in which the TransformGizmoComponent should be added to an entity, rather than a separate entity, and be a thing that modifies that entity's transform in local space.

2. We want a `global` mode in which the TransformGizmoComponent is added to a new entity, that modifies it's target entity in world space. Currently, it acts as though it is in world space yet for nested objects, they are modified in their own local space and a desync occurs between the gizmo.

3. the TransformGizmo class is still a threejs object, with it's events and class based implementation. This should be refactored into the gizmo component itself, such that it uses the component's reactor and the ECS spatial input instead of threejs' custom input handlers.

speigg:
Gizmos should always be in world space. Trying to make them a child is going to lead to problems.

HexaField:
> Gizmos should always be in world space. Trying to make them a child of the entity they are controlling is going to lead to problems. 

Why should they always be in world space?

What if I want to drag a position along it's local axis, rather than the world axis?

speigg:
> Why should they always be in world space?
Because you don't want to have to deal with the complexity of a child entity attempting to control it's parent's movement, it's complex is several subtle ways, including the need to undo any scale (uniform or not) or skew in the parent transform. TLDR; it's overly complicated and not necessary to attach to the gizmo to the child in order to move the entity in it's local space. We should just mirror the way that this works in threejs; the gizmo always exists in the world-space, and it can manipulate a target object in world-space or it's local-space. 

What if I want to drag a position along it's local axis, rather than the world axis?

speigg:
The difficulty comes from using full transformation matrices in the scene graph; if we switch to a pose (position/orientation) model, then it's feasible to attach gizmo entities directly to the entity without too much trouble. 

Another option is to have a TransformGizmoComponent control a separate entity in scene-space, and just use the component as a tag to indicate that the entity should have a gizmo associated with it.

SYBIOTE:
thats what is esentially happening now tbh, the transform component is more like a tag for the controlled entity, the gizmo entity, has the controls and the visual elements in the group which are responsible for gizmo manipulation, the changes are then passed onto the controlled entity

HexaField (approved):

PR #9380, opened 2023-12-05
fix guest avatar upload

aditya-mitra:
## Summary

Allow a guest user (which does not have any scope) to upload avatars.

## QA Steps

[721a6732-ba5e-40e8-9655-247962ba0705.webm](https://github.com/EtherealEngine/etherealengine/assets/55396651/8e329d39-e0ee-4289-8812-fe654327d3c9)

### DIFF
--- a/packages/server-core/src/user/avatar/avatar.hooks.ts
+++ b/packages/server-core/src/user/avatar/avatar.hooks.ts
@@ -37,9 +37,11 @@ import {
 import setLoggedInUser from '@etherealengine/server-core/src/hooks/set-loggedin-user-in-body'
 import logger from '../../ServerLogger'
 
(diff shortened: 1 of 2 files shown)


packages/server-core/src/user/avatar/avatar.hooks.ts
```diff
@@ -101,6 +103,23 @@ const ensureUserAccessibleAvatars = async (context: HookContext<AvatarService>)
   }
 }
 
+const checkUserHasPermissionOrIsOwner = async (context: HookContext<AvatarService>) => {
+  if (await checkScope(context.params.user!, 'globalAvatars', 'write')) {
```
hanzlamateen, line 107: Ideally the response of await operation should be stored in a variable and then that variable should be used in if condition.

hanzlamateen (approved):

PR #9386, opened 2023-12-06
Refactor Project State

HexaField:
## References
closes #_insert number here_


DanielBelmes (approved):

speigg:
Is there a good reason we aren't using our feathers hooks (e.g. `useService`) for interacting w/ the api here instead of doing manual bookkeeping w/ needsUpdate?

HexaField:
> Is there a good reason we aren't using our feathers hooks (e.g. `useService`) for interacting w/ the api here instead of doing manual bookkeeping w/ needsUpdate?

only because the service is used in the admin panel and i didnt want to refactor it all now since it's very important to keep working

PR #9388, opened 2023-12-06
Adding digital ocean cluster connection setup

muradkhateeb78:
## References
closes #_insert number here_


scripts/run-builder.sh
```diff
@@ -12,7 +12,13 @@ mkdir -pv ~/.docker
 cp -v /var/lib/docker/certs/client/* ~/.docker
 touch ./builder-started.txt
 bash ./scripts/setup_helm.sh
-bash ./scripts/setup_aws.sh $EKS_AWS_ACCESS_KEY_ID $EKS_AWS_ACCESS_KEY_SECRET $AWS_REGION $CLUSTER_NAME
+if [[ "$CLOUD_PROVIDER" == "aws" ]]; then
+  bash ./scripts/setup_aws.sh $EKS_AWS_ACCESS_KEY_ID $EKS_AWS_ACCESS_KEY_SECRET $AWS_REGION $CLUSTER_NAME
+elif [[ "$CLOUD_PROVIDER" == "do" ]]; then
+  bash ./scripts/setup_do.sh $DO_API_TOKEN $CLUSTER_NAME
+else
```
hanzlamateen, line 19: I just think, the default case can be aws. Like if `CLOUD_PROVIDER` is missing then assume its running for aws. This will ensure all our existing deployments are not broken / does not needs any change.

hanzlamateen (changes requested):

scripts/run-builder.sh
muradkhateeb78: I have updated this as suggested.

packages/server-core/src/media/storageprovider/s3-do.storage.ts
```diff
@@ -0,0 +1,87 @@
...
+
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023
+Ethereal Engine. All Rights Reserved.
+*/
+
+import { FunctionSummary } from '@aws-sdk/client-cloudfront'
+
+import S3Provider from './s3.storage'
+
+const MAX_ITEMS = 1
```
hanzlamateen, line 30: I dont think this is required

packages/server-core/src/media/storageprovider/s3-do.storage.ts
```diff
@@ -0,0 +1,87 @@
...
+const CFFunctionTemplate = `
```
hanzlamateen, line 31: I dont think this is required

hanzlamateen (changes requested):

packages/server-core/src/media/storageprovider/s3.storage.ts
```diff
@@ -159,12 +164,14 @@ export class S3Provider implements StorageProviderInterface {
...
         : config.aws.cloudfront.domain
+      : config.server.storageProvider === 's3-do'
+      ? config.aws.cloudfront.domain
       : `${config.aws.cloudfront.domain}/${this.bucket}`
 
   originURLs = [this.cacheDomain]
 
-  private bucketAssetURL =
-    config.server.storageProvider === 's3'
+  public bucketAssetURL =
```
hanzlamateen, line 173: All the `'s3-do'` specific check in this file should be moved to S3DOStorageProvider. The class variable should be updated based on DO in constructor of it.

hanzlamateen (changes requested):

packages/server-core/src/media/storageprovider/s3-do.storage.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+`
+
+/**
+ * Storage provide class to communicate with AWS S3 API.
+ */
+export class S3DOProvider extends S3Provider {
+  constructor() {
+    super()
+    this.bucketAssetURL = 'https://etherealengine-static-resources.sfo2.digitaloceanspaces.com'
```
muradkhateeb78: Need to remove this.

PR #9407, opened 2023-12-08
Ensured new users do not get avatars missing resources

barankyle:
Added logic to createNewUser to check that the avatar randomly selected for a new user has a valid model and thumbnail resource, and remove that avatar from the list to choose from if either is missing.

## References
closes #_insert number here_


barankyle:
@hanzlamateen I was initially just going to read the existence (or not) of the resources off of the `avatars` object returned from avatarPath.find, but they weren't there. I finally realized that was because the modelResource and thumbnailResource resolvers are on avatarExternalResolver, and this is an internal call. I wasn't sure we wanted to move them to avatarResolver just for this situation; do you know of a way to force a specific resolver to trigger by adding things to the request alone?

Since these resolvers are for external requests, I'd tried adding `provider` to the params to make it look like an external request, but that ran into a different issue, namely that there was no authentication headers since the user doesn't exist yet. Manually GETting the resources like I did doesn't seem like the most elegant solution, but I'm not sure if there's a better way to do it.

HexaField (approved):

hanzlamateen:
> @hanzlamateen I was initially just going to read the existence (or not) of the resources off of the `avatars` object returned from avatarPath.find, but they weren't there. I finally realized that was because the modelResource and thumbnailResource resolvers are on avatarExternalResolver, and this is an internal call. I wasn't sure we wanted to move them to avatarResolver just for this situation; do you know of a way to force a specific resolver to trigger by adding things to the request alone?
> 
> Since these resolvers are for external requests, I'd tried adding `provider` to the params to make it look like an external request, but that ran into a different issue, namely that there was no authentication headers since the user doesn't exist yet. Manually GETting the resources like I did doesn't seem like the most elegant solution, but I'm not sure if there's a better way to do it.

@barankyle `avatarResolver` will be executed even for internal calls. So as in your case these resources can be move to `avatarResolver`.

hanzlamateen (approved):

PR #9414, opened 2023-12-09
add resizing to PNGS and JPEGS on IOS

DanielBelmes:
## Summary
If IOS and JPEG or PNG url then resize to half it's size if the width and height add up to over 1024

TODO: Remove this from the client and add to CMS


HexaField:
Can we clamp it to max of 1024 instead of half it if it is above?

speigg (approved):

PR #9419, opened 2023-12-11
Updated file-browser to update project files locally

barankyle:
## Summary

In local dev, when project files were created/moved/removed, those operations only occurred in the storage provider, not in projects/projects/<project>. For those doing local work on projects and wanting to push from their local file system, this meant that no changes were being recorded.

file-browser operations now, if isDev is true and acting on project files, make the equivalent operation on the local file system as well.

Fixed bugs in the creation of new scenes. One, the operation was trying to copy a default.scene.json from the project being operated on, not the default project. Two, the path of the output files was missing a slash.

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/projects/scene/scene.class.ts
+++ b/packages/server-core/src/projects/scene/scene.class.ts
@@ -49,9 +49,10 @@ import logger from '../../ServerLogger'
 import { getStorageProvider } from '../../media/storageprovider/storageprovider'
 import { ProjectParams } from '../project/project.class'
 import { getSceneData } from './scene-helper'
-const NEW_SCENE_NAME = 'New-Scene'
 
-const sceneAssetFiles = ['.scene.json', '.thumbnail.ktx2', '.envmap.ktx2']
+const DEFAULT_DIRECTORY = 'packages/projects/default-project'
+const NEW_SCENE_NAME = 'New-Scene'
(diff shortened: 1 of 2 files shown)


mrhegemon (approved):

packages/server-core/src/projects/scene/scene.class.ts
```diff
@@ -190,18 +191,18 @@ export class SceneService
...
     }
 
     if (isDev) {
       const projectPathLocal = path.resolve(appRootPath.path, localDirectory)
-      for (const ext of sceneAssetFiles) {
+      for (const ext of SCENE_ASSET_FILES) {
         fs.copyFileSync(
-          path.resolve(appRootPath.path, `${localDirectory}default${ext}`),
-          path.resolve(projectPathLocal + newSceneName + ext)
+          path.resolve(appRootPath.path, `${DEFAULT_DIRECTORY}/default${ext}`),
```
MoizAdnan, line 204: @barankyle, I think `DEFAULT_DIRECTORY` should be used in `getDirectoryFromData` method of `scene.hooks.ts` at line 66. It should be used to set `data.localDirectory`.

MoizAdnan (changes requested):

dinomut1 (approved):

PR #9432, opened 2023-12-12
Force db refresh on builder

### LINKED ISSUE #9404: Feature: QAT Workflow Action
Create a action for QAT cluster that allows us to spin up a PR for QA.

Add: `workflow_dispatch:` to allow us to run the action manually from Actions panel

barankyle:
## Summary

Using a different ENV_VAR than FORCE_DB_REFRESH since the builder configmap starts with the builder's extraEnv, then overwrites it with api/client/instanceserver values. If builder's FORCE_DB_REFRESH is true but the api's is false, then it'll end up being false on the configmap. If we set the api's to true, then the api servers will do their own refresh, which we don't want.

## References
closes #9404


DanielBelmes (approved):

DanielBelmes:
nice

PR #9436, opened 2023-12-13
Fix projects page search

SYBIOTE:
This pull request fixes the search functionality on the projects page. Previously, the search was not filtering the projects correctly. With this fix, the search will now filter the projects based on the entered query.

the query now filters and shows across both, the installed and community projects section, 

do we want to debounce the query to improve performance ?

### DIFF
--- a/packages/editor/src/components/projects/ProjectsPage.tsx
+++ b/packages/editor/src/components/projects/ProjectsPage.tsx
@@ -72,6 +72,7 @@ import { Button, MediumButton } from '../inputs/Button'
 import { CreateProjectDialog } from './CreateProjectDialog'
 import { DeleteDialog } from './DeleteDialog'
 import { EditPermissionsDialog } from './EditPermissionsDialog'
+
 import styles from './styles.module.scss'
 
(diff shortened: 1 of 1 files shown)


speigg (approved):

packages/editor/src/components/projects/ProjectsPage.tsx
```diff
@@ -241,8 +244,8 @@ const ProjectsPage = () => {
           ? CommunityProjectData.filter((p) => p.name.includes(query) || p.description.includes(query))
           : CommunityProjectData
       ).filter((p) => !installedProjects.value?.find((ip) => ip.name.includes(p.name)))
-
       communityProjects.set(data.sort(sortAlphabetical) ?? [])
+      console.log('DEBUG 2', data, communityProjects, filter.value)
```
speigg, line 248: Please delete this log

speigg (changes requested):
Please delete the debug log that was added. Also, yes if you think debouncing would improve the UX, let's do it.

PR #9437, opened 2023-12-13
Fix removeComponent

speigg:
## Summary

Fixes crash when removing components


packages/engine/src/ecs/functions/ComponentFunctions.test.ts
```diff
@@ -239,6 +239,7 @@ describe('ComponentFunctions', async () => {
       removeComponent(entity, TestComponent)
 
       assert.ok(!hasComponent(entity, TestComponent))
+      assert.ok(TestComponent.stateMap[entity]!.value === undefined)
```
HexaField, line 242: can we assert valueMap here too for completeness

HexaField (changes requested):

HexaField (approved):

PR #9456, opened 2023-12-15
fix non-public avatar selection

### LINKED ISSUE #9454: [Bug] User can select another user's private avatar

aditya-mitra:
## Summary

- on avatar removal, only allow public avatars to be re-updated as user's avatar
- also select from public avatars in oauth identity-providers

## References
closes #9454

### DIFF
--- a/packages/server-core/src/user/avatar/avatar.hooks.ts
+++ b/packages/server-core/src/user/avatar/avatar.hooks.ts
@@ -193,15 +193,7 @@ const updateUserAvatars = async (context: HookContext<AvatarService>) => {
       id: {
         $ne: context.id?.toString() as AvatarID
       },
-      $or: [
-        {
-          isPublic: true
-        },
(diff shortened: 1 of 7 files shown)


HexaField (approved):

packages/server-core/src/user/avatar/avatar.hooks.ts
```diff
@@ -192,7 +192,9 @@ const updateUserAvatars = async (context: HookContext<AvatarService>) => {
     query: {
       id: {
         $ne: context.id?.toString() as AvatarID
-      }
+      },
+      isPublic: true,
+      $limit: 1000
     },
     paginate: false
```
hanzlamateen, line 199: @aditya-mitra as we are specifiying limit in the query, then we shouldnt pass paginate:false. Also, the response should be considered as Paginated object instead of array.

PR #9457, opened 2023-12-15
added tests for github related services

aditya-mitra:
## Summary

- added [nock](https://github.com/nock/nock) for mocking [github rest API](https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28) and other endpoints


packages/server-core/src/projects/project/project.test.ts
```diff
@@ -25,89 +25,241 @@ Ethereal Engine. All Rights Reserved.
...
+
+    it('should update the project details if it already exists', async () => {
+      nock('https://api.github.com')
+        .get(/\/user.*/)
+        .reply(200, getAuthenticatedUser('test-user'))
+
+      await app.service(projectPath).update(
+        testProject.id,
+        {
+          sourceURL: sourceDirectory,
```
barankyle, line 186: This is causing errors for me. The sourceURL is normally an https link to a GH repo, not a local file path. In project-helper.ts, in updateProject() around line 1468, we call getAuthenticatedRepo(), which lowercases the sourceURL, adds '.git' to the end, etc. Then, we call gitCloner.clone on that URL. 

For me, the path to sourceDirectory contains uppercase letters (/home/kyle/Development/Forks/etherealengine...), so when it tries to clone the output of getAuthenticatedRepo, the lowercasing of it means that it can't find the repo.

barankyle (changes requested):

barankyle (approved):

PR #9459, opened 2023-12-15
Input Systems Unit Tests

### LINKED ISSUE #9424: [Tests] Spatial Input System
Add unit tests for: ClientInputSystem, ButtonCleanupSystem, InputComponent, InputSourceComponent, InputSourceCaptureState, ButtonCleanupSystem

- Ensure DOM XR & non-XR input populates to InputSourceComponent
- Ensure ClientInputSystem heuristics work for a variety of scenarios
- Ensure input source assignment reactivity works correctly
- Ensure capturing and releasing input sources works correctly

MichaelEstes:
## Summary
Unit tests for InputComponent, InputSourceComponent, ButtonCleanupSystem, ClientInputSystem
## References
closes #9424 

## QA Steps
run `npm run test`


MichaelEstes:
@cla-bot check

DanielBelmes (changes requested):
I think it's a good start but reading over the task I feel as if the specific bullet points weren't completely met. We'll have to emulate the input but I think we can still find a way to test the clientinput system heuristics by simulating some engine bounding boxes and simulating events.

packages/engine/src/input/systems/ClientInputSystem.ts
```diff
@@ -422,8 +422,8 @@ const bboxHitTarget = new Vector3()
 
 const quat = new Quaternion()
 
-const execute = () => {
-  if (!isClient) return null
+const execute = (_isClient = isClient) => {
```
HexaField, line 425: we should use `nock` to mock this, rather than modifying runtime code

HexaField (changes requested):

packages/engine/src/input/systems/ClientInputSystem.ts
DanielBelmes: this library https://www.npmjs.com/package/nock or is there another?

DanielBelmes: I think we do need to change runtime code somewhere to consider a Mocha testing enviornment as it is very similar to if we were just running this code on a server. We should consider testing systems in a server, editor, client, and mobile enviornment. Might need to use babel or something to build a testing version where our device booleans are modified with a replace/

MichaelEstes: I really don't think there's a way to do this without dependency injection or changing isClient's implementation to take in account that the environment is running tests. The exported object that contains isClient is just a view of what's exported, there's no way to change it in any meaningful way that would be reflected when the execute function is called, even with a mocking library.

HexaField: Ok. What we should do here is remove the `isClient` dependency of the `ClientInputSystem` execute call. The server should be able to run all the logic here, escaping if data is not available (eg `Engine.instance.cameraEntity`). Same with the reactor, it should add all the hooks to document/window/canvas if it exists. This way we can easily mock all the funcitonality in our tests, without worrying about these `isClient` exports

packages/engine/tests/engine/input/systems/ClientInputSystem.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    const mockXRFrame = new MockXRFrame()
+    mockXRFrame.pose.transform.position.x = 0.2134
+    mockXRFrame.pose.transform.position.y = 0.3456
+    mockXRFrame.pose.transform.position.z = 0.2345
+    mockXRFrame.pose.transform.orientation.x = 0.4567
+    mockXRFrame.pose.transform.orientation.y = 0.8455
+    mockXRFrame.pose.transform.orientation.z = 0.2454
+    mockXRFrame.pose.transform.orientation.w = 0.2743
+
+    globalThis.ReferenceSpace.origin = new MockXRReferenceSpace()
```
HexaField, line 153: we should be overriding globals in `before` hooks, and cleaning up in `after` hooks

packages/engine/tests/engine/input/systems/ClientInputSystem.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+          }
+        }
+      },
+      destroy: () => {},
+      interactionRays: []
+    } as unknown as WebContainer3D
+    setComponent(entity, InputComponent)
+    setComponent(entity, VisibleComponent)
+    setComponent(entity, InputSourceComponent, { source: mockXRInputSource })
+    const execute = SystemDefinitions.get(ClientInputSystem)!.execute! as (_: boolean) => void
```
HexaField, line 210: revert type

packages/engine/tests/engine/input/systems/ClientInputSystem.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+            y: 1,
+            z: 1
+          }
+        }
+      }
+    } as unknown as World)
+    setComponent(entity, InputComponent)
+    setComponent(entity, VisibleComponent)
+    setComponent(entity, InputSourceComponent, { source: mockXRInputSource })
+    const execute = SystemDefinitions.get(ClientInputSystem)!.execute! as (_: boolean) => void
```
HexaField, line 256: revert type

packages/engine/tests/engine/input/systems/ClientInputSystem.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+      hand: undefined
+    }) as XRInputSource
+
+    const entity = Engine.instance.originEntity
+
+    setComponent(entity, BoundingBoxComponent)
+    setComponent(entity, InputComponent)
+    setComponent(entity, VisibleComponent)
+    setComponent(entity, InputSourceComponent, { source: mockXRInputSource })
+    const execute = SystemDefinitions.get(ClientInputSystem)!.execute! as (_: boolean) => void
```
HexaField, line 281: revert type

packages/engine/tests/engine/input/systems/ClientInputSystem.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    mockXRFrame.pose.transform.orientation.w = 0.2743
+
+    globalThis.ReferenceSpace.origin = new MockXRReferenceSpace()
+    const entity = Engine.instance.originEntity
+    registerState(XRState)
+    getMutableState(XRState).xrFrame.set(mockXRFrame as unknown as XRFrame)
+    setComponent(entity, InputSourceComponent, { source: mockXRInputSource })
+    setComponent(entity, XRSpaceComponent, new MockXRSpace() as XRSpace)
+    setComponent(entity, TransformComponent)
+    const execute = SystemDefinitions.get(ClientInputSystem)!.execute! as (_: boolean) => void
```
HexaField, line 160: revert type

packages/engine/tests/engine/input/systems/ClientInputSystem.test.tsx
```diff
@@ -0,0 +1,292 @@
...
+    loadEmptyScene()
+  })
+
+  it('should add client input listeners', () => {
+    EngineRenderer.instance = new MockEngineRenderer()
+
+    const mockDocEvents = new MockEventListener()
+    const mockWinEvents = new MockEventListener()
+
+    globalThis.document.addEventListener = mockDocEvents.addEventListener as any
```
HexaField, line 74: overrides should be in `before` & `after` hooks

packages/engine/tests/engine/input/systems/ButtonCleanupSystems.test.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ButtonCleanupSystem } from '../../../../src/input/systems/ButtonCleanupSystem'
+import { loadEmptyScene } from '../../../util/loadEmptyScene'
+
+describe('ButtonCleanupSystem', () => {
+  beforeEach(() => {
+    createEngine()
+    loadEmptyScene()
+  })
+
+  it('should be queryable', () => {
```
HexaField, line 41: this test is unnecessary

packages/engine/tests/engine/input/systems/ButtonCleanupSystems.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    assert(buttonCleanupQuery._enterQuery)
+    assert(buttonCleanupQuery._exitQuery)
+    assert(buttonCleanupQuery._query)
+  })
+
+  it('test button cleanup system', () => {
+    const system = SystemDefinitions.get(ButtonCleanupSystem)!
+    const execute = system.execute
+    assert(typeof execute === 'function')
+    assert(execute() === undefined)
```
HexaField, line 53: we don't need to assert the return type of execute, nor the type of the property, just the functionality

packages/engine/tests/engine/input/components/InputSourceComponent.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+import { loadEmptyScene } from '../../../util/loadEmptyScene'
+import { MockXRInputSource, MockXRSpace } from '../../../util/MockXR'
+
+describe('InputSourceComponent', () => {
+  beforeEach(() => {
+    createEngine()
+    loadEmptyScene()
+  })
+
+  it('should be queryable', () => {
```
HexaField, line 45: unnecessary

packages/engine/tests/engine/input/components/InputSourceComponent.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const inputSourceQuery = defineQuery([InputSourceComponent])
+    assert(inputSourceQuery)
+    assert(inputSourceQuery._enterQuery)
+    assert(inputSourceQuery._exitQuery)
+    assert(inputSourceQuery._query)
+  })
+
+  it('should able to be set as a component', () => {
```
HexaField, line 53: we should break this down into multiple tests, one for each situtation/action that is relevant to the component

packages/engine/tests/engine/input/components/InputComponent.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    const json = { highlight: true, grow: true }
+    setComponent(entity, InputComponent, json)
+    const inputComponent = getComponent(entity, InputComponent)
+
+    assert(inputComponent.grow === json.grow)
+    assert(inputComponent.highlight === json.highlight)
+
+    inputComponent.inputSources.push(entity)
+    assert(inputComponent.inputSources.length === 1)
```
HexaField, line 61: this assertion is unnecessary

packages/engine/tests/engine/input/components/InputComponent.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    const Reactor = InputComponent.reactor
+    const { rerender, unmount } = render(<Reactor />)
```
HexaField, line 64: we should actually be asserting that in the variou situtations, highlight &  grow actually add & remove the relevant components, currently this just tests that the reactor doesn't crash

HexaField (changes requested):

packages/engine/tests/engine/input/systems/ClientInputSystem.test.tsx
```diff
@@ -0,0 +1,325 @@
...
+    assert(mockWinEvents.listeners.length === 0, 'Callbacks were removed from window')
+  })
+})
+
+describe('client input system reactor', () => {
+  let originCopy
+
+  before(() => {
+    originCopy = globalThis.ReferenceSpace.origin
+    globalThis.ReferenceSpace.origin = new MockXRReferenceSpace()
```
HexaField, line 150: this one is fine to mock in-line, since it is an engine runtime property

packages/engine/tests/engine/input/systems/ClientInputSystem.test.tsx
```diff
@@ -0,0 +1,325 @@
...
+      hand: undefined
+    }) as XRInputSource
+
+    const entity = Engine.instance.originEntity
+    registerState(XRUIState)
+    getMutableState(XRUIState).interactionRays.set([
+      new Ray(new Vector3(0.23, 0.65, 0.98), new Vector3(0.21, 0.43, 0.82))
+    ])
+    setComponent(entity, XRUIComponent)
+    XRUIComponent.valueMap[entity] = {
```
HexaField, line 227: we should avoid setting valueMap directly, and just pass this object as the property in setComponent

HexaField (changes requested):

HexaField (approved):

PR #9466, opened 2023-12-18
Fixed builder failing due to missing script

hanzlamateen:
## Summary
Fix builder failing as it was unable to find `npm run record-build-error` as its current directory was set to `packages/client` in the same line. This PR sets back the current directory was set to `/app` before executing `record-build-error` script.

## References
closes #_insert number here_


barankyle:
Daniel and I encountered this in #9467, and I think the fix in that one is better. cd'ing back up from packages/client to the root before running record-build-error would potentially mess up the TWA_LINK section, as that also runs a script in the client package. Running everything from root keeps things consistent.

PR #9467, opened 2023-12-18
Improve heap space usage

DanielBelmes:
## Summary

- Small improvements to how icons are loaded.
- Small improvements to how certain scripts are loaded in
  - dynamically load HLS
  - dynamically load rapier(not that useful but should delay when it gets loaded for ios)
- Moved client to ESM instead of CommonJS as Vite is stopping support for commonJS entirely
- Converted scripts in client to ESM
- Added VITE_SOURCEMAPS to configure if sourcemaps should make it into build
- Now use Vite defines to inject env variables into common/config. Allows it to run both as a node script and a vite source
- ~Updated builder node version(we should consider updating the project all to the latest node and drop CommonJS)~
  - We will do this after 1.6 and start to fix any potential errors
- ~Ensure only one ktx2loader exists as its wasm binary is loaded into the heap and takes up a lot of memory on load~
  - need to make some architecture changes in order for this to work
- Update Vite
- deprecate uneeded vite plugins

## References
closes #_insert number here_


dockerfiles/builder/Dockerfile-builder
```diff
@@ -28,7 +28,7 @@ CMD ["scripts/strip-engine.sh"]
 
 ENV APP_ENV=production
 
-FROM node:18-buster-slim as runner
+FROM node:21-bookworm-slim as runner
```
DanielBelmes, line 31: @barankyle @HexaField It does work just fine as far as I can see but there might be some edge cases. Don't know if this is 100% necesarry for my changes though I did see some bugs with `packages/client/scripts/populate-assetlinks.ts` and node18 with my changes

DanielBelmes:
also would like to note this means dev will no longer have sourcemaps. If we want them in dev then there is a env var available. But we definitly don't want them in production like we've been doing.
@barankyle @speigg @HexaField

DanielBelmes:
note packages/client/scripts/populate-assetlinks.ts will be broken for the time being until we want to upload to Google Play Store

HexaField:
> note `packages/client/scripts/populate-assetlinks.ts` will be broken for the time being until we want to upload to Google Play Store

This is probably a deal breaker if it also breaks PWA support for Meta Quest store support

barankyle:
I don't see where we use this 

> > note `packages/client/scripts/populate-assetlinks.ts` will be broken for the time being until we want to upload to Google Play Store
> 
> This is probably a deal breaker if it also breaks PWA support for Meta Quest store support

I don't see that we currently use this anywhere. We don't publish anything to any store from what I can see.

speigg:
> also would like to note this means dev will no longer have sourcemaps. If we want them in dev then there is a env var available. But we definitly don't want them in production like we've been doing. @barankyle @speigg @HexaField

Can we (and do we want to) server source maps in separate files, so they only load when dev tools are open?

speigg (approved):

speigg (changes requested):
Just some failing tests, but otherwise looks good!

DanielBelmes:
> > also would like to note this means dev will no longer have sourcemaps. If we want them in dev then there is a env var available. But we definitly don't want them in production like we've been doing. @barankyle @speigg @HexaField
> 
> Can we (and do we want to) serve source maps in separate files, so they only load when dev tools are open?

That's fairly easy as well. I think that's a good option.

DanielBelmes:
@speigg @HexaField looks like the errors are brought on by a circular dependency between Engine and XUI since I was moving to using assetLoader which lives in the engine currently. It's a decent save not having two ktx2loaders. Any ideas for a solution

DanielBelmes:
@HexaField @speigg All paths to fixing the cyclic dependencies lead us to considering larger architecture changes so I will shelve the single ktx2loader idea for now. We should return to it soon though.

speigg (approved):

PR #9472, opened 2023-12-19
renderOrder to RenderOrderComponent objectLayers to ObjectLayerComponent

### LINKED ISSUE #9327: replace object layers with ObjectLayerComponent

MichaelEstes:
## Summary
Replaces setting renderOrder and objectLayers with RenderOrderComponent and ObjectLayerComponent

### removes
- Engine.instance.objectLayerList

### deprecates

- setObjectLayers()
- enableObjectLayer()

## References
closes #9327, #9328


MichaelEstes:
@cla-bot check

packages/engine/src/scene/components/ObjectLayerComponent.ts
```diff
@@ -0,0 +1,64 @@
...
+    const component = useComponent(entity, ObjectLayerComponent)
+
+    useEffect(() => {
+      const group = getComponent(entity, GroupComponent)
+      for (const object of group) {
+        object.layers.disableAll()
+        for (const layer of component.objectLayers.value) {
+          object.layers.enable(layer)
+        }
+      }
```
DanielBelmes, line 59: In the prior objectLayer implementation we were running a function `updateWorldObjectLayers` before when seting Object layer. I'm not familiar with the purpose myself but I see it missing from this implementation. There was likely a reason why it's there @HexaField

packages/engine/src/scene/components/RenderOrderComponent.ts
```diff
@@ -0,0 +1,61 @@
...
+import { defineComponent, getComponent, useComponent } from '../../ecs/functions/ComponentFunctions'
+import { useEntityContext } from '../../ecs/functions/EntityFunctions'
+import { GroupComponent } from './GroupComponent'
+
+export const RenderOrderComponent = defineComponent({
+  name: 'RenderOrderComponent',
+
+  onInit(entity) {
+    return {
+      renderOrder: 0
```
DanielBelmes, line 37: If you wanted you could probably just set the return as the number value and instead of passing `{renderOrder: 1}` in set component you can just pass `1`

HexaField (dismissed):
The object layers component should be the same type as threejs, a bitmask, rather than an array of values.

packages/engine/src/scene/components/ObjectLayerComponent.ts
speigg: The purpose of that function was to update lists on `Engine.instance.objectLayerList`. My preference is that we deprecate all that, and introduce a tag for each layer. In other words: 

```ts

const ObjectLayerComponents = [defineComponent({name: 'ObjectLayer0'}), ..., defineComponent({name: 'ObjectLayer31'})]

export const ObjectLayerComponent = defineComponent({
   ...,
   setObjectLayers: (entity:Entity, ...layers: number[) => {
     layers.map((i) => setComponent(entity, ObjectLayerComponents[i]);)
   },
   enableObjectLayer: (entity:Entity, layer: number, enable: boolean) {
     layers.map((i) => setComponent(entity, ObjectLayerComponents[i]);)
   }
})


```

HexaField:
```ts
export const ObjectLayerComponent = defineComponent({
   ...,
   entitiesBylayer: Record<layer: number, Entity[]>
})
```

speigg:
> ```ts
> export const ObjectLayerComponent = defineComponent({
>    ...,
>    entitiesBylayer: Record<layer: number, Entity[]>
> })
> ```

Attaching data to components like this is not ideal; this does not get cleaned up automatically between tests. If we want to store state outside of components, we should use `defineState()`

packages/engine/src/scene/components/ObjectLayerComponent.ts
```diff
@@ -0,0 +1,141 @@
...
+      ObjectLayerComponent.mask[entity] &= ~((1 << i) | 0)
+    }
+  })
+})
+
+export const ObjectLayerComponent = defineComponent({
+  name: 'ObjectLayerComponent',
+  schema: { mask: Types.i32 },
+
+  onSet(entity, component) {
```
speigg, line 49: let's add the `mask` here as the input args, and add a toJSON() that returns the mask, so that this component is serialized properly.

packages/engine/src/scene/components/ObjectLayerComponent.ts
```diff
@@ -0,0 +1,141 @@
...
+      ObjectLayerComponent.mask[entity] |= (1 << i) | 0
+    },
+
+    onRemove(entity, component) {
+      ObjectLayerComponent.mask[entity] &= ~((1 << i) | 0)
+    }
+  })
+})
+
+export const ObjectLayerComponent = defineComponent({
```
speigg, line 45: Maybe rename this to `ObjectLayerMaskComponent` for clarity, and to distinguish from the individual layer components?

packages/engine/src/scene/components/RenderOrderComponent.ts
```diff
@@ -0,0 +1,45 @@
...
+import { GroupComponent } from './GroupComponent'
+
+export const RenderOrderComponent = defineComponent({
+  name: 'RenderOrderComponent',
+
+  onInit(entity) {
+    return 0
+  },
+
+  onSet(entity, component, renderOrder) {
```
speigg, line 36: This needs a similar refactor as the object layers, w/ the `renderOrder` property relying on this data as the source of truth

packages/engine/src/scene/components/RenderOrderComponent.ts
```diff
@@ -0,0 +1,45 @@
...
+
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023 
+Ethereal Engine. All Rights Reserved.
+*/
+
+import { defineComponent, getComponent } from '../../ecs/functions/ComponentFunctions'
+import { GroupComponent } from './GroupComponent'
+
+export const RenderOrderComponent = defineComponent({
+  name: 'RenderOrderComponent',
```
speigg, line 30: let's add a bitecs schema for this as well.

speigg (changes requested):

speigg (approved):

PR #9481, opened 2023-12-21
Parallelize texture fetching through workers

### LINKED ISSUE #9439: Fetch textures through worker in KTX2Loader
KTX2Loader fetches textures in main thread and only decodes in worker thread. We have to fetch textures in worker thread instead. fetch in worker fails randomly for some reason. To avoid this, Implement a queuing mechanism at the Loader and throttle the URLs sent to the workers accordingly.

CITIZENDOT:
## References
closes #9439

## QA Steps

Need to test if the performance is improved on Quest and tune `maxConcurrentTasks` accordingly (in `WorkerPool.js`)


speigg (approved):

speigg:
Are there any benchmarks or performance profiling we can do to make sure this change is beneficial?

speigg:
Did some performance profiling, I think this solve it!

PR #9488, opened 2023-12-22
Fix GLTF Export

HexaField:
Reverts EtherealEngine/etherealengine#9487


HexaField:
@dinomut1 had to revert, there seem to be problems with CSM, grabbables, and avatars

PR #9489, opened 2023-12-22
Retargeting Refactor

AidanCaruso:
## Summary

Break up retargeting function into two parts, one for converting to normalized bone space, a second for binding to a given VRM's normalized bones. Expensive retargeting is now only calculated once for every loaded animation, reducing stuttering when loading in new avatars. Also fixes the loop animation component.

## References
closes #_insert number here_


packages/engine/src/avatar/animation/AvatarAnimationGraph.ts
```diff
@@ -151,13 +152,10 @@ export const playAvatarAnimationFromMixamo = (
...
-    retargetedAnimation = retargetMixamoAnimation(
-      cloneDeep(
-        clipName
-          ? animationsScene.animations.find((clip) => clip.name == clipName) ?? animationsScene.animations[0]
-          : animationsScene.animations[0]
-      ),
-      animationsScene,
+    retargetedAnimation = bindAnimationClipFromMixamo(
+      clipName
+        ? animationsScene.animations.find((clip) => clip.name == clipName) ?? animationsScene.animations[0]
```
DanielBelmes, line 157: Suprised the linter isn't yelling for the double equal

DanielBelmes (approved):

DanielBelmes:
QA'd with @speigg and it was looking great. Didn't see any new errors in the console. We tested about 5 avatars. We'll QA as a group later but I think this is a net improvement.

PR #9497, opened 2023-12-25
Fixes for static-resources not updating in file-browser

MoizAdnan:
- Fixes static-resource table not removing entries of file which are inside a folder that is deleted
- Fixes static-resource table not updating paths of file which are inside a folder that is renamed or moved

### DIFF
--- a/packages/server-core/src/media/file-browser/file-browser.class.ts
+++ b/packages/server-core/src/media/file-browser/file-browser.class.ts
@@ -199,14 +199,14 @@ export class FileBrowserService
 
     const staticResources = (await this.app.service(staticResourcePath).find({
       query: {
-        key: { $like: isDirectory ? `%${path.join(_oldPath, data.oldName)}/%` : path.join(_oldPath, data.oldName) }
+        key: { $like: `%${path.join(_oldPath, data.oldName)}%` }
       },
       paginate: false
     })) as StaticResourceType[]
 
     if (staticResources?.length > 0) {
       for (const resource of staticResources) {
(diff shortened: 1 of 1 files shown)


hanzlamateen (approved):

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -316,14 +319,18 @@ export class FileBrowserService
...
-    })) as Paginated<StaticResourceType>
+        key: { $like: `%${key}%` }
+      },
+      paginate: false
+    })) as StaticResourceType[]
 
-    if (staticResource?.data?.length > 0) await this.app.service(staticResourcePath).remove(staticResource?.data[0]?.id)
+    if (staticResources?.length > 0) {
+      for (const resource of staticResources) {
+        await this.app.service(staticResourcePath).remove(resource.id)
```
HexaField, line 331: We should do this in parallel

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -197,21 +197,24 @@ export class FileBrowserService
...
-        $limit: 1
+        key: { $like: `%${path.join(_oldPath, data.oldName)}%` }
+      },
+      paginate: false
+    })) as StaticResourceType[]
+
+    if (staticResources?.length > 0) {
+      for (const resource of staticResources) {
+        const newKey = resource.key.replace(path.join(_oldPath, data.oldName), path.join(_newPath, fileName))
+        await this.app.service(staticResourcePath).patch(
```
HexaField, line 210: Should do this in parallel

HexaField (changes requested):

PR #9499, opened 2023-12-26
#9435 automatic variant configuration

### LINKED ISSUE #9435: Automatic Variant Configuration
We want to automate the process of adding device-targeted LODs to 3D assets. We already have a variant system that allows us to have one object load specific assets for different device types. This should be automated and made part of the GLTF Transform pipeline.
Currently, the process for setting up LODs for a 3D model is as follows:
0. Compress the model into multiple LODs using manual settings
1. Add a Variant Component to the model:
image
2. Change the variant heuristic to Device:
image
3. Add multiple variant levels to the component:
image
4. For each level, assign a device type and drag and drop a 3d model URL from the file browser into the src input:
image
5. To save this into a .gltf file, the object must now be wrapped within an additional Model Component and exported as its own file:

We should, as part of the batched compression workflow starting with #9402, add an option to generate an additional .gltf or .glb file that has these variants automatically configured. Steps 1-5 should be done procedurally with the resulting LODs from the batched compression, and the result should be saved into the same folder with the suffix `-device-LODs`. 
- Use the ee-tutorial-bubbles implementation as a reference to see how ECS data can be created with code. 
- At the end, use the ExportRelativeGLTF function to export + upload the prefab to the server

Rezmason:
## Summary

We want to automate the process of adding device-targeted LODs to 3D assets. We already have a variant system that allows us to have one object load specific assets for different device types. This should be automated and made part of the GLTF Transform pipeline.

## References
closes #9435

## QA Steps
TBD


dinomut1 (changes requested):
Can we change this to run as part of the Batched Compression pass in the Compression panel? Let's remove the button from the top bar and add a toggle for "Create LOD Prefab" in the Compression panel instead.

dinomut1 (approved):
great job

PR #9509, opened 2023-12-27
Fix Loop Animations for Non-VRM Sourced Rigs

AidanCaruso:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/avatar/functions/updateVRMRetargeting.ts
+++ b/packages/engine/src/avatar/functions/updateVRMRetargeting.ts
@@ -24,12 +24,13 @@ Ethereal Engine. All Rights Reserved.
 */
 
 import { VRM, VRMHumanBoneList } from '@pixiv/three-vrm'
-import { Object3D, Quaternion, Vector3 } from 'three'
+import { Matrix4, Object3D, Quaternion, Vector3 } from 'three'
 import { getComponent, getOptionalComponent } from '../../ecs/functions/ComponentFunctions'
 import { EntityTreeComponent } from '../../ecs/functions/EntityTree'
 import { TransformComponent } from '../../transform/components/TransformComponent'
 import { BoneComponent } from '../components/BoneComponent'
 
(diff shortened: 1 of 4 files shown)


packages/engine/src/avatar/functions/updateVRMRetargeting.ts
```diff
@@ -55,23 +49,10 @@ export const updateVRMRetargeting = (vrm: VRM, deltaTime: number) => {
...
-
-        const parentBoneNode = getComponent(parentEntity, BoneComponent)
-        if (!parentBoneNode) continue
-
-        computeTransformMatrix(parentEntity)
-        parentBoneNode.updateMatrixWorld(true)
-
-        const parentWorldMatrix = parentBoneNode.matrixWorld
-        const localPosition = boneWorldPosition.applyMatrix4(parentWorldMatrix.clone().invert())
+        boneNode.parent!.updateWorldMatrix(true, false)
```
HexaField, line 53: why are we reverting from ECS language to threejs?

The two clone() calls are also going to be unnecessary and expensive

HexaField (changes requested):

packages/engine/src/avatar/functions/updateVRMRetargeting.ts
AidanCaruso: Removed one of the clone calls, but I can't find a way around cloning the matrix world right now.

HexaField: module scope
`const mat4 = new Matrix4()`

function scope
`const localPosition = boneWorldPosition.applyMatrix4(mat4.copy(parentWorldMatrix).invert())`

AidanCaruso: Thanks. Don't know why I didn't realize that 🤦

HexaField (approved):

Issue #9484, opened 2023-12-21
Materials as Standalone Assets


dinomut1:
Currently we have a crude way of saving materials into their own .gltf files. We need to expand this so that those files are identifiable as a Material file, distinct from a normal .gltf file.

A 'Save' button exists in the Material Library Panel that will save the selected material into a .gltf file. This should be modified so that the files are saved as ".material.gltf" files.

The 'Save' Button should open a modal which let's the user name that material, and then save it into a specified folder in a project, similar to how the ModelNodeEditor let's users export model gltf files.

We should then add a "Material" asset type + class to the AssetLoader functions getAssetType and getAssetClass, paving the way for custom drag and drop logic for these files.

PR #9517, opened 2023-12-28
add save path and drag material on object

JT00y:
Materials as Standalone Assets
follow the issues 
https://github.com/EtherealEngine/etherealengine/issues/9484

We want to make material asset distinct from a normal .gltf with drag and drop them onto object. In this way we can easily manage and maintain materials on dev.


dinomut1 (approved):
great job

PR #9521, opened 2023-12-29
Align ObjectLayerComponents array with mask

HexaField:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/ObjectLayerComponent.ts
+++ b/packages/engine/src/scene/components/ObjectLayerComponent.ts
@@ -34,16 +34,17 @@ import {
 } from '../../ecs/functions/ComponentFunctions'
 
 const maxBitWidth = 32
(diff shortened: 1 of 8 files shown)


packages/engine/src/scene/components/ObjectLayerComponent.ts
```diff
@@ -34,16 +34,16 @@ import {
 } from '../../ecs/functions/ComponentFunctions'
 
 const maxBitWidth = 32
-export const ObjectLayerComponents = Array.from({ length: maxBitWidth }, (_, i) => {
+export const ObjectLayerComponents = Array.from({ length: maxBitWidth + 1 }, (_, i) => {
```
HexaField, line 37: should this be + 1 on the array length? is 33 a valid mask?

MichaelEstes (approved):

PR #9531, opened 2023-12-30
Disable Basic Materials In Scene Render Settings

dinomut1:


HexaField:
What exactly is the reason for add this to scenes?

PR #9540, opened 2024-01-01
VRM Autoconversion Logic

AidanCaruso:
## Summary
Adds an override to the model component that allows the user to choose whether an avatar is autoconverted to VRM, that is only toggleable via the node editor if the model is determined to be a non-vrm avatar with a recognized mixamo naming scheme. Also cleans up the autoconversion function, and fixes animations not populating into the animation component.

https://github.com/EtherealEngine/etherealengine/assets/102248647/b2b5f706-6e9e-4368-b485-0b70a1dab77f

## References
closes #_insert number here_


dinomut1 (approved):
great job

PR #9555, opened 2024-01-04
fix controlled string input

SYBIOTE:
controlled string input did not have appropiate props, they have been fixed and now its workign properly and propagating changes up the chain properly


DanielBelmes (approved):

packages/editor/src/components/inputs/StringInput.tsx
```diff
@@ -33,16 +33,27 @@ const inputStyle = {
   margin: 0
 }
 
-interface StyledNumericInputProps {
+interface StyledStringInputProps {
   className?: string
-  onChange?: any
+  onChange?: (e: any) => void
```
speigg, line 38: Is there a reason this needs to be `any`? I'd suggest we stick to stricter types.

speigg (changes requested):

packages/editor/src/components/inputs/StringInput.tsx
SYBIOTE: the e value?

speigg: yes

HexaField (approved):

packages/editor/src/components/inputs/StringInput.tsx
SYBIOTE: fixed

SYBIOTE:
@speigg , there is an issue, if we make it strictly typed, there are a lot of higher-level components using the props assuming they were any and they have been interpreted accordingly, do we do ahead and change all the higher-level components as well to fit the stricter type or we change the lower-level string input component with functions to convert the strict react type to string

SYBIOTE:
the main issue is the props being used as base for other props

HexaField:
I belive you functionality you want is what already exists in `StringInput`. `ControlledStringInput` is specific for fields that have dramatic side effects for any value changes at all (such as the ModelComponent loading a scene).

PR #9556, opened 2024-01-04
add component and state property listeners in behave graph 

SYBIOTE:
add component and state property listener nodes which can react to changes to component and state properties in the engine


packages/editor/src/components/inputs/StringInput.tsx
```diff
@@ -33,16 +33,27 @@ const inputStyle = {
   margin: 0
 }
 
-interface StyledNumericInputProps {
+interface StyledStringInputProps {
   className?: string
-  onChange?: any
+  onChange?: (e: any) => void
```
speigg, line 38: Please do not type these as `any`, this leads to bugs.

speigg (changes requested):

packages/engine/src/behave-graph/nodes/Profiles/Engine/helper/stateHelper.ts
```diff
@@ -125,3 +130,88 @@ export function getStateGetters() {
...
+      skipped.push(stateName)
+      continue
+    }
+    const outputsockets = generateStateNodeSchema(state)
+    if (Object.keys(outputsockets).length === 0) {
+      skipped.push(stateName)
+      continue
+    }
+    const node = makeEventNodeDefinition({
+      typeName: `engine/component/on ${stateName} change`,
```
speigg, line 156: Do we have to make a separate event node for each component? Can we do a single `engine/onComponentChange` node?

packages/engine/src/behave-graph/nodes/Profiles/Engine/helper/componentHelper.ts
```diff
@@ -148,3 +157,89 @@ export function getComponentGetters() {
...
+          uuid: `behave-graph-on-${componentName}-change` + uniqueId(),
+          insert: { with: InputSystemGroup },
+          execute: () => {
+            const componentValue = getComponent(entity, component)
+            if (typeof componentValue !== 'object') {
+              if (prevComponentValue === componentValue) return
+              const value = EnginetoNodetype(componentValue)
+              write(valueOutputs[valueOutputs.length - 1][0] as any, value)
+              commit(flowOutputs[flowOutputs.length - 1][0] as any)
+              prevComponentValue = structuredClone(componentValue)
```
speigg, line 215: Doing it like this is going to be extremely inefficient. Please use a reactor.

speigg (changes requested):

packages/engine/src/behave-graph/nodes/Profiles/Engine/helper/componentHelper.ts
SYBIOTE: i dont think u can embed a reactor in a behave graph node definition?
i know we can define systems as we need, can we do the same for reactors as well?

packages/engine/src/behave-graph/nodes/Profiles/Engine/helper/stateHelper.ts
SYBIOTE: i suppose, but then we have to dynamically change the handles each time, i suppose its possible now, ill do it

packages/editor/src/components/inputs/StringInput.tsx
SYBIOTE: the e value?

packages/engine/src/behave-graph/nodes/Profiles/Engine/helper/componentHelper.ts
SYBIOTE: all rewritten with reactor based logic, turns out you can define UseEffects in a loop as well

packages/editor/src/components/inputs/StringInput.tsx
SYBIOTE: we need this to be any, atleast for the stringInputProps as it is used as a base for other props

packages/engine/src/behave-graph/nodes/Profiles/Engine/helper/stateHelper.ts
SYBIOTE: got now a single node will require too much hardcoded elements

packages/engine/src/behave-graph/nodes/Profiles/Engine/Values/EntityNodes.ts
```diff
@@ -250,6 +268,75 @@ export const setEntityTransform = makeFlowNodeDefinition({
...
+  label: 'Get self entity',
+  in: {},
+  out: { entity: 'entity' },
+  exec: ({ read, write, graph: { getDependency } }) => {
+    const useEntityReactor: any = defineSystem({
+      uuid: 'behave-graph-getEntityReactor-' + uniqueId(),
+      insert: { after: PresentationSystemGroup },
+      execute: () => {},
+      reactor: function () {
+        const entity = useEntityContext()
```
the-reality-engineer: does this actually work? 

The entity context will only exist inside of a component reactor, or if a parent reactor has defined that context. 

Can we write tests for these behave-graph nodes so that we know they work correctly?

speigg (changes requested):

packages/engine/src/behave-graph/nodes/Profiles/Engine/helper/componentHelper.ts
```diff
@@ -148,3 +157,89 @@ export function getComponentGetters() {
...
+      continue
+    }
+    const outputsockets = generateComponentNodeSchema(component, true)
+    if (Object.keys(outputsockets).length === 0) {
+      skipped.push(componentName)
+      continue
+    }
+
+    const node = makeEventNodeDefinition({
+      typeName: `engine/component/on ${componentName} change`,
```
the-reality-engineer: We are mixing conventions here. Can we stick to camelCase please?

packages/engine/src/behave-graph/nodes/Profiles/Engine/helper/componentHelper.ts
the-reality-engineer: I'm still not seeing the reactor logic here

speigg (changes requested):

speigg (approved):

PR #9561, opened 2024-01-05
Mesh BVH and visualizer

### LINKED ISSUE #9463: MeshBVHVisualizer should be tied to root entities via `useMeshOrModel`
contingent on https://github.com/EtherealEngine/etherealengine/pull/9369

MichaelEstes:
## Summary
Creates MeshBVHComponent that handles generating and visualizing MeshBVH

## References
closes #9463


packages/engine/src/scene/functions/loadGLTFModel.ts
```diff
@@ -301,6 +302,9 @@ export const generateEntityJsonFromObject = (rootEntity: Entity, obj: Object3D,
   if (skinnedMesh.isSkinnedMesh) setComponent(objEntity, SkinnedMeshComponent, skinnedMesh)
   else setComponent(objEntity, FrustumCullCameraComponent)
 
+  if (mesh.isMesh && !instancedMesh.isInstancedMesh && !skinnedMesh.isSkinnedMesh)
+    setComponent(objEntity, MeshBVHComponent, { mesh: mesh })
```
HexaField: this is going to add a BVH for every object in the model's heirarchy. we don't want that, just one for each model (or lone mesh)

we need to use useMeshOrModel for the visualizer

HexaField (changes requested):

HexaField (approved):

PR #9563, opened 2024-01-05
improve file browser views

aditya-mitra:
## Summary

- added scrollbar to rc-dock window (scrollbars were not visible because they were hidden in all docks)
- added uniform font across file browser (now remains same when accessing admin or studio)
- added icon size setting when icon view is active to change the size of files and folders (setting persisted to local storage)
- added showing of selected of table columns in list view of files and folders (setting persisted to local storage)

## References
closes #_insert number here_

## QA Steps

list view table column selection

image

icon view icon size selection

image

https://github.com/EtherealEngine/etherealengine/assets/55396651/5efe740f-2d06-420a-9c6e-374535516c3e


dinomut1 (changes requested):
The settings window is getting closed and reopened every time a value on it is changed. This loses focus, so the slider won't work.
Also, let's put the size slider on list view as well to change the size of the list rows

aditya-mitra:
Here is the changing of font size in list view!

https://github.com/EtherealEngine/etherealengine/assets/55396651/68f82aef-b6cf-4558-bb18-3ad56b2b6e34

dinomut1 (approved):
great job

PR #9565, opened 2024-01-05
Got validators out of client

DanielBelmes:
## Summary
Moved schemas to common. Redid imports. Created a type schema file. Duplicated paths into module. 

## References
closes #9523


DanielBelmes:
@HexaField @hanzlamateen I am grossly sleep deprived because of a animal pet emergency I had last night into this morning. This currently gets the validators out of the client. `packages/common/src/schema.type.module.ts` currently duplicates the path. I tested thoroughly and sadly it won't tree shake out the getValidator calls if I import from each of the schema files. The paths have to be separate files. This works fine. Though we really should make each schema file and each server-core reference a single path var.

DanielBelmes:
Dang something is broken with the git build vs my local. Probably my package.json changes. Working on it

HexaField:
> Dang something is broken with the git build vs my local. Probably my package.json changes. Working on it

The deps in the root are fine as they were before. Feathers/feathers is required there at the least

DanielBelmes:
> > Dang something is broken with the git build vs my local. Probably my package.json changes. Working on it
> 
> The deps in the root are fine as they were before. Feathers/feathers is required there at the least

yea I am reverting

PR #9567, opened 2024-01-05
Pass BaseURI correctly to the worker

CITIZENDOT:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/assets/loaders/gltf/KTX2Loader.js
+++ b/packages/engine/src/assets/loaders/gltf/KTX2Loader.js
@@ -129,7 +129,7 @@ class KTX2Loader extends Loader {
 
 		}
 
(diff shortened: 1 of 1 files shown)


speigg (approved):

speigg (approved):

speigg:
@CITIZENDOT please resolve the conflicts

CITIZENDOT:
@speigg  KTX2Loader is [reverted to older version](https://github.com/EtherealEngine/etherealengine/pull/9568) since the parallel loader is causing locations stuck.

When I was looking into the issue, I submitted this (current) PR, but this still does not solve the problem. And, current PR is based on parallel loader, which we're not using now.

CC: @HexaField

PR #9569, opened 2024-01-07
IK Hint Fixes

HexaField:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/avatar/functions/avatarFootHeuristics.ts
+++ b/packages/engine/src/avatar/functions/avatarFootHeuristics.ts
@@ -53,19 +53,20 @@ const speedMultiplier = 2
 const footRotationOffset = new Quaternion()
   .setFromAxisAngle(V_100, Math.PI / 2)
   .multiply(new Quaternion().setFromAxisAngle(V_010, Math.PI))
-  .multiply(new Quaternion().setFromAxisAngle(V_100, -Math.PI / 6))
 
 //step threshold should be a function of leg length
 //walk threshold to determine when to move the feet back into standing position, should be
 export const setIkFootTarget = (localClientEntity: Entity, delta: number) => {
   const userID = getComponent(localClientEntity, UUIDComponent)
 
-  const ikTargetLeftFoot = UUIDComponent.getEntityByUUID((userID + ikTargets.leftFoot) as EntityUUID)
-  const ikTargetRightFoot = UUIDComponent.getEntityByUUID((userID + ikTargets.rightFoot) as EntityUUID)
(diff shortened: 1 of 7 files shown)


packages/engine/src/avatar/systems/AvatarAnimationSystem.ts
```diff
@@ -267,16 +284,26 @@ const execute = () => {
     if (rightFootTargetBlendWeight) {
       _quat2.copy(rightFootTransform.rotation)
       if (isAvatarFlipped) {
-        _quat2.multiply(rightFootRotation)
+        _quat2.multiply(Q_X_90)
+      } else {
+        /** I don't know why flipped models don't need this foot angle applied, nor why it needs to be halved */
```
AidanCaruso, line 289: This is likely because our flipped models happen to only ever be VRM, it would seem raw bones on VRM also use identity quaternions in bind pose, though I'm not sure why *that* is.

packages/engine/src/avatar/functions/avatarFootHeuristics.ts
```diff
@@ -53,39 +53,39 @@ const speedMultiplier = 2
...
-  const ikTargetLeftFoot = UUIDComponent.getEntityByUUID((userID + ikTargets.leftFoot) as EntityUUID)
-  const ikTargetRightFoot = UUIDComponent.getEntityByUUID((userID + ikTargets.rightFoot) as EntityUUID)
-  if (!ikTargetLeftFoot || !ikTargetRightFoot) return
+  const leftFootEntity = UUIDComponent.getEntityByUUID((userID + ikTargets.leftFoot) as EntityUUID)
+  const rightFootEntity = UUIDComponent.getEntityByUUID((userID + ikTargets.rightFoot) as EntityUUID)
 
-  const leftFootTargetBlendWeight = AvatarIKTargetComponent.blendWeight[ikTargetLeftFoot]
-  const rightFootTargetBlendWeight = AvatarIKTargetComponent.blendWeight[ikTargetRightFoot]
+  if (!leftFootEntity || !rightFootEntity) return
+  AvatarIKTargetComponent.blendWeight[leftFootEntity] = 1
```
AidanCaruso, line 66: I assume these were left in from testing. These will cause the ik solve to always run regardless of being in an immersive session.

AidanCaruso (changes requested):

packages/engine/src/avatar/functions/avatarFootHeuristics.ts
HexaField: oh whoops, yes

AidanCaruso (approved):

Issue #9508, opened 2023-12-27
LOD Levels Parameters Data Structure


dinomut1:
Create a data structure to represent LOD levels in hookstate. Each LOD level needs to contain a ModelTransformParameters object, as well as an optional device target. 

Adjust the implementation of the compression panel so that, when batched compression is enabled, instead of operating on a hardcoded set of 3 compression parameters, it operates on an arbitrarily large list of compression parameters. Change the compression settings so that they represent/edit the compression parameters of a specific element in the list. 

The LOD Levels data structure should be stored in local storage. Each file should have its LOD Levels settings saved individually.

Issue #9507, opened 2023-12-27
LOD List


dinomut1:
## LOD List in Compression Panel ##

![image](https://github.com/EtherealEngine/etherealengine/assets/94419856/582df81e-f51c-4055-b179-14e272281b57)

Create a list UI element matching the above wireframe, where the LODs are ordered as clickable elements. The selected list element should be highlighted. There should be an "Add LOD Level" button at the bottom of the list which adds elements. Each element should have an 'x' button on their right side which removes the LOD level if clicked.

The compression settings in the center of the screen should represent the selected LOD level. Changing the settings should update the selected LOD level's parameters.

speigg:
This looks great!

speigg:
Question: does it make sense to think of these optimized assets as having a static linear progression (e.g. LOD 0, LOD 1 ... etc.) when there might be completely different optimization strategies for different circumstances (desktop vs mobile vs PC VR vs mobile XR) ? Wouldn't different circumstances warrant ordering these asset variants in different ways, depending on the resources being optimized (e.g., CPU, GPU, texture memory, etc) and the heuristics chosen?

dinomut1:
> Question: does it make sense to think of these optimized assets as having a static linear progression (e.g. LOD 0, LOD 1 ... etc.) when there might be completely different optimization strategies for different circumstances (desktop vs mobile vs PC VR vs mobile XR) ? Wouldn't different circumstances warrant ordering these asset variants in different ways, depending on the resources being optimized (e.g., CPU, GPU, texture memory, etc) and the heuristics chosen ?
> 
> Perhaps a simple collection of optimization targets is more appropriate, and then the ordering of these can be handled at runtime, based on the optimization heuristics that are chosen (also at runtime) ?

Yes we should make this more multidimensional, as a next step. To start we just need a 1-D list of progressively more optimized assets, with certain elements on that list tagged as device targets. Distance LODs will be constructed by selecting a range of LOD levels starting at the tagged level.
This can be expanded upon to have different orderings, and multiple LOD lists using different optimization strategies depending on the target device.

PR #9583, opened 2024-01-09
Model compression panel with LOD support

### LINKED ISSUE #9506: LOD Presets
## LOD Presets in Compression Panel
brave_susN8L3R6h

Create a preset list UI element in the Compression panel that represents a list of ModelTransformParameters presets. Selecting a preset copies its values into the currently selected LOD level. The selected preset should be highlighted. If the user manually changes any of the ModelTransformParameters settings after selecting a preset, that preset is no longer highlighted.

Add a "Create Preset" button at the bottom of the list which makes a new preset out of the current settings. Created presets should be saved into the user's local storage.

Rezmason:
## Summary

Model compression and image compression are now in separate panels, and the model compression panel now has a list of LODs on the left and a list of presets on the right.

A GLTF is created for every LOD, and if the user is exporting a combined model, then a variant level is generated for every variant metadata blob on every LOD.

## References
closes #9506, #9507, #9508


Rezmason:
TODO:

- Handle case where a preset already exists for the current transform parms

dinomut1 (changes requested):
When opening the compression panel, there should be three LOD levels preconfigured by default to target desktop, mobile and xr

dinomut1 (approved):
great jaerb

PR #9590, opened 2024-01-09
Deprecate directional light useInCSM and add "primary light" option in scene renderer setting

### LINKED ISSUE #9580: Deprecate directional light useInCSM and add "primary light" option in scene renderer setting

MichaelEstes:
## References
closes #9580


packages/editor/src/components/properties/RenderSettingsEditor.tsx
```diff
@@ -103,6 +106,19 @@ const ShadowTypeOptions = [
   }
 ]
 
+const DirectionalLightOptions = () => {
+  const directionalLightOptions = [] as Array<{ label: string; value: string; info?: string }>
+
+  for (const entity of DirectionalLightQuery()) {
```
HexaField, line 112: we should use useQuery here instead of a standard query

packages/engine/src/renderer/WebGLRendererSystem.ts
```diff
@@ -285,6 +285,7 @@ export class EngineRenderer {
 export const RenderSettingsState = defineState({
   name: 'RenderSettingsState',
   initial: {
+    primaryLight: '',
```
HexaField, line 288: should be typed as EntityUUID

HexaField (changes requested):

DanielBelmes:
Would love to get this in if you can fix the conflicts when you're available.

HexaField:
> Would love to get this in if you can fix the conflicts when you're available.

This will break shadows for all our scenes, so we'll need to do it at a time that makes sense.

Otherwise, we could add a migration utility.

DanielBelmes:
> > Would love to get this in if you can fix the conflicts when you're available.
> 
> This will break shadows for all our scenes, so we'll need to do it at a time that makes sense.
> 
> Otherwise, we could add a migration utility.

Migration utility would be nice. It could live in the /studio page and it can parse all scene json in a project and apply a specific migration? We're gonna need something like a "recipe." It needs to detect the discrepancy and then find a solution for it. 

We can also do a migration commit and hope people resave their scenes and remove that logic at a later release. 

We can also detect the issue and instruct in the studio to fix it.

HexaField:
@MichaelEstes @DanielBelmes let's get this PR updated with the latest dev, i think we can easily add a migration script:

in `SceneState.loadScene` where our other migrations are, we should be able to very easily find the first directional light in the scene that has useInCSM enabled, and set the `primaryLight` value to that entity

HexaField (approved):

PR #9605, opened 2024-01-11
Cleanup component functions

the-reality-engineer:
## Summary

Reduce some of the complexity in component functions


packages/engine/src/ecs/functions/EntityTree.ts
```diff
@@ -60,7 +60,7 @@ export const EntityTreeComponent = defineComponent({
   onInit: (entity) => {
     return {
       // api
-      parentEntity: null as Entity | null,
+      parentEntity: UndefinedEntity as Entity,
```
HexaField: `as Entity` redundant

HexaField (changes requested):

HexaField (approved):

PR #9631, opened 2024-01-17
Added publish locations in studio

### LINKED ISSUE #9333: Publish Locations from Studio

MoizAdnan:
Adds `Publish` button in studio toolbar, which opens `LocationDrawer` and lets user publish location for the active scene if they have the permission

closes #9333

### DIFF
--- a/packages/client-core/src/admin/components/Location/LocationDrawer.tsx
+++ b/packages/client-core/src/admin/components/Location/LocationDrawer.tsx
@@ -56,6 +56,7 @@ interface Props {
   mode: LocationDrawerMode
   selectedLocation?: LocationType
   onClose: () => void
+  selectedScene?: SceneID | null
 }
 
 const defaultState = {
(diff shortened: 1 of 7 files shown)


packages/client-core/src/admin/components/Location/LocationDrawer.tsx
```diff
@@ -91,12 +92,24 @@ const LocationDrawer = ({ open, mode, selectedLocation, onClose }: Props) => {
...
-  const sceneMenu: InputMenuItem[] = scenes.get(NO_PROXY).map((el) => {
-    return {
-      value: `${el.project}/${el.name}`,
-      label: `${el.name} (${el.project})`
-    }
-  })
+  const sceneName = selectedScene ? selectedScene.split('/')[1] : ''
+  const projectName = selectedScene ? selectedScene.split('/', 1)[0] : ''
+
+  if (selectedScene) state.scene.set(selectedScene)
```
HexaField: we shouldn't be setting state in the body of a react component. it should be in a useEffect listening for changes to selectedScene

HexaField (changes requested):
just one small fix, otherwise this is fantastic!

HexaField (approved):

PR #9632, opened 2024-01-17
Resource State

### LINKED ISSUE #9535: Reference Counting for disposable threejs objects

### LINKED ISSUE #9629: Create ResourceManagerState
Part of Resource Manager Epic [#9613](https://github.com/EtherealEngine/etherealengine/issues/9613)

### LINKED ISSUE #9674: Update MeshBVH Package Version
### Version

0f27df2(1.5.0)

### Operating Systems

- [X] macOS
- [X] Windows
- [X] Linux
- [X] Android
- [X] IOS

### Browser (if applicable)

_No response_

look at: https://github.com/gkjohnson/three-mesh-bvh

### What happened?

Not applicable

### Steps to reproduce

```shell
Use editor
```

### Relevant Info

_No response_

MichaelEstes:
## Summary
Resource Manager and State
## References
closes #9629 
closes https://github.com/EtherealEngine/etherealengine/issues/9535
closes https://github.com/EtherealEngine/etherealengine/issues/9674


packages/engine/src/assets/state/ResourceState.ts
```diff
@@ -0,0 +1,150 @@
...
+  metadata: {
+    size?: number
+  }
+  onGPU: boolean
+}
+
+export const ResourceState = defineState({
+  name: 'ResourceManagerState',
+  initial: () => ({
+    resources: {} as Record<string, Resource>
```
the-reality-engineer: Let's maintain a separate record for each type of resource so that we can hold different metadata for different kinds of things.

packages/engine/src/assets/functions/resourceHooks.ts
```diff
@@ -0,0 +1,156 @@
...
+      (request) => {
+        if (!unmounted) progress.set(request)
+      },
+      (err) => {
+        if (!unmounted) error.set(err)
+      }
+    )
+
+    return () => {
+      unmounted = true
```
the-reality-engineer: We should use an AbortController to cancel fetching resources here

packages/engine/src/assets/state/ResourceState.ts
```diff
@@ -0,0 +1,246 @@
...
+  const resources = resourceState.nested('resources')
+  if (!resources[url].value) {
+    console.error('ResourceManager:unload No resource exists for url: ' + url)
+    return
+  }
+
+  const resource = resources[url]
+
+  resource.references.set((entities) => {
+    const index = entities.indexOf(entity)
```
the-reality-engineer, line 368: We should be performing reference counting here; a single entity may use the same resource more than once.

the-reality-engineer (changes requested):

packages/engine/src/assets/functions/resourceHooks.ts
```diff
@@ -31,6 +31,16 @@ import { LoadingArgs } from '../classes/AssetLoader'
 import { GLTF } from '../loaders/gltf/GLTFLoader'
 import { AssetType, ResourceManager, ResourceType } from '../state/ResourceState'
 
+function getAbortController(url: string, callback: () => void): AbortController {
```
HexaField: this should be called `createAbortController` since it creates one

packages/engine/src/assets/state/ResourceState.ts
MichaelEstes: I added this in, but then removed it because some hooks from Three js are helpful in determining the loading state, but they only pass the url of the asset. Having different metadata types for each asset type is still implemented, but through a discriminated union approach instead of separate records.

DanielBelmes:
Have put up on QAT to smoke test. Looking good!

DanielBelmes (approved):

HexaField (approved):

PR #9634, opened 2024-01-17
Postproccesing improvements

DanielBelmes:
## References
closes #_insert number here_


packages/engine/src/renderer/functions/configureEffectComposer.ts
```diff
@@ -81,14 +85,20 @@ export const configureEffectComposer = (
...
 
-  const smaaEffect = new SMAAEffect()
+  const smaaEffect = new SMAAEffect({
+    preset: SMAAPreset.HIGH,
+    edgeDetectionMode: EdgeDetectionMode.COLOR
+  })
   composer.SMAAEffect = smaaEffect
 
   const outlineEffect = new OutlineEffect(scene, camera, getState(HighlightState))
+  outlineEffect.selectionLayer = ObjectLayers.HighlightEffect
```
DanielBelmes, line 95: Had to change the layer of the outline because the default is 10 and we use that for transform Gizmo. @JT00y you'll have to likely do the same when using outline effect

packages/engine/src/scene/components/PostProcessingComponent.tsx
```diff
@@ -45,12 +44,8 @@ export const PostProcessingComponent = defineComponent({
...
     if (!json) return
 
-    if (json.enabled) component.enabled.set(json.enabled)
-    if (json.effects) {
-      for (const [name, effect] of Object.entries(json.effects)) {
-        component.effects[name].merge(effect)
-      }
-    }
+    if (typeof json.enabled === 'boolean') component.enabled.set(json.enabled)
+    if (typeof json.effects === 'object') component.effects.set(json.effects)
```
HexaField: we need to deeply clone any objects to avoid mutating original scene data. the merge previously is also to avoid removing effects that are not part of the current saved schema

packages/engine/src/scene/components/PostProcessingComponent.tsx
```diff
@@ -62,52 +57,17 @@ export const PostProcessingComponent = defineComponent({
...
-              value={value}
-            />
-          ))
-        )
-        .flat()}
-    </>
-  )
-}
+  useEffect(() => {
+    getMutableState(PostProcessingSettingsState).effects.set(component.effects.get(NO_PROXY_STEALTH))
```
HexaField: similar to above, we need a separate reactor for each effect, and do a merge, or we may be losing data

HexaField (dismissed):

the-reality-engineer (approved):

PR #9639, opened 2024-01-18
Set `MediaComponent` properly

CITIZENDOT:
## References
closes #_insert number here_


HexaField (changes requested):
This needs to be done in the studio editor nodes instead of the components themselves

HexaField (approved):

PR #9642, opened 2024-01-18
Fix Highlight effect and object layer bug

DanielBelmes:
## Summary
The logic behind setting object layer was incorrect and causing the highlight effect logic to not work properly.
Furthermore the default layer for the highlight effect is 10 which clashes with the transform gizmo. I have moved it to 11

## References
closes #_insert number here_


DanielBelmes:
Merging in for Demo in 30min

PR #9653, opened 2024-01-22
Scene Hookstate Encapsulation

dinomut1:
To fix circular hookstate reactor loops in the scene loading system, each scene is defined as a separate hookstate object


mrhegemon (changes requested):
not needed

PR #9656, opened 2024-01-23
replace file urls with configured file server

### LINKED ISSUE #9624: Non cached asset links sometimes populate into scene files
Screenshot from 2024-01-16 11-33-18
Links to the s3 bucket directly instead of cloudfront links are coming from somewhere. Either the file browser or another flow in the studio

aditya-mitra:
## Summary

Instead of using direct storageprovider links, this PR replaces them with the `config.client.fileServer`

## References
closes #9624

### DIFF
--- a/packages/server-core/src/media/file-browser/file-browser.class.ts
+++ b/packages/server-core/src/media/file-browser/file-browser.class.ts
@@ -30,7 +30,7 @@ import path from 'path/posix'
 
 import { processFileName } from '@etherealengine/common/src/utils/processFileName'
 
-import config, { isDev } from '@etherealengine/common/src/config'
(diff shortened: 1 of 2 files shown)


packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -126,6 +126,9 @@ export class FileBrowserService
     const total = result.length
 
     result = result.slice(skip, skip + limit)
+    result.forEach((file) => {
+      file.url = `${config.client.fileServer}/${file.key}`
```
HexaField: we should use `getCachedURL` here

HexaField (changes requested):

HexaField (approved):

PR #9657, opened 2024-01-23
Revamp Tailwind Theming

HexaField:


packages/client-core/src/common/services/ThemeService.tsx
```diff
@@ -0,0 +1,93 @@
...
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023 
+Ethereal Engine. All Rights Reserved.
+*/
+
+import { useHookstate } from '@hookstate/core'
+import { useEffect } from 'react'
+
+import { defineState, getMutableState } from '@etherealengine/hyperflux'
+
+const lightTheme = {
```
aditya-mitra, line 31: Can we add these variables to our main css file?
https://stackoverflow.com/a/66333605

HexaField: yeah we will do that eventually, it's not a priority until we have finished replacing all the MUI routes

PR #9664, opened 2024-01-24
Material Asset Fixes

dinomut1:
Fixes around saving materials, editing materials, drag and drop materials.


packages/engine/src/scene/functions/loadGLTFModel.ts
```diff
@@ -201,11 +201,21 @@ export const proxifyParentChildRelationships = (obj: Object3D) => {
...
+        if (hasComponent(objEntity, EntityTreeComponent)) {
+          const childEntities = getComponent(objEntity, EntityTreeComponent).children.filter((child) =>
+            hasComponent(child, SceneObjectComponent)
+          )
+          const result: Object3D[] = []
+          for (const childEntity of childEntities) {
+            if (!hasComponent(childEntity, GroupComponent)) continue
+            const objs = getComponent(childEntity, GroupComponent)
+            if (objs.length === 0) continue
+            result.push(objs[0])
```
HexaField: this is going to lead to problems as some entities have more than one object in the group

HexaField (changes requested):

packages/engine/src/scene/functions/loadGLTFModel.ts
dinomut1: Yes but now that everything is one entity per Object3D, this shouldn't lead to any missed elements in the hierarchy, except for helpers and ancillary objects (such as lights)

dinomut1: Is there a situation somewhere in the engine where we're using the proxified "children" property to grab multiple object3Ds from the same entity?

HexaField: There are many cases where multiple objects exist on the same entity, particularly with helpers and other complex logic. I previously added a warning log in the group component to warn of multiple objects and that that pattern is going to be deprecated, but that warning was removed. We need to ensure there is a one to one relationship always, if we are going to be accessing things one to one.

dinomut1: Yes but all of those helpers are still parented to the scene itself (so they still render, as can be seen in this branch), and their transform logic is still handled by the ECS. They shouldn't be detected in the scene graph unless you're explicitly looking for them using the entity tree rather than threejs children property.

dinomut1: Again, are there any situations where we're using the proxified "children" property to grab these helpers or extra objects, instead of directly querying the group component through the ECS?
At this point, it seems the only time we're really utilizing the proxified "children" property is when we're doing gltf exports, or grabbing the first element of the group component (like for animations) just as is happening in this code

PR #9665, opened 2024-01-25
add last login to user table

### LINKED ISSUE #9503: In user table there should be a last login column
We should maintain a last login field in user table and populate it whenever a user accessed the EE. This can be done with the help of authenticate hook.

aditya-mitra:
## Summary

- added `lastLogin` column to the user table by migration
- added last login column to the admin users table

## References
closes #9503 

## QA Steps

image


packages/server-core/src/hooks/authenticate.ts
```diff
@@ -124,3 +125,7 @@ const checkWhitelist = (context: HookContext<Application>): boolean => {
 
   return false
 }
+
+const addLastLogin = (context: HookContext<Application>) => {
```
hanzlamateen: This method needs to be async and method call should be await

hanzlamateen (changes requested):

packages/server-core/src/hooks/authenticate.ts
```diff
@@ -88,6 +89,7 @@ export default async (context: HookContext<Application>, next: NextFunction): Pr
       const user = await context.app.service(userPath).get(key.data[0].userId)
       context.params.user = user
       asyncLocalStorage.enterWith({ user })
+      addLastLogin(context)
```
hanzlamateen: This needs to be await as well

hanzlamateen (changes requested):

hanzlamateen (approved):

PR #9666, opened 2024-01-25
Move ECS to it's own package

HexaField:
## References
closes #_insert number here_


packages/engine/package.json
```diff
@@ -34,6 +34,7 @@
     "@behave-graph/struct": "https://gitpkg.now.sh/etherealengine/behave-graph/packages/struct?dev",
     "@dimforge/rapier3d-compat": "0.11.2",
     "@etherealengine/common": "^1.6.0",
+    "@etherealengine/engine": "^1.6.0",
```
DanielBelmes: Is this not cyclic? Shouldn't we use relative imports if we're in the package

HexaField: Oh whoops, I meant to make this the ecs package

PR #9667, opened 2024-01-25
Fixed renamed scene not updating in publish location drawer

MoizAdnan:


packages/server-core/src/projects/scene/scene.class.ts
```diff
@@ -245,6 +246,16 @@ export class SceneService
       }
     }
 
+    const location = (await this.app.service(locationPath).find({
+      query: { $limit: 1, sceneId: { $like: `${directory}${oldSceneName}.scene.json` as SceneID } }
+    })) as Paginated<LocationType>
+
+    if (location.data.length > 0) {
+      await this.app
```
HexaField: we need to be doing this for every location returned, not just the first

HexaField (changes requested):

HexaField (approved):

Issue #9512, opened 2023-12-28
Thumbnail Handling


dinomut1:
- Create a client-side "generateThumbnail" function which takes a GLTF model, image or video, and generates a thumbnail either through rendering or downsampling

- Add logic to File Browser Panel and Assets Panel where, when static resource queries occur, every response object that doesn't have a thumbnail defined is added to a thumbnail generation queue

- Add an extra `thumbnail` column in static-resources table

- Create a feathers service `upload-thumbnail`, which uses a Frankenstein monster combination of the File Browser and Static Resources to upload an image to the `thumbnails` folder of a project and then associate that thumbnail with an asset in the static resources. The service should consist of a single function which accepts a blob, representing the thumbnail image, and a static resource key, which points to the static resource that this thumbnail should be associated with. 
1. The blob is uploaded to the server. The server gleans the target project from where the static resource is. If that project does not have a `thumbnails` folder, create it
2. Save the thumbnail image into the `thumbnails` folder.
3. If the target static resource already has a thumbnail, delete it
4. Associate the static resource with the thumbnail through the `thumbnail` column in the target static resource

- Add a hook onto the deletion of static resources which deletes the associated thumbnail, but only if the thumbnail resides in a project's `thumbnails` folder

Rezmason:
Dev notes

```
Make some kind of Feathers service
	Rerender stuff
		Models: default lighting/camera situation
		Images
		Videos
		128x128

Static resources
	All assets are stored in there
	Every asset any project has are in there
	addendum
		Thumbnail URL
		Handle bookkeeping
		Seems like most of the time people just have a solution where they generate a bunch of JPGs
		Some folder in the backend with a bunch of generated thumbnails
			Periodically updated, deleted/regenerated
				Kinda like a filesystem event
			Separate files

Ask Kyle and some other backend people
	because Dave doesn't know the exact way to render these server side
	On our deployments, we don't have graphics cards
		And there's this special thing we do where on our servers we don't load any textures or anything in-memory
			because there's no GPU
	Seems like we'll have to render on the client and upload to the server
```

barankyle:
Crash course in feathers and the backend in general:

We have a schema file for a service in packages/common/src/schemas/<schema-type>/<service-name>.schema.ts. e.g., For the backend service file-browser-upload, which is located at packages/server-core/src/media/file-browser-upload, its schema file is at packages/common/src/schemas/media/file-browser-upload.ts

[...]

PR #9672, opened 2024-01-26
#9512 file thumbnails

Rezmason:
## Summary
Files that are visual should visualize what they file! 
Whenever an eligible file has no thumbnail, the client should make one and register it with an endpoint on the server. 
The server should save the thumbnail in the static resources with a unique ID and associate it with the corresponding asset.

## References
closes #_insert number here_


packages/editor/src/components/assets/FileBrowser/FileIcon.tsx
```diff
@@ -0,0 +1,199 @@
...
+  if (blob == null) {
+    return ''
+  }
+
+  const maybe = uploadToFeathersService(
+    fileThumbnailPath,
+    [new File([blob], file.key)],
+    {
+      assetKey: file.key,
+      isCustom: false,
```
HexaField: where is isCustom set to true? is this redundant?

PR #9697, opened 2024-01-31
Add script to automatically install EE on Ubuntu for localdev

heysokam:
## Summary
Adds a script to automatically install EE on Ubuntu for localdev.  
The fileoverview doc-comment at the top of the file explains what its goals are and how it works.  

The script can be downloaded and run directly from a clean Ubuntu LTS installation with the oneliner command:
```bash
wget https://github.com/path/to/this/file.sh && bash -i file.sh
```
After this, the user will only need to run `npm run dev` and the engine will launch.

This should solve the pain of installing Node+Docker+Dependencies as a first-time user for:
- User that just wants to casually try the engine
- User being onboarded for the first time
- Quickly setting up the engine for development/contribution on a clean Ubuntu LTS install.

## QA Steps
- [x] Test the script on Ubuntu LTS 22.04 inside a VM
  _The line `systemctl start --user docker-desktop` failed during tests due to KVM not running inside Ubuntu+Virtualbox, but the rest worked well._
- [x] Test the script in real hardware.
  _Worked without issues._
- [x] Possibly missing `python` and `mediasoup` dependencies.
  _Worked well without explicitly installing anything else_


heysokam:
~~Possibly missing `python` or `mediasoup` dependencies.~~
edit: Worked without any other explicit package installations.

~~Definitely needs to be tested on real hardware before merging, as the KVM<->VirtualboxVM conflict might be hiding potential bugs behind it.~~
edit: Tested and worked as expected. The issues were caused by the VM.

heysokam:
@DanielBelmes @HexaField The PR is now tested on hardware and works correctly.
The issues I found before were just because of running inside a VM.

I updated the instructions to remove some redundancy, and added an explanation on the `-i` flag decision and what causes it.
Let me know if you have some other QA steps you can think of. But should be ready to merge otherwise :writing_hand:

DanielBelmes (approved):

DanielBelmes:
We can QA but onboarding people with it until we have Control Center ready. Great work. Thank you

PR #9702, opened 2024-02-01
new tailwind components for button, text & input

aditya-mitra:
## Summary

- removed `backgroundColor`
- added variants
- improved button icons to only use `ReactNode` type
- removed usage of defaultProps (used prop default assignment instead)
- improved other styling
- added tailwind `Text` component
- added tailwind `TextInput` component

## References
closes #_insert number here_

### button component

image

image

image

---

### text component

image

image

---

 ### text input component

image

image


CITIZENDOT:
- SVG icon should scale with button size IMO.
- These designs doesn't have user interaction (on hover, on click etc..), If this is what we want, Looks good to me.

HexaField (approved):

PR #9709, opened 2024-02-03
Updated media settings and fixed bugs.

### LINKED ISSUE #9675: Set mediasoup codecs, bitrates, etc. from environment variables or settings
Currently, pretty much all mediasoup settings are hard-coded. They should be set from environment variables and/or instanceserver settings so that they can be optimized for various deployment situations.

barankyle:
## Summary

Added new client-settings for audio, video, and screenshare. For audio, admins can set the max bitrate for opus (also hard-coded DTX to be on). For video and screenshare, admins can select VP8, VP9, or h24 codec, for video the maximum resolution that the user's webcam will get, and for video and screenshare using simulcast (i.e. not VP9), the maximum bitrates for the high/mid/low-end streams.

Fixed a bug where a lack of consumers would crash the instanceservers since the transport would be closed. The correct check is if all producers and consumers have been closed; seemed like a typo from a recent conversion.

Fixed a bug in some useEffect listeners that were inadvertently closing all consumers except the most recently created one.

## References
closes #9675


.github/workflows/test-deploy.yml
```diff
@@ -0,0 +1,91 @@
+name: test-deploy
+
+on:
+  push:
+    branches:
+      [9675-mediasoup-config]
```
HexaField: guessing this script needs to be removed

HexaField (changes requested):

.github/workflows/test-deploy.yml
DanielBelmes, line 1: This action yaml is locked to this branch. Is it still needed? Or do you mean to move to dev?

.github/workflows/test-deploy.yml
barankyle: Yes, it's gone.

packages/client-core/src/transports/SocketWebRTCClientFunctions.ts
```diff
@@ -1157,17 +1193,43 @@ export const startScreenshare = async (network: SocketWebRTCClientNetwork) => {
...
+  const clientSettingState = getState(AdminClientSettingsState).client[0]
+  const screenshareSettings = clientSettingState.mediaSettings.screenshare
   const currentChannelInstanceConnection = channelConnectionState.instances[network.id]
   const channelId = currentChannelInstanceConnection.channelId
 
   await waitForTransports(network)
   const transport = MediasoupTransportState.getTransport(network.id, 'send') as WebRTCTransportExtension
 
+  let codec, encodings
+  if (screenshareSettings) {
```
DanielBelmes: Since it's similar to the video settings above we could clean this up slightly by making it a function that returns [codec, encodings].

DanielBelmes (approved):
Looks good. I trust your judgement on the state usages and mediasoup. I probably can't catch any errors in that regard(mediasoup config). There was some small code nitpicks. I didn't go too crazy. I'm assuming from the prior github action you have QA'd on deployment and local.

barankyle:
Note for myself, there now seems to be a bug in dev that's not fixed in this PR where media that's already running doesn't get consumers on new clients.

HexaField (approved):

PR #9710, opened 2024-02-03
Fix material placement to correctly handle registered materials

dinomut1:
slightly modifies material placement logic to correctly handle placing the same material onto multiple objects in a scene


JT00y (commented):
it fixed the problem when apply same material to multiple object the material change not update on all the object.

JT00y (approved):
it fixed the problem when apply same material to multiple object the material change not update on all the object.

PR #9716, opened 2024-02-04
Video Texture support for UVOL

CITIZENDOT:
## References
closes #_insert number here_


packages/engine/src/scene/constants/UVOLTypes.ts
```diff
@@ -208,7 +208,7 @@ export interface GeometryInput {
   targets: Record<string, GLBTarget | DRACOTarget | UniformSolveTarget>
 }
 
-export type TextureFormat = 'ktx2' | 'astc/ktx2'
+export type TextureFormat = 'ktx2' | 'astc/ktx2' | 'vp9'
```
the-reality-engineer: Does the fact that the video is encoded in `vp9` actually matter? Do we only support vp9? Couldn't we also support vp8 or h264 or h265 without any other changes to our code?

the-reality-engineer (changes requested):

packages/engine/src/scene/constants/UVOLTypes.ts
CITIZENDOT: > Does the fact that the video is encoded in vp9 actually matter

It doesn't actually. Any video which is playable by `HTMLVideoElement` works.

the-reality-engineer: Then we shouldn't use `vp9` as the texture format. The texture format should be `video`.

the-reality-engineer: Looks like you did that already, thanks!

the-reality-engineer (approved):

PR #9722, opened 2024-02-06
Fix Camera colliding with invisible collider, Reset Zoom level

DanielBelmes:
## Summary
Was seeing issues on sky station with the camera beeing very zoomed in.  I also noticed when maxDistance was smaller than zoom the zoom value wasn't being reset.

**Note:**
I did try some improvements to the camera raycasting to improve how it clips into walls and I did see improvements but there were some edge cases when the avatar got too close to a wall. I can still work on that for this PR i'm sure there is some check I can do for the radius off the normal offset I was applying.

## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/MeshBVHComponent.ts
+++ b/packages/engine/src/scene/components/MeshBVHComponent.ts
@@ -104,7 +104,7 @@ export const MeshBVHComponent = defineComponent({
                 if (toGenerate == 0) {
                   component.generated.set(true)
                 }
-                if (model.cameraOcclusion) {
+                if (model.cameraOcclusion && mesh?.visible) {
                   ObjectLayerMaskComponent.enableLayers(currEntity, ObjectLayers.Camera)
                 }
               }
--- a/packages/engine/src/scene/components/MeshBVHComponent.ts
+++ b/packages/engine/src/scene/components/MeshBVHComponent.ts
@@ -104,7 +104,7 @@ export const MeshBVHComponent = defineComponent({
                 if (toGenerate == 0) {
                   component.generated.set(true)
                 }
-                if (model.cameraOcclusion && mesh?.visible) {
+                if (model.cameraOcclusion) {
                   ObjectLayerMaskComponent.enableLayers(currEntity, ObjectLayers.Camera)
                 }
               }
--- a/packages/spatial/src/camera/systems/CameraSystem.tsx
+++ b/packages/spatial/src/camera/systems/CameraSystem.tsx
@@ -148,9 +148,7 @@ export const getMaxCamDistance = (cameraEntity: Entity, target: Vector3) => {
   raycaster.firstHitOnly = true // three-mesh-bvh setting
   raycaster.far = followCamera.maxDistance
   raycaster.set(target, targetToCamVec.normalize())
-  const hits = raycaster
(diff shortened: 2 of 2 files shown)


packages/engine/src/scene/components/MeshBVHComponent.ts
```diff
@@ -104,7 +104,7 @@ export const MeshBVHComponent = defineComponent({
                 if (toGenerate == 0) {
                   component.generated.set(true)
                 }
-                if (model.cameraOcclusion) {
+                if (model.cameraOcclusion && mesh?.visible) {
```
DanielBelmes: weird. There is already a visible.value check above. Investigating

HexaField: this should be done at runtime anyway, or it might get desynced

maybe we just need to add VisibleComponent to the query for the camera intersection?

DanielBelmes: Alright I went ahead and did that with a fancy filter since hits usually had a couple values

packages/spatial/src/camera/systems/CameraSystem.tsx
```diff
@@ -147,7 +148,9 @@ export const getMaxCamDistance = (cameraEntity: Entity, target: Vector3) => {
   raycaster.firstHitOnly = true // three-mesh-bvh setting
   raycaster.far = followCamera.maxDistance
   raycaster.set(target, targetToCamVec.normalize())
-  const hits = raycaster.intersectObjects(sceneObjects, false)
+  const hits = raycaster
+    .intersectObjects(sceneObjects, false)
+    .filter((item) => getOptionalComponent(item.object.entity, VisibleComponent) === true)
```
HexaField: this filter is unnecessary, add VisibleComponent to cameraLayerQuery

packages/spatial/src/camera/systems/CameraSystem.tsx
```diff
@@ -156,8 +159,9 @@ export const getMaxCamDistance = (cameraEntity: Entity, target: Vector3) => {
   //Check the cone for minimum distance
   cameraRays.forEach((rayDir, i) => {
     raycaster.set(target, rayDir)
-    const hits = raycaster.intersectObjects(sceneObjects, false)
-
+    const hits = raycaster
+      .intersectObjects(sceneObjects, false)
+      .filter((item) => getOptionalComponent(item.object.entity, VisibleComponent) === true)
```
HexaField: unnecessary here too

HexaField (changes requested):

packages/spatial/src/camera/systems/CameraSystem.tsx
DanielBelmes: I am blind :) thank you

DanielBelmes: but it is cool

DanielBelmes: done

DanielBelmes: or not

DanielBelmes: now done

HexaField (approved):

PR #9737, opened 2024-02-09
Performance profiling

### LINKED ISSUE #9735: Performance profiling
- Create budgets from exposed GPU info in WebGL
- Create budgets heuristic in variant component
- Build metadata in variant component for asset level resources for each LOD during asset compression step in editor
- Track loaded assets in resource manager against budgets
- Query variants on asset load/unload in resource manager to adjust budget heuristic for variant components

MichaelEstes:
## Summary
Run a series of checks on startup to build a performance profile

## References
closes #9735


packages/spatial/src/renderer/functions/configureEffectComposer.ts
```diff
@@ -84,8 +85,32 @@ export const configureEffectComposer = (
 
   const effects: any[] = []
 
+  const performanceTier = getState(PerformanceState).tier
+
+  let smaaPreset
+  switch (performanceTier) {
```
HexaField: this is nice, we should see if we can also update this property dynamically in a reactor without recreating the entire postprocessing pipeline

HexaField (approved):

PR #9742, opened 2024-02-10
Synchronous event sourcing with new runner

HexaField:
## References
closes #_insert number here_


packages/ecs/src/SystemFunctions.ts
```diff
@@ -50,6 +50,7 @@ export interface SystemArgs {
   insert: InsertSystem
   timeStep?: number | 'variable'
   execute?: () => void
+  runner?: () => void
```
the-reality-engineer: Why can't we just use the `execute` function? It seems confusing to be introducing yet another separate execution model into systems.

PR #9744, opened 2024-02-12
Optimize builder job to reduce its run time. 

muradkhateeb78:
## Summary

The purpose is to optimize the code such that the builder job takes little less time while building the code  changes.

### DIFF
--- a/scripts/build_docker_builder.sh
+++ b/scripts/build_docker_builder.sh
@@ -24,9 +24,9 @@ docker buildx build \
     --load \
     --build-arg NODE_ENV=$NODE_ENV \
     -t $ECR_URL/$REPO_NAME-root:${TAG} \
-    -t $DOCR_REGISTRY/$REPO_NAME-root:latest_$STAGE \
+    -t $ECR_URL/$REPO_NAME-root:latest_$STAGE \
     -f dockerfiles/root/Dockerfile-root .
-docker push --all-tags $DOCR_REGISTRY/$REPO_NAME-root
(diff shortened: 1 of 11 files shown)


scripts/build_docker_builder.sh
```diff
@@ -15,11 +15,26 @@ if [ $PRIVATE_ECR == "true" ]
...
+cp packages/projects/default-project/package.json ./project-package-jsons/projects/default-project
+find packages/projects/projects/ -name package.json -exec bash -c 'mkdir -p ./project-package-jsons/$(dirname $1) && cp $1 ./project-package-jsons/$(dirname $1)' - '{}' \;
+
+docker buildx build \
+    --load \
+    --build-arg NODE_ENV=$NODE_ENV \
+    -t $ECR_URL/$REPO_NAME-root:${TAG} \
+    -t $ECR_URL/$REPO_NAME-root:latest_$STAGE \
+    -f dockerfiles/root/Dockerfile-root .
+docker push --all-tags $ECR_URL/$REPO_NAME-root
```
barankyle: We should be able to just add the --push option to docker buildx build, and use --cache-to and --cache-from of type gha, like in the other buildx build commands in this file.

barankyle (changes requested):

scripts/build_docker_builder.sh
muradkhateeb78: This has been changed accordingly.

barankyle (approved):

PR #9746, opened 2024-02-12
temporary disable of material preview panel

DanielBelmes:
## Summary
Disable MaterialPreview Panel for now until we can handle it not mutating the scene.


DanielBelmes:
@AidanCaruso disabled the lights in the preview and this also handles this error.

PR #9777, opened 2024-02-19
fix for EEECSImporter

HexaField:
## References
closes #_insert number here_


HexaField:
better fix with backwards compat merged elsewhere

PR #9791, opened 2024-02-23
Multi domain token usage

barankyle:
## References
closes #_insert number here_


packages/client/package.json
```diff
@@ -13,15 +13,16 @@
   "npmClient": "npm",
   "scripts": {
     "check-errors": "tsc --noemit",
-    "dev": "cross-env APP_ENV=development NODE_OPTIONS=--max_old_space_size=20480 vite",
-    "build": "cross-env NODE_OPTIONS=--max_old_space_size=10240 vite build",
+    "dev": "npm run update-root-cookie-accessor-domain && cross-env APP_ENV=development NODE_OPTIONS=--max_old_space_size=20480 vite",
```
HexaField: we should include this script inside the vite.config.ts instead of polluting the commands here

packages/ecs/src/Engine.ts
```diff
@@ -49,7 +57,7 @@ export class Engine {
     const UndefinedEntity = bitECS.addEntity(HyperFlux.store)
   }
 
-  api: FeathersApplication<ServiceTypes>
+  api: FeathersClient
```
HexaField: we shouldn't overload this here, since we access it on the backend and it will have different types

HexaField: we should be able to declare a type override in client-core that will only be applied in packages that import client-core

HexaField (changes requested):

packages/ecs/src/Engine.ts
barankyle: Weird, I made this change because of type errors when I started, but maybe something has changed in dev since then, because now it's not showing any type mismatches when I revert to what it was.

packages/server-core/src/createApp.ts
```diff
@@ -97,7 +97,11 @@ export const configurePrimus =
     const origin = [
       'https://' + appConfig.server.clientHost,
       'capacitor://' + appConfig.server.clientHost,
-      'ionic://' + appConfig.server.clientHost
+      'ionic://' + appConfig.server.clientHost,
+      'https://cool.pants.com',
```
HexaField: Left over test domains? We should probably remove capacitor & ionic stuff too

HexaField (changes requested):

packages/server-core/src/networking/allowed-domains/allowed-domains.class.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export class AllowedDomainsService implements ServiceInterface<Boolean | BadRequest | undefined, Params> {
+  app: Application
+
+  constructor(app: Application) {
+    this.app = app
+  }
+
+  async find(params?: Params) {
+    const domainToCheck = params?.query.domainToCheck
```
hanzlamateen: I think this implementation should be done in hooks to allow overrides by project. i.e. packages/server-core/src/projects/scene/scene.class.ts

hanzlamateen (changes requested):

barankyle:
Note to myself, something about the cookie seems to mess up OAuth login on the root domain if it becomes disconnected from a database user. Locally, I'd reinited the database several times while the cookie was still on localhost:3000, and it broke all OAuth providers. The `grant` session didn't seem to persist across the entire OAuth flow.

PR #9796, opened 2024-02-26
Rework Client Input Sourcing

the-reality-engineer:
## Summary

- [x] Rename `cameraEntity` to `viewerEntity` and `originEntity` to `localFloorEntity`
- [x] Separate the various kinds of input events into clearer use hooks in `ClientInputSystem`
- [x] Adds `InputPointerComponent` to store 2D input data per input source, which is then translated to the `InputSourceComponent`, which stores 3D input data per input source
- [x] adds `RendererComponent` which binds a renderer to an entity
- [x] Update relevant controls (e.g., FlyCameraControl)
- [x] Update allocation of Input Sources on InputComponent

## References

https://github.com/EtherealEngine/etherealengine/issues/9893


packages/ecs/src/Engine.ts
```diff
@@ -77,16 +77,19 @@ export class Engine {
...
 
   /**
    * The xr origin group
+   * TODO: remove this
    */
   origin = new Group()
 
   /**
    * The camera entity
+   * TODO: rename this to viewerEntity (align w/ WebXR naming)
```
HexaField: we can add getters that point to this.cameraEntity and deprecate cameraEntityy

packages/spatial/src/input/components/InputSourceComponent.tsx
```diff
@@ -40,7 +41,10 @@ import {
 import { Entity, UndefinedEntity } from '@etherealengine/ecs/src/Entity'
 import { useEntityContext } from '@etherealengine/ecs/src/EntityFunctions'
 import { defineQuery } from '@etherealengine/ecs/src/QueryFunctions'
-import { XRSpaceComponent } from '../../xr/XRComponents'
+import { Raycaster } from 'three'
+import { TransformComponent } from '../../SpatialModule'
```
HexaField: we careful not to import from a module file like this from within the same module

HexaField (changes requested):

HexaField (approved):

PR #9806, opened 2024-02-28
Scene Snapshot and State Refactor

HexaField:
## Summary

Splits Scene state into two states:
- SceneState tracks metadata, extraneous state for scene loaded, modified etc
- SceneSnapshotState tracks event sourced snapshots, upon changing a snapshot applies it to the SceneState source of truth
- removes activeScene, specifying it explicitly based on context (studio sceneID or current location sceneID)

## References
closes #_insert number here_


packages/engine/src/scene/functions/createSceneEntity.ts
dinomut1, line 1: We shouldn't delete this function, it's a very useful helper

dinomut1 (changes requested):

packages/editor/src/components/assets/ModelCompressionPanel.tsx
dinomut1, line 1: We shouldn't be defining a createTempEntity function within this UI panel file. CreateSceneEntity is a more flexible general use helper function that was sufficient here and has many potential uses elsewhere in the codebase

dinomut1 (changes requested):

packages/editor/src/components/assets/ModelCompressionPanel.tsx
HexaField: we don't have another use case for this yet. we can already do effectively the same thing by passing json and an empty object into generateEntityFromEntityJSON

dinomut1: This is a very useful function for proceduralism in general. There is no reason to remove it and it is not part of the scope of this PR anyways

PR #9807, opened 2024-02-28
Avatar transparency dithering

AidanCaruso:
## References
closes #_insert number here_


packages/engine/src/avatar/components/TransparencyDitheringComponent.ts
```diff
@@ -54,14 +53,16 @@ export const TransparencyDitheringComponent = defineComponent({
...
+      ditheringExponent: 5,
+      position: new Vector3()
     }
   },
 
   onSet: (entity, component, json) => {
     if (!json) return
     if (matches.number.test(json.ditheringDistance)) component.ditheringDistance.set(json.ditheringDistance)
     if (matches.number.test(json.ditheringExponent)) component.ditheringExponent.set(json.ditheringExponent)
+    if (matchesVector3.test(json.position)) component.position.set(json.position)
```
HexaField: we should not be copying vector3 by reference, instead by value, and updating that value in a system. if we don't do this, we run the risk of mutating the upstream data accidentally

HexaField: let's also call this `center`

HexaField (changes requested):

packages/engine/src/avatar/components/TransparencyDitheringComponent.ts
AidanCaruso: > we should not be copying vector3 by reference, instead by value, and updating that value in a system. if we don't do this, we run the risk of mutating the upstream data accidentally

Agree, I do not intend on keeping the current reference to the camera's position in the component.

packages/engine/src/avatar/components/TransparencyDitheringComponent.ts
```diff
@@ -54,15 +54,15 @@ export const TransparencyDitheringComponent = defineComponent({
...
+      center: new Vector3()
     }
   },
 
   onSet: (entity, component, json) => {
     if (!json) return
     if (matches.number.test(json.ditheringDistance)) component.ditheringDistance.set(json.ditheringDistance)
     if (matches.number.test(json.ditheringExponent)) component.ditheringExponent.set(json.ditheringExponent)
-    if (matchesVector3.test(json.position)) component.position.set(json.position)
+    if (matchesVector3.test(json.center)) component.center.value.copy(json.center)
```
HexaField: for proper reactivity, we need to use `component.center.set(new Vector3().copy(json.center))`

HexaField (changes requested):

packages/engine/src/avatar/functions/avatarFunctions.ts
```diff
@@ -170,6 +171,8 @@ export const setupAvatarProportions = (entity: Entity, vrm: VRM) => {
  * successfully loaded.
  */
 export const setupAvatarForUser = (entity: Entity, model: VRM) => {
+  if (entity == Engine.instance.localClientEntity) setComponent(entity, TransparencyDitheringComponent)
```
HexaField, line 174: i think we should add this to all avatars, would help with other being trying to block your vision in VR

HexaField (changes requested):

packages/engine/src/avatar/functions/avatarFunctions.ts
AidanCaruso: I considered this but ultimately didn't add it to all avatars because I thought it would be a bit too intrusive. We can give it a shot though

HexaField (approved):

PR #9809, opened 2024-02-28
Revert "Update CSM shader (#9798)"

DanielBelmes:
This reverts commit 94ce5fc9753cb8f4c27cc122846112164decd94f.

I will investigate how to fix the regression soon. But it's best that we just restore the CSM to the less broken state it was in

## References
closes #_insert number here_


DanielBelmes:
Already done and I missed it

PR #9821, opened 2024-03-01
Tailwind Admin User

aditya-mitra:
## Summary

- added admin users in tailwind

## References
closes #_insert number here_


packages/client-core/src/admin2/components/user/AddEditUserModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+      }
+    }
+
+    if (!avatarOptions.find((av) => av.value === user.avatarId)) {
+      avatarOptions.push({ name: user.avatar.name || user.avatarId, value: user.avatarId })
+    }
+  }
+
+  const submitLoading = useHookstate(false)
+  const errors = useHookstate(getDefaultErrors())
```
CITIZENDOT, line 94: we can probably use the value directly instead of a function. are we using this function elsewhere?

packages/client-core/src/admin2/components/user/AddEditUserModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+      avatarId: avatarId.value as AvatarID,
+      isGuest: user?.isGuest,
+      scopes: scopes.value
+    }
+    submitLoading.set(true)
+    if (user?.id) {
+      await userMutation.patch(user.id, userData).then(() => PopoverState.hidePopupover())
+    } else {
+      await userMutation.create(userData).then(() => PopoverState.hidePopupover())
+    }
```
CITIZENDOT: - Why are we using both async/await and "then" at the same time?
- In case of errors, we should catch them, and show it in the modal without closing.

CITIZENDOT (changes requested):

packages/client-core/src/admin2/components/user/AddEditUserModal.tsx
aditya-mitra: corrected!

packages/client-core/src/admin2/components/user/AddEditUserModal.tsx
aditya-mitra: Had added it for resetting/clearing the errors when form is resubmitted. Hookstate takes the reference of the object when resetting it.

this does not work:
```ts
errors.set(defaultErrors);
```
but this does:
```ts
errors.set(getDefaultErrors());
```
because the `getDefaultErrors()` function creates a new object each time.

PR #9826, opened 2024-03-04
Added redirect back to studio on non-accessable projects

### LINKED ISSUE #9443: [Bug]: Studio shows projects user doesn't have access to, users can access studio window with those projects (though they don't load)
### Version

0f27df2(1.5.0)

### Operating Systems

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

### Browser (if applicable)

_No response_

When a non-admin user goes to /studio, the 'Installed Projects' area should just be populated with projects that they have access to, i.e for which they have a project-permission record.

If a user goes to /studio?project=<project_name>, and project_name is a project they do not have a permission for, they should be redirected back to the plain /studio route. 

In file-browser, when disallowed information is stripped out of the returned results, the count should be updated to reflect that.

### What happened?

When a non-admin user goes to /studio, they are shown all projects in Installed Projects.

When a user goes to /studio?project=<project_name> for a project that they do not have permission for, the main studio controls appear, but nothing is loaded, which is appropriate since they do not have access to that project. But the file browser shows an endless loading message, which could be confusing; they should be redirected to /studio, which should then have just the projects they have permissions for.

When the file browser makes a call to get a directory's files, if the user does not have permission for that project, in addition to stripping out the results they are not allowed to access, the count should be updated, as well. For instance, if an admin installed eepro-globe-theater, and a non-admin made a new project test1, the non-admin, upon going to the top-level projects folder in the file browser, would see '1-3 of 3' in the count next to the pagination controls, even though they would only see their project test1 listed.

Related to the above, if that non-admin goes to /studio?project=default-project, then clicks on the file browser, the file browser will be at /projects/default-project and show e.g. 1-25 of 25, even though no files show up since that user does not have permission for that project.

### Steps to reproduce

```shell
Make a guest user in a separate browser context and, as an admin, give them studio scope. Go to /studio as the guest, and you'll see all of the projects.

Go to /studio?project=default-project as the guest and you'll see the main studio controls without any project/scene. Interact with the file browser as described above to see those problems.
```

### Relevant Info

_No response_

MoizAdnan:
closes #9443


packages/client-core/src/common/services/FileBrowserService.ts
```diff
@@ -67,7 +67,7 @@ export const FileBrowserService = {
       fileBrowserState.merge({
         files: files.data,
         skip: files.skip,
-        total: files.total,
+        total: files.data.length,
```
HexaField: is this correct? shouldn't `total` be total count of items, not just however many is returned in our pagination?

packages/server-core/src/projects/project/project.hooks.ts
```diff
@@ -202,7 +202,7 @@ const ensurePushStatus = async (context: HookContext<ProjectService>) => {
       context.projectPushIds = context.projectPushIds.concat(matchingAllowedRepos.map((repo) => repo.id))
     }
 
-    if (!(await checkScope(context.params.user!, 'projects', 'read')))
+    if (!(await checkScope(context.params.user!, 'admin', 'admin')))
```
HexaField: why are we changing this from projects:read to admin:admin? don't we still want to use our scoped permissions? admin:admin should only be for critical infrastructure

HexaField (changes requested):

packages/client-core/src/common/services/FileBrowserService.ts
MoizAdnan: Updated

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -152,7 +151,7 @@ export class FileBrowserService
     }
 
     return {
-      total,
+      total: result.length,
```
HexaField: this will take the sliced result, what is there currently is correct

HexaField (changes requested):

packages/server-core/src/media/file-browser/file-browser.class.ts
MoizAdnan: I did this for the user seeing higher page count than the total items when going to the root directory in file browser. As the total returns all projects count while the result only has the projects, user has access to.

MoizAdnan: reverted

HexaField: oh gotcha, yes that change is correct then, my bad

HexaField:
@MoizAdnan can you please re-review this after the recent project changes, and resolve conflicts

packages/server-core/src/media/file-browser/file-browser.class.ts
barankyle: I think this needs to be restored to what it was here, since that was the intended behavior.

HexaField: @MoizAdnan

MoizAdnan: This was an older change and I had reverted it sometime back. As this PR https://github.com/EtherealEngine/etherealengine/pull/9836 ,  fixes the total bug

PR #9836, opened 2024-03-05
show correct total for user with project-permissions

### LINKED ISSUE #9827: [Bug] File-browser shows incorrect total pages in root directory
In file-browser when a user (who doesn't have access to all projects) goes to the root folder, the total number of pages is higher than the number of items the file-browser service returns. For example, if a user has access to one project `test1` they would see '1-3 of 3' in the count next to the pagination controls, even though they would only see their project `test1` listed.

This is due to the `result.total` in file-browser service's find being set before project-permission verification. If, the `result.total` is set after the verification to fix this issue, then the total is of the sliced result and will be incorrect.

aditya-mitra:
## Summary

- show the total from project permission when the user is non admin

## References
closes #9827


packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -132,12 +136,15 @@ export class FileBrowserService
...
-        .options({ nestTables: true })
+      const projectPermissions: { 'project-permission': ProjectPermissionType; project: ProjectType }[] =
+        await knexClient
+          .from(projectPermissionPath)
+          .join(projectPath, `${projectPermissionPath}.projectId`, `${projectPath}.id`)
+          .where(`${projectPermissionPath}.userId`, params.user!.id)
+          .select()
+          .options({ nestTables: true })
+
+      total = projectPermissions.length
```
MoizAdnan: Although this solves the bug for 'projects/' directory, this will return incorrect total for subdirectories. Moving this line inside `if (directory === 'projects/')` should fix this problem but I'm not sure if this is the desired way to approach this. @HexaField what do you think?

MoizAdnan (changes requested):

packages/server-core/src/media/file-browser/file-browser.class.ts
HexaField: This seems like a good solution. As long as we aren't using git every time we need to get the permissions.

MoizAdnan (approved):

HexaField:
@MoizAdnan can you please re-review this after the recent project changes

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -123,21 +127,28 @@ export class FileBrowserService
...
     let result = await storageProvider.listFolderContent(directory)
-    const total = result.length
+    let total = result.length
 
     result = result.slice(skip, skip + limit)
     result.forEach((file) => {
       file.url = getCachedURL(file.key, storageProvider.cacheDomain)
     })
 
+    console.log('debug1 the directory was', directory)
```
DanielBelmes: console log left over from debug

aditya-mitra: removed!

MoizAdnan (approved):

PR #9840, opened 2024-03-06
Dragging a new version of the same file into model src field does not load new asset

### LINKED ISSUE #9799: [Bug] Dragging a new version of the same file into model src field does not load new asset
this is due to loadGLTF not reloading because it's checking the value of modelComponent.src.value

MichaelEstes:
## References
closes #9799


DanielBelmes:
qaing and functionality looks really stable

DanielBelmes (approved):

PR #9843, opened 2024-03-06
Implemented changes in instance provision service for server addon refactor.

hurairahmateen:
## Summary
- Added `{ 'metadata.labels.serverSize': params.serverSize }` to `provisionConstraints` which specify the server size type.
- Handles error handling for for both creating and updating locations.

## References
closes #_insert number here_


packages/server-core/src/networking/instance-provision/instance-provision.class.ts
```diff
@@ -462,6 +462,27 @@ export async function checkForDuplicatedAssignments({
...
+ * A method that calculates and returns the server size category based on the maximum number of users per instance.
+ * This method categorizes the server size into 'small', 'medium', 'large', or 'default' based on the
+ * value of `maxUsersPerInstance`.
+ * @param maxUsersPerInstance
+ * @returns
+ */
+
+export async function getServerSize(maxUsersPerInstance: number) {
+  // Determine the size based on maxUsersPerInstance
+  if (maxUsersPerInstance > 20 && maxUsersPerInstance <= 50) {
```
barankyle: I feel like <= 20 should be small, 20-50 should be medium, and > 50 should be large. I don't want a default size, as it isn't clear from the naming convention what a default size would be (though I get from this function that it would be <= 20).

hanzlamateen: So if there is no default size, does this means that each instance, should be a paid one or the ones that come as part of subscription? And we do not allow a default ones?

barankyle (approved):

hanzlamateen (approved):

PR #9845, opened 2024-03-06
Refactor mesh bvh worker to use proper worker pool

HexaField:
## References
closes #_insert number here_


the-reality-engineer (approved):

packages/engine/src/scene/components/PrimitiveGeometryComponent.ts
```diff
@@ -145,6 +145,7 @@ function GeometryReactor() {
   useLayoutEffect(() => {
     if (!mesh) return
     mesh.value.geometry = geometryComponent.geometry.get(NO_PROXY)
+    mesh.set(mesh.get(NO_PROXY_STEALTH)) // reactivly update mesh
```
the-reality-engineer, line 148: I believe doing `mesh.geometry.set(…)` would accomplish the same goal.

PR #9848, opened 2024-03-06
[Bug] Intersection calls are slow prior to BVH completing it's calculation

### LINKED ISSUE #9825: [Bug] Intersection calls are slow prior to BVH completing it's calculation
We need to make sure that raycast intersection queries do not run if the BVH has not completed. We should probably force threejs to not return anything, rather than falling back to it's own implementation.

MichaelEstes:
## Summary
Waits for MeshBVH to be generated before populating raycast hits without falling back to the slower three js raycast

## References
closes #9825


HexaField (commented):
Just commenting, we need to make sure transform gizmo and xrui still work

HexaField (approved):

PR #9862, opened 2024-03-08
Render order component is used

HexaField:
## References
closes #_insert number here_


DanielBelmes (approved):
From some experimentation earlier this week. We aren't passing 'visible' down from groups because of how we are proxifying group relationships. This will likely suffer the same fate of applying to the group but not its children. But I imagine this PR is serving a demo purpose.

PR #9863, opened 2024-03-08
Fixed bug with user disconnect on world instanceserver.

barankyle:
## Summary

If a user disconnected because they had signed in from being a guest, and the guest user was then destroyed because the logged-in user already existed, the instanceserver's attempt to create a 'user left' message was throwing an error because looking up that user by ID returned nothing. Now catching that and creating a generic user left message since their name is now impossible to find, with a null senderId because messages can't have a senderId of a non-existent user.

## References
closes #_insert number here_


HexaField:
I think if the leave as a guest, maybe we shouldn't specify that they have left at all. This is caused by the removal of world state, we should be keeping track of this data somewhere still. This is a good fix to get things working again though.

PR #9871, opened 2024-03-10
Exlude UVOL from Normal Pass in post processing

DanielBelmes:
## Summary
Since Uvol geometry doesn't move it was being caught up in the normal pass used in SSAO in its bind pose and was causing a ghost like artifact with SSAO.

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/renderer/constants/ObjectLayers.ts
+++ b/packages/spatial/src/renderer/constants/ObjectLayers.ts
@@ -56,5 +56,7 @@ export const ObjectLayers = {
   // transform gizmo
   HighlightEffect: 11 as const,
 
+  UVOL: 30 as const,
+
   AssetPreview: 31 as const
 } as Record<string, number>
--- a/packages/spatial/src/renderer/passes/CustomNormalPass.js
+++ b/packages/spatial/src/renderer/passes/CustomNormalPass.js
@@ -1,3 +1,30 @@
+/*
+Original Source: https://github.com/pmndrs/postprocessing
+CPAL-1.0 License
+
+The contents of this file are subject to the Common Public Attribution License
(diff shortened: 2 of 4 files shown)


packages/spatial/src/renderer/constants/ObjectLayers.ts
```diff
@@ -56,5 +56,7 @@ export const ObjectLayers = {
   // transform gizmo
   HighlightEffect: 11 as const,
 
+  UVOL: 30 as const,
+
   AssetPreview: 31 as const
 } as Record<string, number>
```
DanielBelmes, line 62: We can't go over 31 atm due to maxBitWidth. We should consider refactoring some of these if we can

packages/spatial/src/renderer/passes/CustomNormalPass.js
```diff
@@ -0,0 +1,223 @@
...
+	 * @param {WebGLRenderTarget} outputBuffer - A frame buffer that serves as the output render target unless this pass renders to screen.
+	 * @param {Number} [deltaTime] - The time between the last frame and the current one in seconds.
+	 * @param {Boolean} [stencilTest] - Indicates whether a stencil mask is active.
+	 */
+
+	render(renderer, inputBuffer, outputBuffer, deltaTime, stencilTest) {
+
+		const renderTarget = this.renderToScreen ? null : this.renderTarget;
+    this.renderPass.camera.layers.disable(ObjectLayers.UVOL)
+		this.renderPass.render(renderer, renderTarget, renderTarget);
```
DanielBelmes, line 204: This is where the only change is. Disable and enable of the layer on the camera

the-reality-engineer (approved):

PR #9878, opened 2024-03-10
Improved buffer management in UVOL2Component

the-reality-engineer:
Buffer now correctly maintains metadata about buffered media, in sorted time range segments that are merged when media segments are fetched and buffered.


the-reality-engineer:
@CITIZENDOT  can you review this and make sure it looks good to merge ?

CITIZENDOT (approved):

PR #9890, opened 2024-03-12
Fixed issue in build_and_publish_package.sh

barankyle:
## Summary

For some reason, having a builder name be the same as the context name is now disallowed. Changed the context name so they're not identical.

Updated buildkit to v0.13

## References
closes #_insert number here_


DanielBelmes (approved):
gotta love docker

PR #9904, opened 2024-03-13
Clean up hyperflux loose ends

HexaField:
## Summary

- [x] Remove $from field from actions
- [x] Remove UserID from hyperflux
- [x] Move PeerID to hyperflux

## References
closes #_insert number here_


packages/client-core/src/admin/services/SceneService.ts
```diff
@@ -40,7 +41,7 @@ export const AdminSceneState = defineState({
     fetched: false,
     updateNeeded: true,
     lastFetched: Date.now(),
-    singleScene: { scene: {} } as SceneDataType
+    singleScene: { scene: {} } as Omit<SceneDataType, 'scene'> & { scene: SceneJsonType }
```
DanielBelmes, line 44: the type gymnastics here are a bit funky. Is typebox not getting the right type from the schema?

HexaField: No, the type was dependent on the ecs package, which common can not have. This stuff will be replaced with gltf soon, so it's more or less temporary.

DanielBelmes: oh scrolled down. Sorry. Was working off local from when I clicked in github. I guess it changes the question to. You can't pass SceneJsonType to typebox?

packages/projects/ProjectConfigInterface.ts
```diff
@@ -65,7 +65,7 @@ export interface ProjectConfigInterface {
    * This is loaded on ALL instances of the engine if enabled
    * Allows for running of custom logic regardless of which scene or route is loaded
    */
-  worldInjection?: () => Promise<{ default?: () => Promise<void> }>
+  worldInjection?: () => Promise<{ default?: () => Promise<void> } | any>
```
DanielBelmes, line 68: was this meant to stay? I feel like we can have a stricter type here

packages/client-core/src/admin/services/SceneService.ts
HexaField: We might be able to

DanielBelmes (approved):
Alright looks good. Though please double check the world injection type.

PR #9905, opened 2024-03-13
Tailwind Admin Routes

aditya-mitra:
## Summary

- migrated admin routes table into tailwind
- fix bugs in table pagination

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/admin2/DefaultAdminRoutes.tsx
+++ b/packages/client-core/src/admin2/DefaultAdminRoutes.tsx
@@ -25,7 +25,7 @@ Ethereal Engine. All Rights Reserved.
 
 import React, { lazy } from 'react'
 import { HiOutlineCube } from 'react-icons/hi'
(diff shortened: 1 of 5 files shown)


packages/client-core/src/admin2/DefaultAdminRoutes.tsx
```diff
@@ -101,5 +103,12 @@ export const DefaultAdminRoutes: Record<string, AdminRouteStateType> = {
     component: Resources,
     access: false,
     icon: <HiOutlineTableCells />
+  },
+  routes: {
+    name: 'user:dashboard.routes',
+    scope: 'routes',
+    component: Routes,
+    access: false,
+    icon: <HiGlobeAlt />
```
CITIZENDOT: More accurate icon is: `HiOutlineGlobeAlt`

CITIZENDOT (changes requested):

packages/client-core/src/admin2/DefaultAdminRoutes.tsx
aditya-mitra: changed!

PR #9916, opened 2024-03-14
Tailwind Admin Channels

aditya-mitra:
## Summary

- migrate admin channels into tailwind

## References
closes #_insert number here_


packages/client-core/src/admin2/components/channel/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+                fullWidth
+                onClick={() => {
+                  PopoverState.showPopupover(<AddEditChannelModal />)
+                }}
+              >
+                {t('admin:components.channel.createChannel')}
+              </Button>
+            </div>
+          </div>
+        </div>
```
CITIZENDOT, line 102: image
Since there is ample space to the left, we can try to avoid wrapping.

CITIZENDOT (changes requested):

packages/client-core/src/admin2/components/channel/index.tsx
aditya-mitra: image
implemented!

PR #9923, opened 2024-03-14
Devops fixes

DanielBelmes:
## Summary
* Ensured .env is bash syntax
* Added sleep to dev docker to ensure servers are up.**(If you know of a way to test server health before configuring feathers/koa/knex let me know)**
* Ensured anything using certs uses env vars
* removes file server from dev tabs
* removed genereate certs
* made sure minio will only be launched if configured so resources aren't wasted

## References
closes #_insert number here_


package.json
```diff
@@ -55,7 +55,7 @@
     "dev-agones-silent": "npm run dev-agones &> /dev/null",
     "dev-client": "cd packages/client && npm run dev",
     "dev-taskserver": "cd packages/taskserver && npm run dev",
-    "dev-docker": "cd scripts && ./start-containers.sh",
+    "dev-docker": "cd scripts && ./start-containers.sh && sleep 10",
```
DanielBelmes, line 58: this is sketch but it works. We launch our app too quickly and sql usually isn't up yet.

barankyle (approved):

DanielBelmes:
looking into test failures

PR #9926, opened 2024-03-15
Updated build process with more generalized repo information.

barankyle:
## Summary

There was a notable bug with the recent change to build the root image as part of build_docker_builder.sh: the service Dockerfiles built in the builder process were sourcing their root images from an AWS repo, but if a deployment was not linked to a Github action building images for that deployment, then its root image would either always be stale or not exist at all.

This generalizes the source and destination repos for Docker images. Dockerfiles have a generic REPO_URL arg, and pull the tagged image matching the tag from the builder image being run. The source and destination repos and names have been generalized too, so that it is more clear that one can source from a Docker Hub repo and push to an AWS repo. This should also make it easier to add other cloud sources and destinations like Digital Ocean - just add an else if clause to the necessary files.

This change does mean that pushing will only go to a single destination; one cannot push to both Docker Hub and AWS ECR. But that situation did not seem to be occurring anyway.

## References
closes #_insert number here_

https://github.com/EtherealEngine/etherealengine-docs/pull/128


muradkhateeb78 (approved):
Hey Kyle, I have reviewed the changes in detail and they look good to me.

PR #9931, opened 2024-03-15
Clean up dithering component and basic materials

AidanCaruso:
## Summary
Refactors the dithering component to use a component for every distance dithering center point and addresses the original implementation requiring the shader logic to be reinjected with every uniform update to get around material race conditions. This is done by using deterministic UUIDs for newly generated lambert materials under the forceBasicMaterials condition, preventing duplicate materials from being generated and stored in the state, and registering/unregistering original and lambert materials when setting forceBasicMaterials.

## References
closes #_insert number here_


packages/engine/src/avatar/components/TransparencyDitheringComponent.tsx
```diff
@@ -0,0 +1,165 @@
...
+  localPosition = 0
+}
+
+export const transparencyDitheringState = defineState({
+  name: 'transparencyDitheringState',
+  initial: { materialIds: {} as Record<Entity, string[]> }
+})
+
+export const maxDitherPoints = 4 //should be equal to the length of the vec3 array in the shader
+export const TransparencyDitheringComponent = Array.from({ length: maxDitherPoints }, (_, i) => {
```
HexaField, line 49: do we need multiple components here, or do we just want to make the data an array?

HexaField (dismissed):

packages/engine/src/avatar/components/TransparencyDitheringComponent.tsx
AidanCaruso: I went with an array of components as opposed to an array of component values because it seemed straightforward and easier to introspect than either of the other options I considered in that vein. When you have arrays of values in the component you can expect the arrays to end up with different lengths, which would likely require helper functions for adding/removing dithering points to avoid some confusing inconsistency with correspondance to the shader uniforms. I saw using set/remove component for this as a UX simplifcation that ensures there will always be values that correspond to a dither point for a given index, without abstracting the data into an array of objects within the component. Is using an array of components in this context bad practice, is it less performant than I thought?

HexaField: no worries, it shouldn't be much of a performance concern.

packages/engine/src/avatar/components/TransparencyDitheringComponent.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+    onSet: (entity, component, json) => {
+      if (!json) return
+      if (matchesVector3.test(json.center)) component.center.set(new Vector3().copy(json.center))
+      if (matches.number.test(json.distance)) component.distance.set(json.distance)
+      if (matches.number.test(json.exponent)) component.exponent.set(json.exponent)
+      if (matches.number.test(json.calculationType)) component.calculationType.set(json.calculationType)
+    },
+
+    reactor: () => {
```
HexaField: we should use a query reactor here instead of a new reactor for each component in the array

packages/engine/src/avatar/systems/AvatarAnimationSystem.ts
```diff
@@ -330,23 +330,28 @@ const execute = () => {
     updateVRMRetargeting(rigComponent.vrm, entity)
   }
 
-  //update local client entity's dithering component and camera attached logic
+  //self avatar entity's dithering component and camera attached logic
```
HexaField: let's move all this logic to a new system

packages/engine/src/mocap/MotionCaptureSystem.ts
```diff
@@ -92,6 +92,7 @@ const handleMocapData = (
     timeSeriesMocapData.set(fromPeerID, new RingBuffer(10))
   }
   timeSeriesMocapData.get(fromPeerID)!.add(results)
+  console.log(results)
```
HexaField: clean up logs

HexaField (changes requested):

HexaField (approved):

PR #9933, opened 2024-03-16
Switch mocap wasm, optimizations and more filtering

AidanCaruso:
## Summary
The goal of this PR is to get smoother mocap with a frame target of 60fps on my hardware, at least.

## References
closes #_insert number here_


packages/engine/src/mocap/solveMotionCapturePose.ts
```diff
@@ -394,19 +394,19 @@ export function solveMotionCapturePose(
...
     if (MotionCaptureRigComponent.solvingLowerBody[entity]) {
       MotionCaptureRigComponent.solvingLowerBody[entity] = 0
     }
+    //zero bone quats to filter them out in motion capture system
+    for (const boneName of lowerBody) {
+      //only leg bones
+      MotionCaptureRigComponent.rig[boneName].x[entity] = 0
+      MotionCaptureRigComponent.rig[boneName].y[entity] = 0
+      MotionCaptureRigComponent.rig[boneName].z[entity] = 0
+      MotionCaptureRigComponent.rig[boneName].w[entity] = 0
```
HexaField, line 408: should we be setting `w` to 1 instead of 0?

packages/engine/src/mocap/MotionCaptureSystem.ts
```diff
@@ -168,7 +168,7 @@ const execute = () => {
               MotionCaptureRigComponent.rig[boneName].w[entity]
             )
             .normalize(),
-          getState(ECSState).deltaSeconds * 50
+          getState(ECSState).deltaSeconds * 35
```
HexaField: what is this multiplier doing exactly?

HexaField (dismissed):

packages/engine/src/mocap/MotionCaptureSystem.ts
AidanCaruso: this increases the alpha of the lerp function, so the rate of change between the slerped quaternion and the goal quaternion is increased

packages/engine/src/mocap/solveMotionCapturePose.ts
AidanCaruso: At the moment, a zeroed out quaternion acts as a flag for the motion capture system that prevents a quaternion from being written to the rig. No quaternions written means the animation mixer will take over and these bones will just be animated.

With that logic, if we changed this to 1, an identity quaternion would be written to the rig for each of these bones.

the-reality-engineer (approved):

Issue #9908, opened 2024-03-13
[Bug]: Rotation Gizmo Not Working


dinomut1:
### Version

08c3b70(1.6.0)

### Operating Systems

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

### What is supposed happened?

Not applicable

### What happened?

[Screencast from 03-13-2024 03:55:45 PM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/6d6fcc14-be6b-4cb5-aac0-1e3696f5a7bf)

### Steps to reproduce

```shell
Try to rotate anything using gizmo. Not correctly registering mouse hover, and click and drag doesn't register either
```

SYBIOTE:
this is fixed closing this

PR #9935, opened 2024-03-17
Fix gizmo errors and variables

SYBIOTE:
This pull request fixes gizmo errors and variables in the code.

Closes:
#9908


DanielBelmes (changes requested):
While QA'ing I kept getting into a state where dragging state was being stuck on true despite me not dragging and result in me being stuck on a single axis.

packages/editor/src/components/toolbar/tools/PlayModeTool.tsx
```diff
@@ -81,6 +82,8 @@ const PlayModeTool = () => {
       // getMutableState(EngineState).isEditing.set(false)
       // run all visual script logic
       visualScriptQuery().forEach((entity) => dispatchAction(VisualScriptActions.execute({ entity })))
+      transformGizmoControlledQuery().forEach((entity) => removeComponent(entity, TransformGizmoControlledComponent))
```
DanielBelmes, line 86: Sadly you still have to make it that they can not appear when play mode is active. It does remove the transformGizmo but if I click on the heirachy it still creates a new transform gizmo.

DanielBelmes: Looking really really good though. It's UX is flawless. I just wanna cover this one edge case and lets merge.

DanielBelmes (approved):

PR #9955, opened 2024-03-20
9947 improve project fetch

### LINKED ISSUE #9947: [Bug]: Fetching all projects can be slow because of populating permissions on each one
### Version

08c3b70(1.6.0)

### Operating Systems

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

### Browser (if applicable)

_No response_

Fetching projects from the API should not take a visibly long time.

### What happened?

Fetching projects on a deployment can take several seconds, which is mildly annoying, moreso when you refresh a page showing projects and have to wait another several seconds for them to be populated.

### Steps to reproduce

```shell
Go to a deployment with several dozen projects, like dev, and go to /admin/projects or /studio. The list of projects will take several seconds to display because it's waiting for the API request for project.find to return.
```

### Relevant Info

From testing, it seems like the main culprit is the project resolver searching for each projects' project-permissions individually. Each request doesn't take too long, but they aren't being started at the same time, the glut of requests might be slowing each of them down, and the overall effect is to add an appreciable amount of time - on the order of 1.5 seconds.

This is happening twice in a project.find request from the client. In ensurePushStatus in project.hooks.ts, there's a call on ~line 200 to make a project.find matching an array of repository paths, and this is unnecessarily getting project permissions resolved on it. Then, when the top-level API call finishes, it's also calling the projectPermissions resolver on each project.

Part of the problem is that most calls to project.find are explicitly turning off pagination, and thus when there's several dozen projects, it has to make several dozen calls to project-permissions individually. But also, many of the calls to project.find don't need permissions to be returned; the only time they're used is on the root studio page, when viewing/editing a project's permissions. Other components like loadWebappInjection.ts and loadEngineInjection.ts make a project.find but don't do anything with permissions.

A solution might be to remove this resolver, and explicitly fetch a project's permissions only when needed on the studio page, or wherever else we might choose to have users interact with project permissions.

barankyle:
## Summary

Fixed inefficiency in resolving project's permissions
    
All project methods were trying to resolve a project's permissions.
This would in turn require a couple additional database lookups.
When many projects were trying to be fetched at once, this would
make the overall project.find take several seconds.

The only place that project permissions are currently used is on the
main studio page, and they're not displayed or edited unless the user
has opened the context modal for a specific project.

Adding a project's permissions in the resolver has been made an opt-in
query param which gets moved to context.params from context.params.query.
Otherwise, it just returns an empty array, speeding the process up
quite a bit.

The studio main page has been altered so that when a project's context
is loaded, it fetches just that project's permissions and manually adds
them onto the project.

## References
closes #9947


packages/client-core/src/admin/components/Project/UserPermissionDrawer.tsx
```diff
@@ -78,8 +78,12 @@ const UserPermissionDrawer = ({ open, project, onClose }: Props) => {
...
 
     try {
-      await ProjectService.createPermission(userInviteCode, project.id)
-      await ProjectService.fetchProjects()
+      const newPermission = await ProjectService.createPermission(userInviteCode, project.id)
+      if (
+        Array.isArray(project?.projectPermissions) &&
+        project.projectPermissions.findIndex((permission) => permission.id === newPermission.id) < 0
+      )
+        project.projectPermissions = project.projectPermissions.concat([newPermission])
```
HexaField: `project` here is coming from a reactive store (inside useFind), it will cause desync issues inside hookstate. instead, we should make a `projectPermissions` state inside this react component that is set here, thus can be reacted to

HexaField (changes requested):

HexaField:
Looks good, I'll let @MoizAdnan have the final word.

MoizAdnan (approved):

Issue #9900, opened 2024-03-13
Track code generated assets in the resource manager


MichaelEstes:
Assets generated by code (eg. Lights) are not tracked by the resource manager, adding an api to the resource manager to track will allow them to be referenced counted and disposed of when no longer in use.

PR #9956, opened 2024-03-20
System profiling, device profile data with fallback, track generated resources through resource manager

### LINKED ISSUE #9941: Integrate performance profiling and system profiling into tunable parameters
- [ ] Build ee-development-test-suite benchmark scenes & benchmark tests for:
  - [ ] Physics
  - [ ] Avatar rendering
  - [ ] IK calculations
  - [ ] ... figure out more hot spots
- [ ] Priority queues have dynamic limits based on performance profiler

MichaelEstes:
## Summary

#9941 
- Exposes engine version in vite.config
- Add functions for aggregating system performance
- Store GPU and device information in PerformanceState
- Pull request [82](https://github.com/EtherealEngine/ee-development-test-suite/pull/82) in ee-development-test-suite relies on these changes to run benchmarks

#9942 
- Adds GPU timing code through WebGL extensions with fallback for non-supported browsers in PerformanceState.ts and WebGLRendererSystem.ts

#9900 
- Adds hooks in resourceHooks.ts for managing code generated resources that allocate GPU memory through ResourceState.ts
- Splits code from components in the happy path that are creating code generated resources into their own components that use the managed hook flow

Misc
- Fix bugs around layers being used as layer masks
- Component tests
- Fixed small bug in the system debug window that was preventing system's frame time from being displayed

## References
closes #9941 #9900 #9942


packages/spatial/src/renderer/components/MeshComponent.ts
```diff
@@ -48,13 +48,23 @@ export const MeshComponent = defineComponent({
...
+ *
+ * Adds a MeshComponent to an entity with the passed in Geometry and Material
+ * return the geometry and material wrapped in a state, if the state of the geometry or material is updated it will update the geometry and material for the mesh
+ *
+ * @param entity entity to add the mesh component to
+ * @param geometry a Geometry instance to add to the mesh
+ * @param material a Material instance to add to the mesh
+ * @returns [Mesh, State<Geometry>, State<Material>]
+ */
 export function useMeshComponent<
```
HexaField: why a separate useMeshComponent hook instead of just making this the component reactor for the MeshComponent

MichaelEstes: The typing of the mesh's geometry and material is lost when going directly through components, I think you're right about the logic being scoped to the component reactors and I can make that change, but I think the hook is still useful as type inference wrapper.

packages/engine/src/scene/components/ObjectGridSnapComponent.ts
```diff
@@ -23,136 +23,145 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
-    lineSegmentList.push(new Vector3(xLeft, bbox.min.y, z), new Vector3(xLeft, bbox.max.y, z))
-    lineSegmentList.push(new Vector3(xRight, bbox.min.y, z), new Vector3(xRight, bbox.max.y, z))
-  }
-  for (let i = 0; i < lineSegmentList.length; i++) {
-    lineSegmentList[i].applyMatrix4(matrixWorld)
+  onSet: (entity, component, json) => {
+    if (!json || !json.bbox || !json.bbox.isBox3) throw new Error('BoundingBoxHelperComponent: Requires Box3')
+    component.bbox.set(json.bbox)
+
+    if (matchesMatrix4.test(json.matrixWorld)) component.matrixWorld.set(json.matrixWorld)
```
HexaField: if this matrixWorld is only being used internally, we don't need/want a deserializer for it. furthermore, is this distinct from the transform matrixWorld? if not, we should remove it

HexaField (changes requested):

packages/spatial/src/renderer/functions/performanceHooks.ts
```diff
@@ -27,24 +27,46 @@ import { State, useHookstate, useMutableState } from '@etherealengine/hyperflux'
...
-    performanceTier.set(performanceState.tier.value)
-  }, [performanceState.tier])
+    performanceTier.set(performanceState.gpuTier.value)
+  }, [performanceState.gpuTier])
 
   return performanceTier
 }
 
-export const usePerformanceOffset = (): State<number> => {
+export const useCPUPerformanceTier = (): State<number> => {
```
HexaField: i think these hooks are a bit overengineered, they can be as simple as `useHookstate(getMutableState(PerformanceState).cpuTier)`, unless this is in preperation for more sophisticated and configurable heuristic functions, in which case these still seem premature

HexaField (changes requested):

packages/engine/src/audio/components/PositionalAudioComponent.ts
```diff
@@ -25,16 +25,19 @@ Ethereal Engine. All Rights Reserved.
...
 
-import { defineComponent, useComponent, useOptionalComponent } from '@etherealengine/ecs/src/ComponentFunctions'
-import { getMutableState, none, useHookstate } from '@etherealengine/hyperflux/functions/StateFunctions'
+import {
+  defineComponent,
+  removeComponent,
+  setComponent,
+  useComponent,
+  useOptionalComponent
+} from '@etherealengine/ecs/src/ComponentFunctions'
```
HexaField: for the ecs and hyperflux packages, prefer module level '@etherealengine/packagename' imports instead of deep imports

packages/engine/src/audio/components/PositionalAudioHelperComponent.ts
```diff
@@ -0,0 +1,153 @@
...
+import { useEffect } from 'react'
+import { BufferAttribute, BufferGeometry, InterleavedBufferAttribute, Line, LineBasicMaterial, MathUtils } from 'three'
+import { AudioNodeGroup } from '../../scene/components/MediaComponent'
+
+export const PositionalAudioHelperComponent = defineComponent({
+  name: 'PositionalAudioHelperComponent',
+
+  onInit: (entity) => {
+    return {
+      name: 'positional-audio-helper',
```
HexaField, line 40: is this property necessary to be configurable? does this name field do anything functional? if not, we should just hardcode it and not expose it as an API

packages/engine/src/avatar/components/AvatarIKComponents.ts
```diff
@@ -68,13 +67,14 @@ export const AvatarIKTargetComponent = defineComponent({
...
 
     useEffect(() => {
-      if (!debugEnabled.value) return
-      const helper = new AxesHelper(0.5)
-      addObjectToGroup(entity, helper)
-      setObjectLayers(helper, ObjectLayers.AvatarHelper)
-      setComponent(entity, VisibleComponent)
-      return () => {
-        removeObjectFromGroup(entity, helper)
+      if (!debugEnabled.value) {
```
HexaField: prefer a cleanup function to remove the component, such that when the AvatarIKTargetComponent is removed, this component is also removed

packages/engine/src/scene/components/EnvMapBakeComponent.ts
```diff
@@ -108,26 +104,16 @@ export const EnvMapBakeComponent = defineComponent({
...
-      setVisibleComponent(helperEntity, true)
-      setObjectLayers(helper, ObjectLayers.NodeHelper)
-      bake.helperEntity.set(helperEntity)
-
-      return () => {
-        removeEntity(helperEntity)
-        if (!hasComponent(entity, EnvMapBakeComponent)) return
-        bake.helperEntity.set(none)
+      if (!debugEnabled.value) {
+        removeComponent(entity, MeshHelperComponent)
```
HexaField: prefer return cleanup function for removing the component

packages/engine/src/scene/components/MediaComponent.ts
```diff
@@ -487,25 +481,17 @@ export function MediaReactor() {
...
-    setComponent(helperEntity, EntityTreeComponent, { parentEntity: entity })
-    setVisibleComponent(helperEntity, true)
-    setObjectLayers(helper, ObjectLayers.NodeHelper)
-    media.helperEntity.set(helperEntity)
-
-    return () => {
-      removeEntity(helperEntity)
-      media.helperEntity.set(none)
+    if (!debugEnabled.value) {
+      removeComponent(entity, MeshHelperComponent)
```
HexaField: prefer return cleanup function for removing the component

packages/engine/src/scene/components/MountPointComponent.ts
```diff
@@ -75,26 +67,12 @@ export const MountPointComponent = defineComponent({
...
-      setVisibleComponent(helperEntity, true)
-      setObjectLayers(helper, ObjectLayers.NodeHelper)
-      mountPoint.helperEntity.set(helperEntity)
-
-      return () => {
-        removeEntity(helperEntity)
-        if (!hasComponent(entity, SpawnPointComponent)) return
-        mountPoint.helperEntity.set(none)
+      if (!debugEnabled.value) {
+        removeComponent(entity, ArrowHelperComponent)
```
HexaField: prefer return cleanup function for removing the component

packages/engine/src/scene/components/PortalComponent.ts
```diff
@@ -179,27 +177,15 @@ export const PortalComponent = defineComponent({
...
-      )
-
-      portalComponent.helperEntity.set(helperEntity)
-
-      return () => {
-        removeEntity(helperEntity)
-        if (!hasComponent(entity, PortalComponent)) return
-        portalComponent.helperEntity.set(none)
+      if (!debugEnabled.value) {
+        removeComponent(entity, ArrowHelperComponent)
```
HexaField: prefer return cleanup function for removing the component

packages/engine/src/scene/components/PrimitiveGeometryComponent.ts
```diff
@@ -91,7 +92,7 @@ function GeometryReactor() {
   const entity = useEntityContext()
   const geometryComponent = useComponent(entity, PrimitiveGeometryComponent)
   const transform = useComponent(entity, TransformComponent)
-  const mesh = useState<Mesh>(new Mesh())
+  const [mesh] = useResource(new Mesh(), entity)
```
HexaField, line 95: will this create a new mesh each time the reactor re-renders?

packages/engine/src/scene/components/ScenePreviewCamera.ts
```diff
@@ -100,22 +94,13 @@ export const ScenePreviewCameraComponent = defineComponent({
...
-      setComponent(helperEntity, EntityTreeComponent, { parentEntity: entity })
-      setVisibleComponent(helperEntity, true)
-      previewCamera.helperEntity.set(helperEntity)
-
-      return () => {
-        removeEntity(helperEntity)
-        if (!hasComponent(entity, ScenePreviewCameraComponent)) return
-        previewCamera.helperEntity.set(none)
+      if (!debugEnabled.value) {
+        removeComponent(entity, CameraHelperComponent)
```
HexaField: prefer return cleanup function for removing the component

packages/spatial/src/common/debug/DebugComponentUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+      setComponent(helperEntity, NameComponent, helper.name)
+    }
+    setComponent(helperEntity, EntityTreeComponent, { parentEntity: entity })
+    setComponent(helperEntity, UUIDComponent, generateEntityUUID())
+    setComponent(helperEntity, ObjectLayerMaskComponent, layerMask)
+    setVisibleComponent(helperEntity, true)
+    component.entity.set(helperEntity)
+
+    return () => {
+      if (helper) removeObjectFromGroup(helperEntity, helper)
```
HexaField, line 69: remove from group should be redundant if we are also calling removeEntity

packages/spatial/src/common/debug/LightHelperComponent.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (typeof json.name === 'string') component.name.set(json.name)
+    if (typeof json.size === 'number') component.size.set(json.size)
+    if (matchesColor.test(json.color)) component.color.set(json.color)
+  },
+
+  reactor: function () {
+    const entity = useEntityContext()
+    const component = useComponent(entity, LightHelperComponent)
+
+    const getLightHelperType = (light: Light) => {
```
HexaField, line 70: let's move this to module scope since it is pure

packages/spatial/src/common/debug/MeshHelperComponent.ts
```diff
@@ -0,0 +1,86 @@
...
+      helperEntity,
+      component.geometry.value,
+      component.material.value
+    )
+
+    useEffect(() => {
+      setComponent(helperEntity, ObjectLayerMaskComponent, component.layerMask.value)
+    }, [component.layerMask])
+
+    useDidMount(() => {
```
HexaField, line 70: what exactly is this useDidMount here distinct from a useEffect, and can these hooks be done inside useMeshComponent since they are using state passed in and mutating state coming from the hook?

packages/spatial/src/renderer/components/DirectionalLightComponent.ts
```diff
@@ -95,33 +98,22 @@ targetLineGeometry.setAttribute(
...
-      light,
+      light: null! as DirectionalLight,
       color: new Color(),
       intensity: 1,
       castShadow: false,
       shadowBias: -0.00001,
       shadowRadius: 1,
       cameraFar: 200,
-      helper: null as DirectionalLightHelper | null
+      useInCSM: true
```
HexaField, line 116: merge error - useInCSM has since been removed and now exists as an inversion of control on RenderSettingsComponent.

packages/spatial/src/renderer/components/DirectionalLightComponent.ts
```diff
@@ -164,90 +147,64 @@ export const DirectionalLightComponent = defineComponent({
...
-      return () => {
-        removeEntity(lightHelperEntity)
-        lightPlane.geometry.dispose()
-        lightPlane.material.dispose()
-        targetLine.geometry.dispose()
-        targetLine.material.dispose()
-        if (!hasComponent(entity, DirectionalLightComponent)) return
-        directionalLightComponent.helper.set(none)
+      if (!debugEnabled.value) {
+        removeComponent(entity, LineSegmentComponent)
```
HexaField, line 196: prefer return cleanup function for removing the component

packages/spatial/src/renderer/components/HemisphereLightComponent.ts
```diff
@@ -66,25 +73,38 @@ export const HemisphereLightComponent = defineComponent({
...
-      light.light.value.color.set(light.skyColor.value)
-    }, [light.skyColor])
+      light.intensity = hemisphereLightComponent.intensity.value
+    }, [hemisphereLightComponent.intensity])
 
     useEffect(() => {
-      light.light.value.intensity = light.intensity.value
-    }, [light.intensity])
+      if (!debugEnabled.value) {
+        removeComponent(entity, LightHelperComponent)
```
HexaField: prefer return cleanup function for removing the component

packages/spatial/src/renderer/components/InfiniteGridHelper.ts
```diff
@@ -134,102 +128,142 @@ void main() {
...
+    if (typeof json.distance == 'number') component.distance.set(json.distance)
+    if (typeof json.gridHeight == 'number') component.gridHeight.set(json.gridHeight)
+  },
+
+  reactor: function () {
+    const entity = useEntityContext()
+    const component = useComponent(entity, InfiniteGridHelperComponent)
+    const [mesh, geometry, material] = useMeshComponent(
+      entity,
+      new PlaneGeometry(2, 2, 1, 1),
```
HexaField, line 164: again, this will create new objects every render. though now that i think about it, it might just be a garbage inefficiency if class creation doesn't result in any side effects inside of three

packages/spatial/src/renderer/components/InfiniteGridHelper.ts
```diff
@@ -248,68 +282,39 @@ export const InfiniteGridComponent = defineComponent({
     if (typeof json.color === 'string') component.color.set(new Color(json.color))
     if (json.color instanceof Color) component.color.set(json.color)
     if (typeof json.distance === 'number') component.distance.set(json.distance)
+
+    setComponent(entity, InfiniteGridHelperComponent, {
```
HexaField: onSet should have no side effects, please move this to reactor mount

packages/spatial/src/renderer/components/MeshComponent.ts
```diff
@@ -36,5 +50,85 @@ export const MeshComponent = defineComponent({
   onSet: (entity, component, mesh: Mesh) => {
     if (!mesh || !mesh.isMesh) throw new Error('MeshComponent: Invalid mesh')
     component.set(mesh)
+    setComponent(entity, MaterialComponent, mesh.material)
```
HexaField: onset should have no side effects

packages/spatial/src/renderer/components/MeshComponent.ts
HexaField: That is fair, but I am uneasy about this hook still. Hooks should not be mutating data, and the strict binding of geometry component and material component to the mesh component is going to need to be changed soon for GLTF loading anyway, this will make that more dificult. These components should be separable, even though in practice they act together.

packages/spatial/src/renderer/components/PointLightComponent.ts
```diff
@@ -90,79 +90,62 @@ export const PointLightComponent = defineComponent({
...
-
-      setObjectLayers(helper, ObjectLayers.NodeHelper)
-
-      light.helperEntity.set(helperEntity)
-
-      return () => {
-        removeEntity(helperEntity)
-        light.helperEntity.set(none)
+      if (!debugEnabled.value) {
+        removeComponent(entity, LightHelperComponent)
```
HexaField: prefer return cleanup function for removing the component

packages/spatial/src/renderer/components/SpotLightComponent.ts
```diff
@@ -107,101 +102,76 @@ export const SpotLightComponent = defineComponent({
...
-
-      helper.layers.set(ObjectLayers.NodeHelper)
-
-      light.helperEntity.set(helperEntity)
-
-      return () => {
-        removeEntity(helperEntity)
-        light.helperEntity.set(none)
+      if (!debugEnabled.value) {
+        removeComponent(entity, LightHelperComponent)
```
HexaField: prefer return cleanup function for removing the component

HexaField (changes requested):

packages/engine/src/audio/components/PositionalAudioHelperComponent.ts
MichaelEstes: I just forgot to set it in the component, I've tried to have all the debug components reflect the names of the components that they're tied to, ie. an entity named MyAudio with a PositionalAudioComponent will have it's helper entity named MyAudio-positional-audio-helper. I could update the api to get the name automatically from the name component in the helper, but some components make multiple helpers so having it configurable allows having an index in the name

packages/spatial/src/renderer/components/DirectionalLightComponent.ts
HexaField: this is still relevant

packages/spatial/src/renderer/components/DirectionalLightComponent.ts
HexaField: this is still relevant

packages/spatial/src/common/debug/DebugComponentUtils.ts
HexaField: still relevant

packages/spatial/src/common/debug/LightHelperComponent.ts
HexaField: still relevant

packages/spatial/src/common/debug/MeshHelperComponent.ts
HexaField: still would like clarity on this

packages/spatial/src/renderer/components/InfiniteGridHelper.ts
MichaelEstes: Ya, I implemented this one incorrectly at the API level, I'll update it. The other places that were creating objects weren't updated to use the correct overload of useResource, those are fixed.

packages/spatial/src/common/debug/MeshHelperComponent.ts
MichaelEstes: In this case the useDidMount isn't needed, I switched it back to useEffect. MeshHelperComponent is a debug only component, I've renamed to to DebugMeshComponent to make it more explicit. it's a one off pass through component to keep the code concise, I'd rather change this component to not use useMeshComponent than change how useMeshComponent works for everything else

DanielBelmes:
moved to jira linked branch https://github.com/EtherealEngine/etherealengine/pull/10111

PR #9958, opened 2024-03-20
Implement non reactive canSleep

DanielBelmes:
## Summary
Ubiquity is utilizing the canSleep functionality and we didn't expose it. There's no good way to make it reactive without recreating the rigid body and I wasn't okay with doing so.

My only other idea would be to implement reactor init props that we can pass in so that the canSleep property isn't on the component and therefore no exposed as a reactive property.

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/physics/components/RigidBodyComponent.ts
+++ b/packages/spatial/src/physics/components/RigidBodyComponent.ts
@@ -77,7 +77,7 @@ export const RigidBodyComponent = defineComponent({
       angularVelocity: proxifyVector3(this.angularVelocity, entity),
       /** If multiplier is 0, ridigbody moves immediately to target pose, linearly interpolating between substeps */
       targetKinematicLerpMultiplier: 0,
-      canSleep: true
+      canSleep: true /** Not a reactive property. Only on Init*/
     }
   },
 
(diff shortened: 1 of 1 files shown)


AidanCaruso (approved):

packages/spatial/src/physics/components/RigidBodyComponent.ts
```diff
@@ -76,7 +76,8 @@ export const RigidBodyComponent = defineComponent({
       linearVelocity: proxifyVector3(this.linearVelocity, entity),
       angularVelocity: proxifyVector3(this.angularVelocity, entity),
       /** If multiplier is 0, ridigbody moves immediately to target pose, linearly interpolating between substeps */
-      targetKinematicLerpMultiplier: 0
+      targetKinematicLerpMultiplier: 0,
+      canSleep: true /** Not a reactive property. Only on Init*/
```
HexaField: since this is an API, it should be moved above the `// internal` comment

HexaField (changes requested):

packages/spatial/src/physics/components/RigidBodyComponent.ts
DanielBelmes: Oh Sorry I understood it incorrectly. Thought internal was everything bellow

DanielBelmes: Ohhhh. Nevermind. Double misunderstood.

PR #9963, opened 2024-03-21
Add SceneKillHeightSystem to handle killing entities below a certain height

### LINKED ISSUE #9910: Add kill floor height to scene settings
- [x] add kill floor number value to SceneSettingsComponent
- [x] In a new pre-physics simulation group system for the kill floor, iterate through each SceneSettingsComponent and then iterate the query for `[RigidbodyComponent, Not(RigidbodyFixedTagComponent), NetworkAuthorityTagComponent]` and if the rigidbody y value is below the kill floor, teleport the rigidbody back to it's SpawnPoseState position & rotation by setting & dirtying the TransfromComponent

SYBIOTE:
## Summary
Add SceneKillHeightSystem to handle killing entities below a certain height

closes #9910

## QA Steps
add dynamic rigid body to default scene in editor with collider.
use visual script to add network authority tag.
confirm that rigid body rests under the scene kill height.

### DIFF
--- a/packages/engine/src/scene/systems/SceneKillHeightSystem.tsx
+++ b/packages/engine/src/scene/systems/SceneKillHeightSystem.tsx
@@ -60,7 +60,6 @@ const execute = () => {
       // reset entity
       const uuid = getComponent(entity, UUIDComponent)
       const spawnState = getState(SpawnPoseState)[uuid]
-      console.log('reseting ', entity, spawnState?.spawnPosition)
       setComponent(entity, TransformComponent, {
         position: spawnState?.spawnPosition,
         rotation: spawnState?.spawnRotation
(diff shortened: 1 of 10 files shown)


packages/engine/src/scene/systems/SceneKillHeightSystem.tsx
```diff
@@ -0,0 +1,94 @@
...
+      const spawnState = getState(SpawnPoseState)[uuid]
+
+      setComponent(entity, TransformComponent, {
+        position: spawnState?.spawnPosition,
+        rotation: spawnState?.spawnRotation
+      })
+      TransformComponent.dirtyTransforms[entity] = true
+
+      const { cameraAttachedRigidbodyEntity } = getState(PhysicsState)
+      if (entity !== cameraAttachedRigidbodyEntity) return
```
HexaField: these returns need to be continues

HexaField (changes requested):

HexaField (approved):

PR #9965, opened 2024-03-22
9928 optimize asset builds

### LINKED ISSUE #9928: Optimize asset-only projects to not be downloaded for rebuild
A lot of our projects don't have any code. We should move the non-code APIs from the `xrengine.config.ts` to the `package.json` for the project, and only have an `xrengine.config.ts` for projects that have code assets. Following this, we can optimize the build process accordingly.

barankyle:
## References
closes #9928


packages/server-core/src/media/storageprovider/s3.storage.ts
```diff
@@ -303,14 +310,28 @@ export class S3Provider implements StorageProviderInterface {
...
+              : ObjectCannedACL.public_read,
           Body: Buffer.alloc(0),
           Bucket: this.bucket,
           ContentType: 'application/x-empty',
           Key: key + '/'
         }
       : {
-          ACL: ObjectCannedACL.public_read,
+          ACL:
+            projectRegex.test(key) &&
```
HexaField: can we make this a module level function to reuse across the few instances we already do, to avoid discrepancies & regressions in the future

HexaField (changes requested):

packages/server-core/src/media/storageprovider/s3.storage.ts
barankyle: This originates in #9345, this PR is based on that since I wanted to re-use some of the regexes I'd made in that one. I made it an exported function there.

HexaField (approved):

HexaField (changes requested):
We should move the info needed or asset projects in the package.json to a new manifest.json

Later on, this will host a lot more information, such as module imports and scene list

dinomut1 (approved):

HexaField:
I think this is good to go but will wait for @MoizAdnan @hanzlamateen to have the last say

hanzlamateen (approved):

PR #9970, opened 2024-03-22
Removed/moved some dependencies.

barankyle:
## Summary

Moved react-icons out of root package.json
Only needed in ui and client packages. Was being unnecessarily installed in api/instanceserver builds as a result.

Removed @etherealengine/volumetric package since all its functionality is now in the engine itself.

## References
closes #_insert number here_

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,6 @@
     "fix-esm": "^1.0.1",
     "minio": "7.1.3",
     "patch-package": "^6.5.1",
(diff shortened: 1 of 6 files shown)


HexaField (changes requested):
need to also remove @etherealengine/volumetric from vite config exclude

HexaField (changes requested):
still a build error

HexaField (approved):

PR #9985, opened 2024-03-26
remove `react-color` and use `react-colorful` as colorinput

### LINKED ISSUE #9977: Find modern replacement for react-color color picker
It has a sub-dependency that is 20MB

aditya-mitra:
## Summary

- Removed `react-color` from client-core and editor packages replacing with `@uiw/react-color`
- Also its bundling from vite.config
- `@uiw/react-color-sketch` has [smaller install size (195)](https://packagephobia.com/result?p=@uiw/react-color-sketchl) than `react-color`'s having 7 dependencies and [larger install size (9.09MB)](https://packagephobia.com/result?p=react-color)
- it also smaller install size and is [more actively maintained](https://github.com/uiwjs/react-color) than [`react-colorful`](https://github.com/omgovich/react-colorful)

## References
closes #9977

## QA Steps

image

image


HexaField (changes requested):
Works great, but let's add our own hex field underneath the color picker, with a copy button & text field like the old one had. This is super important for the studio. Also, let's disable scrolling and try to have the background transparent for the corners which appear white
Screenshot from 2024-03-26 12-05-39

HexaField (approved):

PR #9986, opened 2024-03-26
Remove Mediapipe/pose package

AidanCaruso:
## Summary
Removes the pose package, uses vision-tasks properties when available, and brings landmarks to indices map in-house.
## References
closes #_insert number here_


packages/engine/src/mocap/mocapUtils.ts
```diff
@@ -0,0 +1,60 @@
...
+The Original Code is Ethereal Engine.
+
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Ethereal Engine team.
+
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023 
+Ethereal Engine. All Rights Reserved.
+*/
+
+export const landmarkIndices = {
```
HexaField: let's rename this file `MocapConstants` and this const `LandmarkIndices` to keep the CapitalCase naming consistent, and the values should be UPPER_CASE

HexaField (changes requested):

HexaField (approved):

PR #10001, opened 2024-03-26
unify uuid everywhere

DanielBelmes:
## References
closes #_insert number here_


packages/client/package.json
```diff
@@ -71,7 +71,7 @@
     "tailwind-merge": "^1.13.2",
     "tailwindcss": "^3.3.2",
     "ts-node": "10.9.1",
-    "vite": "5.0.9",
+    "vite": "5.1.3",
```
DanielBelmes: gotta revert probably. But honestly we should update. There are security issues with 5.0.9

DanielBelmes:
@barankyle I changed some v1 uuid calls to v4 assuming that we just need random id's there. Idk if you do actually want to have machine specific influenced uuids.

barankyle:
> @barankyle I changed some v1 uuid calls to v4 assuming that we just need random id's there. Idk if you do actually want to have machine specific influenced uuids.

There wasn't any particular reason for v1 uuids, changing them to v4 should be fine. IIRC both v1 and v4 are the same format (32 hex characters split as <8>-<4>-<4>-<4>-<12>), so the schema we have for uuids should still apply.

barankyle:
Is the point of the new in-house generateEntityUUID function that it casts the UUID as an `EntityUUID`, and thus we don't have to remember to do that everywhere that's expecting an `EntityUUID`? Otherwise it seems like we could just use uuid.v4() everywhere.

DanielBelmes:
> Is the point of the new in-house generateEntityUUID function that it casts the UUID as an `EntityUUID`, and thus we don't have to remember to do that everywhere that's expecting an `EntityUUID`? Otherwise it seems like we could just use uuid.v4() everywhere.

I haven't linked the issue this is assigned to yet. That choice is a vestige of that. I think the idea was that we had a lot of places casting to "as EntityUUID" but it really wasn't as many as expected. I also should add some inline documentation of its purpose if we keep it. But yea I lean more towards not needing a new function. Thoughts @HexaField ?

packages/client/package.json
HexaField: let's do this in a separate PR and deploy to QAT

HexaField:
> I haven't linked the issue this is assigned to yet. That choice is a vestige of that. I think the idea was that we had a lot of places casting to "as EntityUUID" but it really wasn't as many as expected. I also should add some inline documentation of its purpose if we keep it. But yea I lean more towards not needing a new function. Thoughts @HexaField ?

That's fair enough. The idea is to be able to swap out the underlying EntityUUID generator more easily by having it abstracted to a contextualized function

barankyle (approved):

PR #10011, opened 2024-03-27
fix remove all components from erroring itself

DanielBelmes:
## Summary
Remove all component was failing silentlty. I ensured it would show errors in console. I also made sure that it would continue if it does error. Furthermore, I fixed the original bug in the object3d proxify as internal three code was triggering our throw. Now if a set to nil is attempted it will only print a warning and ignore.

## References
closes #_insert number here_


SYBIOTE (approved):
this fixes additional issues with the gizmo and debug mode as well, might potentially solve reparenting as well

HexaField (commented):
Can we test gltf export and scene switching

SYBIOTE:
switch scene in local dev doesn't causes any errors. cannot find export gltf option in the studio

HexaField:
closing in favour of https://github.com/EtherealEngine/etherealengine/pull/10014

PR #10012, opened 2024-03-28
fix infinite redirect for protected routes

### LINKED ISSUE #9961: [Bug]: Infinite redirect when going to studio or admin pages without required scopes
### Version

08c3b70(1.6.0)

### Operating Systems

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

### Browser (if applicable)

_No response_

If a user without admin scopes goes to an admin page, or any user without editor:write goes to /studio, they get stuck in an infinite redirect loop. 

### What happened?

If a user without editor:write goes to /studio, the useEffect on line 86 of client/src/pages/editor/editor.tsx is redirecting them to the root page with a redirectUrl of /studio, and then client-core/src/common/services/RouterService.ts is navigating them back to /studio at around line 117, which then goes back to root with the redirectUrl, etc. Similar thing with client-core/src/admin/adminRoutes.tsx, which on ~line 68 is sending users without the admin:admin scope back to root with a redirect to their current path, which then redirects them based on that redirectUrl, etc.

### Steps to reproduce

```shell
Make a user without editor:write or admin:admin scopes. Go to /studio and and of the /admin/<service> pages.
```

### Relevant Info

_No response_

aditya-mitra:
## Summary

- fix infinite redirect when a protected route is accessed by user without permission
- added the `history` package to navigate outside of react components
- remove the state usage in `RouterState` and directly navigate using `history.push()`
- also remove types package for `react-router-dom`

## References
closes #9961


HexaField:
We might be over engineering this. I'm wondering if a simple navigate is going to be a more robust solution, rather than handling our own navigation state.

HexaField (approved):
Looks good! I'll let someone else test and review but it works here.

barankyle (approved):

PR #10015, opened 2024-03-31
Fix loop animation regressions

AidanCaruso:
## Summary

- Fixes imported avatar animations requiring the rig to be named 'Armature', in favor of a hips lookup
- Fixes reactivity issues with animation packs
- Use VRM is enabled by default for VRMs and only exposed on non-VRM mixamo avatars.

## References
closes #_insert number here_


HexaField (changes requested):
in avatarTestByIndex i am seeing `Cannot read properties of undefined (reading 'rawHumanBones')`

HexaField (approved):

PR #10038, opened 2024-04-04
fix video and vec2/vec3 Scrub

### LINKED ISSUE #9939: [Bug]: VideoComponent no longer rendering video onto mesh plane
### Version

08c3b70(1.6.0)

### Operating Systems

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

### Browser (if applicable)

_No response_

Video is supposed to render onto mesh

### What happened?

Mesh appears black

### Steps to reproduce

```shell
Drag video from assets into scene
```

### Relevant Info

_No response_

DanielBelmes:
## Summary
Video was broken because of something weird in three.js

I also made vec2Scrub return the vector instead of the pointer event

## References
closes [#9939](https://github.com/EtherealEngine/etherealengine/issues/9939)

### DIFF
--- a/packages/editor/src/components/inputs/Vector2Input.tsx
+++ b/packages/editor/src/components/inputs/Vector2Input.tsx
@@ -93,7 +93,14 @@ export const Vector2Input = ({ value, onChange, onRelease, uniformScaling, hideL
         onRelease={onReleaseX}
         prefix={
           hideLabels ? null : (
-            <Vector3Scrubber {...rest} tag="div" value={vx} onChange={onChangeX} onPointerUp={onRelease} axis="x">
+            <Vector3Scrubber
+              {...rest}
+              tag="div"
+              value={vx}
+              onChange={onChangeX}
+              onPointerUp={(ev) => onRelease?.(value)}
+              axis="x"
+            >
               X
             </Vector3Scrubber>
           )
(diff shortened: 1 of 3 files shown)


packages/editor/src/components/inputs/Vector2Input.tsx
```diff
@@ -93,7 +93,14 @@ export const Vector2Input = ({ value, onChange, onRelease, uniformScaling, hideL
         onRelease={onReleaseX}
         prefix={
           hideLabels ? null : (
-            <Vector3Scrubber {...rest} tag="div" value={vx} onChange={onChangeX} onPointerUp={onRelease} axis="x">
+            <Vector3Scrubber
+              {...rest}
+              tag="div"
+              value={vx}
+              onChange={onChangeX}
+              onPointerUp={(ev) => onReleaseX(value.x)}
```
HexaField, line 101: this seems different to what is happening in the Vector3Input - if that one works, i wonder why this doesnt. we should try to standardize these

HexaField (dismissed):

packages/editor/src/components/inputs/Vector2Input.tsx
DanielBelmes: It doesn't work :) I'll fix that too

DanielBelmes: fixed that as well

MichaelEstes (approved):

PR #10042, opened 2024-04-04
Point & Click Placement Mode

dinomut1:
Adds point & click mode for asset placement


packages/engine/src/scene/functions/serializeWorld.ts
```diff
@@ -49,4 +58,17 @@ export const serializeEntity = (entity: Entity) => {
   return jsonComponents
 }
 
+export const toEntityJson = (entity: Entity) => {
```
HexaField, line 61: I'm worried about learning on serializing the simulation layer into the authoring layer, as it will be serializing simulation side effects (such as physics simulation, runtime entities, visual script mutations etc).

Rather than this, we should

packages/editor/src/systems/ClickPlacementSystem.ts
```diff
@@ -0,0 +1,194 @@
...
+      if (!clickState.selectedAsset.value) return
+      const parentEntity = getParentEntity()
+      if (!parentEntity) return
+      const placementEntity = clickState.placementEntity.value
+      if (!placementEntity) return
+      const sceneID = getComponent(parentEntity, SourceComponent)
+      setComponent(placementEntity, SourceComponent, sceneID)
+      const snapshot = SceneSnapshotState.cloneCurrentSnapshot(sceneID)
+      const uuid = getComponent(placementEntity, UUIDComponent)
+      snapshot.data.entities[uuid] = toEntityJson(placementEntity)
```
HexaField: what if instead of serializing this already deserialzied entity, we just load in the new data, it should only be the ModelComponent & TransformComponent that need to be set, right? We have the tooling for this already in EntityControlFunctions, which are supposed to be the only API for mutating the authoring layer

packages/editor/src/systems/ClickPlacementSystem.ts
```diff
@@ -0,0 +1,194 @@
...
+  insert: { before: TransformSystem },
+  reactor: () => {
+    Cache.enabled = true
+
+    const clickState = useState(getMutableState(ClickPlacementState))
+    const editorState = useState(getMutableState(EditorHelperState))
+    const engineState = useState(getMutableState(EngineState))
+    const sceneState = useState(getMutableState(SceneState))
+
+    const renderers = defineQuery([RendererComponent])
```
HexaField: this will define a new query every single time the reactor rerenders, either move this to module scope or use `useQuery`

packages/editor/src/systems/ClickPlacementSystem.ts
```diff
@@ -0,0 +1,194 @@
...
+      }
+      clickState.placementEntity.set(createPlacementEntity(parentEntity))
+    }
+
+    useEffect(() => {
+      const placementMode = editorState.placementMode.value
+      const renderer = getComponent(renderers()[0], RendererComponent)
+      const canvas = renderer.canvas
+      if (placementMode === PlacementMode.CLICK) {
+        canvas.addEventListener('click', clickListener)
```
HexaField: we need to use our spatial input abstractions instead of listener events. This also is not going to be referencing the same instance of the click listener

packages/editor/src/systems/ClickPlacementSystem.ts
```diff
@@ -0,0 +1,194 @@
...
+      dispatchAction(SceneSnapshotAction.createSnapshot(snapshot))
+      if (getState(ObjectGridSnapState).enabled) {
+        getMutableState(ObjectGridSnapState).apply.set(true)
+      }
+      clickState.placementEntity.set(createPlacementEntity(parentEntity))
+    }
+
+    useEffect(() => {
+      const placementMode = editorState.placementMode.value
+      const renderer = getComponent(renderers()[0], RendererComponent)
```
HexaField: we can't just get use the first renderer from this query, we need to either use `Engine.instance.viewerEntity` or support this functionality across multiple render contexts (see `CameraOrbitSystem.ts`)

packages/editor/src/systems/ClickPlacementSystem.ts
```diff
@@ -0,0 +1,194 @@
...
+    return null
+  },
+  execute: () => {
+    const editorHelperState = getState(EditorHelperState)
+    if (editorHelperState.placementMode !== PlacementMode.CLICK) return
+    const clickState = getState(ClickPlacementState)
+    const placementEntity = clickState.placementEntity
+    if (!placementEntity) return
+
+    const objectLayerQuery = defineQuery([ObjectLayerComponents[ObjectLayers.Scene]])
```
HexaField: we must not define new queries each frame

packages/editor/src/systems/ClickPlacementSystem.ts
```diff
@@ -0,0 +1,194 @@
...
+    //const sceneObjects = Array.from(Engine.instance.objectLayerList[ObjectLayers.Scene] || [])
+    const camera = getComponent(Engine.instance.cameraEntity, CameraComponent)
+    const pointerScreenRaycaster = new Raycaster()
+
+    const mouseEntity = InputPointerComponent.getPointerForCanvas(Engine.instance.viewerEntity)
+    if (!mouseEntity) return
+    const mouse = getComponent(mouseEntity, InputPointerComponent).position
+    pointerScreenRaycaster.setFromCamera(mouse, camera) // Assuming 'camera' is your Three.js camera
+
+    pointerScreenRaycaster.setFromCamera(mouse, camera) // Assuming 'camera' is your Three.js camera
```
HexaField: duplicate call

HexaField (changes requested):

HexaField:
Screenshot from 2024-04-05 11-31-01
UI is also a bit busted

packages/editor/src/systems/ClickPlacementSystem.tsx
```diff
@@ -0,0 +1,280 @@
...
+
+export const ClickPlacementSystem = defineSystem({
+  uuid: 'ee.studio.ClickPlacementSystem',
+  insert: { before: TransformSystem },
+  reactor: () => {
+    const clickState = useState(getMutableState(ClickPlacementState))
+    const editorState = useState(getMutableState(EditorHelperState))
+    const sceneState = useState(getMutableState(SceneState))
+
+    const renderers = defineQuery([RendererComponent])
```
HexaField: this needs to be useQuery

packages/editor/src/systems/ClickPlacementSystem.tsx
```diff
@@ -0,0 +1,280 @@
...
+
+    const getParentEntity = () => {
+      if (!sceneState.scenes.value) return null
+      const sceneID = sceneState.scenes.keys[0] as SceneID
```
HexaField: why are we implicitly getting index 0 here? we must be getting the scene explicitly (and for multi-scene support)

packages/editor/src/systems/ClickPlacementSystem.tsx
```diff
@@ -0,0 +1,280 @@
...
+    const clickListener = () => {
+      if (!clickState.selectedAsset.value) return
+      const parentEntity = getParentEntity()
+      if (!parentEntity) return
+      const placementEntity = clickState.placementEntity.value
+      if (!placementEntity) return
+      const sceneID = getComponent(parentEntity, SourceComponent)
+      setComponent(placementEntity, SourceComponent, sceneID)
+      setComponent(placementEntity, EntityTreeComponent, { parentEntity })
+      const snapshot = SceneSnapshotState.cloneCurrentSnapshot(sceneID)
```
HexaField: can we move this snapshot manipulation logic to EditorControlFunctions to keep it in the same context as the other ones?

packages/editor/src/systems/ClickPlacementSystem.tsx
```diff
@@ -0,0 +1,280 @@
...
+        mesh.material = material
+      }
+      clickState.materialCache.set([])
+    }
+
+    useEffect(() => {
+      const placementMode = editorState.placementMode.value
+      const renderer = getComponent(renderers()[0], RendererComponent)
+      const canvas = renderer.canvas
+      if (placementMode === PlacementMode.CLICK) {
```
HexaField: use spatial input instead of event listeners, also, this does not clean up the click listener as a different instance of the function will be created each time the reactor rerenders

packages/editor/src/systems/ClickPlacementSystem.tsx
```diff
@@ -0,0 +1,280 @@
...
+    const editorHelperState = getState(EditorHelperState)
+    if (editorHelperState.placementMode !== PlacementMode.CLICK) return
+    const clickState = getState(ClickPlacementState)
+    const placementEntity = clickState.placementEntity
+    if (!placementEntity) return
+    const buttons = InputSourceComponent.getMergedButtons()
+    const sceneObjects = objectLayerQuery().flatMap((entity) => getComponent(entity, GroupComponent))
+    //const sceneObjects = Array.from(Engine.instance.objectLayerList[ObjectLayers.Scene] || [])
+    const camera = getComponent(Engine.instance.cameraEntity, CameraComponent)
+    const pointerScreenRaycaster = new Raycaster()
```
HexaField, line 257: we should be using spatial input here, rather than a custom raycaster

HexaField (dismissed):

packages/engine/src/scene/functions/serializeWorld.ts
dinomut1: Yes if this function is used incorrectly it could end up capturing side effects, but there are many cases where just fully serializing an entity is the correct course of action. We should have this function available for those cases. I'm using it in the click placement system to add the placement entity to the authoring layer, since it does (and should) only exist in the simulation layer up until the click placement event. In that case, the entity is guaranteed to only be a model component and a transform component which should be serialized directly from their runtime values, and this function is correctly called in that case.

packages/editor/src/systems/ClickPlacementSystem.tsx
dinomut1: Yes this specific logic doesn't yet exist in the current editor control functions, since we are both changing the scene of the entity and adding it to the authoring layer at the same time

MichaelEstes (approved):

HexaField:
failing checks & tests

dinomut1 (commented):
Still seeing bugs around the geometry snapping to the wrong place on click, and after a few clicks still getting a breaking error around generate bounding sphere

dinomut1:
[Screencast from 05-31-2024 01:59:19 PM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/ff72b9d4-b51e-47cc-aeda-7d0a612830de)
Looks like after placement there are still some frames where the bounding box starts raycasting with its own geometry and that offsets where the snap actually occurs

packages/editor/src/components/toolbar/tools/TransformTool.tsx
```diff
@@ -26,14 +26,18 @@ Ethereal Engine. All Rights Reserved.
 import HeightIcon from '@mui/icons-material/Height'
 import OpenWithIcon from '@mui/icons-material/OpenWith'
 import SyncIcon from '@mui/icons-material/Sync'
+
+import AdsClickIcon from '@mui/icons-material/AdsClick'
```
SYBIOTE, line 30: can we use equivalent react icons, we plan to be mui free

SYBIOTE (commented):
the changes are to the old studio for the file browser panel, we would have to transfer them to the new studio later, so i suggest we implement them in the new studio now

PR #10043, opened 2024-04-04
ECS Materials

AidanCaruso:
## Summary

- Replaces material library hookstate object with components on the ECS

- Material Component is used on every entity that uses a material

  - Material component is broken up into MaterialInstance, MaterialState, MaterialPrototype and MaterialPlugin

  - MaterialInstance is placed on entities with GroupComponents that use materials, and contains UUIDs that point to material entities containing state and relevant data

  - MaterialState contains the material state, pointer to the MaterialPrototype and MaterialPlugin entities, and a list of all instance entities using the material

  - MaterialPrototype contains the material's prototype constructor, name and arguments

  - MaterialPlugin contains the OBC plugin and its parameters

  - Material names and sources are hashed to prevent duplicate materials from the use of duplicate assets

- Material names are unique across the entire scene, if a material name is in use, a numerical suffix is added to any subsiquent duplicate names

 - Setting material prototypes and plugins are done entirely through the ECS via setComponent, reactors propagate these changes to the threejs material state

- Material library studio panel accomodates these changes

  - Global material view to see all materials in a scene by querying material entities with material state

  - Local material view to see materials used by a selected entity by querying material entities associated with the selected entity

- Material instances are now automatically removed by the material component's onRemove call if it's a MaterialInstance

- Moves any material system code that does not make asset sourcing references to spatial package

## References
closes #_insert number here_


packages/engine/src/scene/materials/components/MaterialComponent.ts
```diff
@@ -23,21 +23,47 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
   parameters: { [field: string]: any }
   plugins: string[]
   src: MaterialSource
   status: MaterialStatus
   instances: Entity[]
 }
+
+export const MaterialComponent = defineComponent({
+  name: 'MaterialComponent',
+  jsonID: 'EE_material',
```
HexaField: we don't want this to be serializable, therefore it should not have a jsonID

packages/spatial/src/renderer/components/GroupComponent.tsx
```diff
@@ -103,6 +116,36 @@ export function addObjectToGroup(entity: Entity, object: Object3D) {
   proxifyVector3WithDirty(TransformComponent.position, entity, TransformComponent.dirtyTransforms, obj.position)
   proxifyQuaternionWithDirty(TransformComponent.rotation, entity, TransformComponent.dirtyTransforms, obj.quaternion)
   proxifyVector3WithDirty(TransformComponent.scale, entity, TransformComponent.dirtyTransforms, obj.scale)
+
+  if (!(object as Mesh).material) return
```
HexaField: this needs to be moved elsewhere, as it is references SourceComponent and other engine package imports/concepts

moreso, we probably want to move some of the material state, or create new and cleaner ones in the spatial package, according to our engine modularity hierarchy. anything pertaining to rendering should be in the spatial package, and anything pertaining to sources and assets should be in the engine package

HexaField (changes requested):

PR #10044, opened 2024-04-05
Break Up and Type GLTF Loader

### LINKED ISSUE #10050: Split up GLTFLoader into multiple typed files

HexaField:
## Summary

Breaks up the threejs GLTFLoader into typed files (found a full-blown bug!) for easier comprehension and compartmentalisation.

## References
closes #10050


packages/engine/src/scene/systems/EnvironmentSystem.tsx
```diff
@@ -44,6 +45,8 @@ const EnvmapReactor = (props: { backgroundEntity: Entity }) => {
   const groupComponent = useComponent(entity, GroupComponent)
 
   useEffect(() => {
+    // TODO use spatial queries
+    if (!haveCommonAncestor(entity, props.backgroundEntity)) return
```
DanielBelmes, line 49: What bug was this fixing?

HexaField: I forgot to include this in a previous PR. This ensures that the skybox only affects the envmaps of the entities in the scene it is loaded in, essentially preventing disparate scenes from affecting eachother

PR #10045, opened 2024-04-05
Add GLTF Scene Support

HexaField:
## Summary

- adds support for loading a .gltf file as a scene instead of a .scene.json
- replaces 'scene' service with 'asset' table
- replaces 'portal' service 'spawn-point' table (unused currently, breaks portals, requires a lot more work)
- scenes are referenced by file path instead of sending json over feathers
- fixes material and uuid gltf export/import problems preventing some scenes from being exported
- fixes multi-scene support for PostprocessingComponent, SDFComponent, CSM & RendererSettingsComponent
- defines a new component for each child entity added to a SceneComponent children list
- adds a bunch of complex reactive ECS query helper functions

## References
closes #_insert number here_


packages/engine/src/scene/functions/loadGLTFModel.ts
```diff
@@ -220,6 +223,8 @@ export const proxifyParentChildRelationships = (obj: Object3D) => {
 }
 
 export const generateEntityJsonFromObject = (rootEntity: Entity, obj: Object3D, entityJson?: EntityJsonType) => {
+  if (!obj.uuid) throw new Error('Object3D must have a UUID')
+  console.log('uuid', obj.uuid)
```
AidanCaruso: Clean up logs?

the-reality-engineer (commented):
Huge step forward. Can we add test coverage for this? Perhaps keep an exported gltf somewhere (default project) that we can load?

MoizAdnan (changes requested):
I have been QAing the studio, and found these issues:

1.  `Create Scene` button throws this error and no scene is created: 
```GeneralError: insert into `asset` (`createdAt`, `id`, `projectId`, `updatedAt`) values ('2024-04-18 06:33:54', 'fa97c3fc-e632-4b06-a475-127685a3b633', '7bce813a-1e14-44f0-bc4a-8f75f5d11ea4', '2024-04-18 06:33:54') - ER_NO_DEFAULT_FOR_FIELD: Field 'name' doesn't have a default value```

2. `Rename Scene` works fine but it doesn't change the filename in the file storage or the `assetURL` I dont know whether this is intended or not.
3. `Delete Scene` throws this error and doesn't work: `sceneFunctions.tsx:55 Error: Cannot read properties of undefined (reading 'toString')`

Other than this looks great!

File: packages/editor/src/components/EditorContainer.tsx [hunk 2]
```diff
@@ -177,12 +178,12 @@ const onSaveAs = async () => {
...
       if (result?.name && projectName) {
-        await saveScene(projectName, result.name, abortController.signal)
+        await saveSceneJSON(sceneAssetID, projectName, result.name, abortController.signal)
         getMutableState(SceneState).sceneModified.set(false)
-        const newSceneData = (await Engine.instance.api
+        const [newSceneData] = (await Engine.instance.api
           .service(scenePath)
-          .get('', { query: { project: projectName, name: result.name, metadataOnly: true } })) as SceneDataType
-        setSceneInState(newSceneData.scenePath)
+          .find({ query: { assetURL: getState(EditorState).scenePath! } })) as any as AssetDataType[]
```
hanzlamateen: Wouldnt it be a Paginated object? Because if you want an array, you need to pass `paginate: false` in query

hanzlamateen (changes requested):

packages/editor/src/components/EditorContainer.tsx
HexaField: good catch!

PR #10055, opened 2024-04-12
Added support for role ARN usage in aws CLI in builder

barankyle:
## Summary

AWS SDK is now set up to get credentials from a credentials file, instead of pulling public/secret keys directly from the config. This allows for assumed roles to be used, as that can be configured in the credentials file. The S3 storage provider writes the credentials file on server start, getting values from appconfig.

Fixed some spacing bugs in bash files where closing brackets were not spaced properly from double quotes.

## References
closes #_insert number here_


hanzlamateen (approved):
Looks good

PR #10058, opened 2024-04-12
9997 cdn invalidation fix

### LINKED ISSUE #9997: [Bug]: On Deployment: Saving a file with the same name as an existing file does not invalidate cloudfront cache
### Version

08c3b70(1.6.0)

### Operating Systems

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] Android
- [ ] IOS

### Browser (if applicable)

_No response_

When a file is saved with the same name as an existing file, it should overwrite the file, and the url pointing to that file should now point to the newly saved file

### What happened?

Saving a file with the same name as an existing file does not update the cloudfront url. This forces the user to save every file with a unique name to avoid the bug.

### Steps to reproduce

```shell
1. On deployment, place a 3d model into a scene
2. Modify that model in some way, then save the changes using the model component export. Make sure to keep the file name the same
3. The model component will immediately try to reload that URL, and it will load the old file. Dragging and dropping the file will also produce the same old file
```

### Relevant Info

90% sure this is a cloudfront cache invalidation problem. But it may be deeper problem with S3 itself

barankyle:
## References
closes #9997 

Linked to EtherealEngine/ethereal-engine-ops/pull/13


packages/server-core/src/media/invalidation/invalidation.docs.ts
```diff
@@ -0,0 +1,44 @@
...
+} from '@etherealengine/common/src/schemas/media/invalidation.schema'
+
+export default createSwaggerServiceOptions({
+  schemas: {
+    invalidationDataSchema,
+    invalidationQuerySchema,
+    invalidationSchema
+  },
+  docs: {
+    description: 'Static resource service description',
```
hanzlamateen: Description need to be updated

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -207,7 +210,12 @@ export class FileBrowserService
     const fileName = await getIncrementalName(data.newName, _newPath, storageProvider, isDirectory)
     const result = await storageProvider.moveObject(data.oldName, fileName, _oldPath, _newPath, data.isCopy)
 
-    await storageProvider.createInvalidation([_oldPath, _newPath])
+    await this.app.service(invalidationPath).create({
```
hanzlamateen: I think this can be done in a single call as

```
await this.app.service(invalidationPath).create([{
  path: _oldPath
},{
  path: _newPath
}])
```

packages/server-core/src/media/invalidation/migrations/20240405000547_invalidation.ts
```diff
@@ -0,0 +1,65 @@
...
+
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023 
+Ethereal Engine. All Rights Reserved.
+*/
+
+import { invalidationPath } from '@etherealengine/common/src/schemas/media/invalidation.schema'
+import type { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  const oldTableName = 'static_resource'
```
hanzlamateen: We dont need this const oldTableName

packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -1634,14 +1637,20 @@ export const getCommitSHADate = async (projectName: string): Promise<{ commitSHA
...
+export const deleteProjectFilesInStorageProvider = async (
+  app: Application,
+  projectName: string,
+  storageProviderName?: string
+) => {
   const storageProvider = getStorageProvider(storageProviderName)
   try {
     const existingFiles = await getFileKeysRecursive(`projects/${projectName}`)
     if (existingFiles.length) {
       await Promise.all([
```
hanzlamateen: I think this should not be parallel rather first it should async delete resource and then add the entry to invalidation service.

From:
```
await Promise.all([
  storageProvider.deleteResources(existingFiles),
  app.service(invalidationPath).create({
    path: `projects/${projectName}*`
  })
])
```

To:
```
await storageProvider.deleteResources(existingFiles),
await app.service(invalidationPath).create({
 path: `projects/${projectName}*`
})

```

packages/server-core/src/projects/scene/scene.hooks.ts
```diff
@@ -217,8 +218,12 @@ const createSceneInStorageProvider = async (context: HookContext<SceneService>)
     )
   )
   try {
-    await context.storageProvider.createInvalidation(
-      SCENE_ASSET_FILES.map((asset) => `${data.directory}${context.newSceneName}${asset}`)
+    await Promise.all(
```
hanzlamateen: Instead of making multiple requests to invalidation service. Pass array of {path: ''} object in 1 create call.

packages/server-core/src/projects/scene/scene.hooks.ts
```diff
@@ -287,7 +292,12 @@ const renameSceneInStorageProvider = async (context: HookContext<SceneService>)
     if (await storageProvider.doesExist(oldSceneJsonName, directory!)) {
       await storageProvider.moveObject(oldSceneJsonName, newSceneJsonName, directory!, directory!)
       try {
-        await storageProvider.createInvalidation([directory + oldSceneJsonName, directory + newSceneJsonName])
+        await context.app.service(invalidationPath).create({
```
hanzlamateen: Can be done in single call instead of multiple calls.

packages/server-core/src/projects/scene/scene.hooks.ts
```diff
@@ -393,7 +403,13 @@ const saveSceneInStorageProvider = async (context: HookContext<SceneService>) =>
   }
 
   try {
-    await storageProvider.createInvalidation(SCENE_ASSET_FILES.map((asset) => `${directory}${name}${asset}`))
+    await Promise.all(
```
hanzlamateen: Can be done in single call.

packages/server-core/src/projects/scene/scene.hooks.ts
```diff
@@ -493,7 +509,13 @@ const removeSceneInStorageProvider = async (context: HookContext<SceneService>)
   await storageProvider.deleteResources(SCENE_ASSET_FILES.map((ext) => `${directory}${sceneName}${ext}`))
 
   try {
-    await storageProvider.createInvalidation(SCENE_ASSET_FILES.map((asset) => `${directory}${sceneName}${asset}`))
+    await Promise.all(
```
hanzlamateen: Can be done in single call.

hanzlamateen (changes requested):

hanzlamateen (approved):

PR #10071, opened 2024-04-16
Improve mac os onboarding, disable minio, and update mediasoup

DanielBelmes:
## References
closes #_insert number here_


HexaField (changes requested):
Unfortunately changing the storage provider breaks our server tests in CI/CD. We need to fix this before merging this PR.

DanielBelmes:
> Unfortunately changing the storage provider breaks our server tests in CI/CD. We need to fix this before merging this PR.

Right!!!! I couldn't remember why we didn't do it. Maybe I'll spend a little time fixing that. It's worth it. If I can find a. Quick solution.

barankyle (approved):

HexaField (approved):

PR #10075, opened 2024-04-18
[UI/UX] Studio Update - Tailwind components for studio

aditya-mitra:
## Summary

the new studio is present in `/studio` route
and the old studio is present in `/studio-old` route

## References
closes #_insert number here_


packages/editor/tsconfig.json
```diff
@@ -37,5 +37,5 @@
     "../server-core/src/",
     "./**/*.ts",
     "./**/*.tsx"
-  ]
+, "../ui/src/components/editor/Vector3Input.tsx"  ]
```
HexaField: tsconfig likes to try to fix imports for us sometimes lol

HexaField (changes requested):

packages/editor/src/components/Editor2Container.tsx
```diff
@@ -0,0 +1,447 @@
...
+  if (projectName) {
+    try {
+      await inputFileWithAddToScene({ projectName })
+    } catch (err) {
+      NotificationService.dispatchNotify(err.message, { variant: 'error' })
+    }
+  }
+}
+
+const onSaveScene = async () => {
```
HexaField: let's export all these functions from the original file so we arent duplicating them

HexaField (changes requested):

packages/editor/src/components/Editor2Container.tsx
SYBIOTE: done, we will eventually delete the old route so they will eventually converge to the same file

packages/editor/tailwind.config.js
```diff
@@ -0,0 +1,76 @@
...
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Ethereal Engine team.
+
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023 
+Ethereal Engine. All Rights Reserved.
+*/
+
+/** @type {import('tailwindcss').Config} */
+
+module.exports = {
```
HexaField: can we move all the tailwind configs to the root of the monorepo, and import them to the packages where they are needed. this way we can ensure they are all always the same.

HexaField (dismissed):

packages/editor/tailwind.config.js
aditya-mitra: Done in https://github.com/EtherealEngine/etherealengine/pull/10237

DanielBelmes:
Was QAing looking good. Three things I saw.

- We have removed useAllComponents and replaced propagating the properties from the authoring layer. 
```ts
  const componentEditors = useHookstate(getMutableState(ComponentEditorsState)).get(NO_PROXY)
  const node = useHookstate(GLTFNodeState.getMutableNode(entity))
  const components: Component[] = []
  for (const jsonID of Object.keys(node.extensions.value!)) {
    const component = ComponentJSONIDMap.get(jsonID)!
    if (!componentEditors[component.name]) continue
    components.push(component)
  }
```
- There's no way to create a prefab anymore(as far as I found). We should get an updated design for where and how we want to make prefabs.
- Lastly, I think since we don't have the simple version yet we should hide the (Simple, Advanced) toggle. Unless we do have it and the button is just broken :) Feel free to correct me.

dinomut1 (dismissed):
studio-old bugs:

Add Entity dropdown broken

Add Component dropdown broken

File browser settings: 
	-cannot close modal, locks camera while open
	-slider doesn't work
	-generate thumbnails button doesn't work

Top left dropdown:
	-only "New Scene" button works (very difficult to click on, only successful once). Clicks through every other option
	-dropdown doesn't close, locks camera while open

loop animation node editor crashes studio

DanielBelmes:
> studio-old bugs:
> 
> Add Entity dropdown broken
> 
> Add Component dropdown broken
> 
> File browser settings: -cannot close modal, locks camera while open -slider doesn't work -generate thumbnails button doesn't work
> 
> Top left dropdown: -only "New Scene" button works (very difficult to click on, only successful once). Clicks through every other option -dropdown doesn't close, locks camera while open
> 
> loop animation node editor crashes studio

RESOLVED

DanielBelmes (approved):

PR #10077, opened 2024-04-18
Modal fixes & UI improvements

CITIZENDOT:
## Summary
- Fix Button alignment issue in `Add Project` Modal
- Fix `CircleCheck` icon alignment issue in `Add Project` Modal
- Set `submitLoading` correctly on `Add Project` Modal
- Close `Update Engine` modal, after successful update.

## References
closes #_insert number here_


packages/ui/src/components/tailwind/PasswordInput/index.tsx
```diff
@@ -43,7 +43,7 @@ const PasswordInput = forwardRef<HTMLInputElement, InputProps>(
         {...props}
         type={show.value ? 'text' : 'password'}
         endComponent={
-          <Button onClick={toggleShow} className="bg-transperant pointer-events-auto px-0">
+          <Button onClick={toggleShow} className="bg-transperant pointer-events-auto mr-3.5 px-0">
```
aditya-mitra, line 46: why did we put an `mr` here?
I checked on this branch on local but did not notice any differences.

aditya-mitra (changes requested):
Looks good to me!
Just need to look at 1 review.

packages/ui/src/components/tailwind/PasswordInput/index.tsx
CITIZENDOT: There was padding defined on endComponent's parent div: https://github.com/EtherealEngine/etherealengine/pull/10077/files#diff-536e7deed9813bd08522fd75316c4bb3c93ddddb460b83a954f0edeb878dcaa7

If padding is defined on parent component, it's not possible to control the space on the right. (In case we don't want to have the space at the right).

For example here: https://github.com/EtherealEngine/etherealengine/pull/10077/files#diff-948aac0b79362811bad963c4947160d968be29a46baa0f0540153e43dcec520c

Here endComponent should not have any space at the right. We cannot achieve it by setting spacing on parent. Hence margin-right directly on endComponent. (We can control this space directly through endComponent)

aditya-mitra (approved):

hanzlamateen (approved):

PR #10078, opened 2024-04-18
fix(magic-link): Fix guest identity provider removal upon login token validation and retain only the latest generated token

kattsushi:
## Summary

This PR addresses two key issues:

Ensures that upon validation of a login token, the guest identity provider is removed from the user. After setup, the identity-provider table will contain only one entry for users logging in via SSO, with any previous guest entries being deleted. However, for email logins, the identity-provider table incorrectly retains the guest entry, which is now resolved to remove the guest entry once the user logs in using a magic link.

Resolves the problem of duplicate tokens being generated for the same user in the login-token service. Now, when a duplicate token is generated, the previously generated token is properly removed.

## References
closes [#IR-1290](https://tsu.atlassian.net/browse/IR-1290)


kattsushi:
@cla-bot check

hanzlamateen (changes requested):
I still see the guest identity provider there for email after I verified it using magic link as you can see in row 2 and 3 of below image. Though when I logged it with linkedin SSO the guest entry for it was deleted.

image

kattsushi:
> I still see the guest identity provider there for email after I verified it using magic link as you can see in row 2 and 3 of below image. Though when I logged it with linkedin SSO the guest entry for it was deleted.
> 
> image

thanks for your comments and help @hanzlamateen this is already fixed

Screenshot 2024-04-24 at 9 43 52

packages/server-core/src/user/magic-link/magic-link.class.ts
```diff
@@ -6,8 +6,8 @@ Version 1.0. (the "License"); you may not use this file except in compliance
 with the License. You may obtain a copy of the License at
```
hanzlamateen, line 6: It seems like currently, it removing guest identity-provider when a magic link is send. This is not correct. The guest identity provider should be removed when a user login using url that he got in his email.

Please have a look of chat we had regarding making these changes in `login.class.ts`

hanzlamateen (changes requested):

hanzlamateen:
@kattsushi I just tried and its still now working after I login using magic link. I see 2 entries for same user as in below image

Please share video recording next time and show database entries as well.

image

hanzlamateen (approved):

PR #10085, opened 2024-04-21
UVOL Refactor

CITIZENDOT:
## Subtasks Checklist

- [x] Playlist Component
- [x] BufferDataContainer to track buffered/pending data
- [x] NodeEditor & basic pause/play logic
- [x] Geometry Buffering
- [x] Texture Buffering
  - [x] KTX2 Buffering
  - [x] Video Texture (needs QA)
- [x] Update geometry
  - [x] Render geometry
  - [x] Dispose previous geometries
- [x] Update textures
  - [x] Render textures
  - [x] Dispose previous textures
- [x] Autotune targets (geometry & textures)
- [ ] QA

## References
closes #_insert number here_

## QA Steps
image
PlaylistComponent :arrow_up:


DanielBelmes (changes requested):
- None of the node editors(new Uvol and Playlist) are updating snapshots/authoring layer so they are not being saved.
- Can't remove playlist component for some reason. No errors
- Changing playlist playmode is erroring (`TypeError: Cannot set properties of undefined (setting 'playMode') in EditorControlFunctions`). 
Other than those. Playback looks good

CITIZENDOT:
Fixed these.

But PlaylistComponent is supposed to be set on the VolumetricComponent at all times. Should we enforce it somehow? May be, deny remove when Volumetric is set on the entity?

DanielBelmes (approved):

PR #10096, opened 2024-04-25
IR-1614 adding proper debug labels in the SystemDebug rootsystems

SamMazerIR:
## Summary
adding proper debug labels in the SystemDebug menu for the 4 root systems

## QA Steps
open debug menu (with ~) and click the systems button, note that the root objects now display their system names rather than "root"


AidanCaruso (approved):
This is useful, not a big fan of magic string checking for "root", not sure of a better way to check

PR #10099, opened 2024-04-27
Add Feature Flag System

HexaField:
## References
closes #_insert number here_


packages/common/src/schemas/setting/feature-flag-setting.schema.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const featureFlagSettingMethods = ['find', 'get', 'create', 'patch', 'remove'] as const
+
+// Main data model schema
+export const featureFlagSettingSchema = Type.Object(
+  {
+    id: Type.String({
+      format: 'uuid'
+    }),
+    flags: Type.Any(),
```
hanzlamateen: I think we should have flagName and flagValue in db table. And each flag should be individual row.

hanzlamateen (changes requested):

hanzlamateen (approved):

PR #10102, opened 2024-04-29
GLTF ECS Loader

HexaField:
## Summary

Adds support to deserialize nodes & extensions from GLTF files instead of loading them with the threejs loader.

## Subtasks Checklist

- [x] New GLTF loader that for now supports just nodes and extensions
- [x] Fix Loading UI System
- [x] Drop first class support for SceneState, deprecating the APIs

## References
closes #_insert number here_


packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -0,0 +1,192 @@
...
+    const abortController = new AbortController()
+    const signal = abortController.signal
+
+    const loader = new FileLoader()
+
+    loader.setResponseType('arraybuffer')
+    loader.setRequestHeader({})
+    loader.setWithCredentials(false)
+
+    loader.load(
```
DanielBelmes, line 139: Resource Loader hooks?

packages/engine/src/scene/components/ColliderComponent.ts
```diff
@@ -177,7 +166,7 @@ export const ColliderComponent = defineComponent({
     const groupComponent = useOptionalComponent(entity, GroupComponent)
 
     useLayoutEffect(() => {
-      SceneAssetPendingTagComponent.removeResource(entity, ColliderComponent.jsonID)
+      setComponent(entity, InputComponent)
```
DanielBelmes, line 169: Pointing out as it's out of the norm. Assuming we're making an assumption that we want Colliders to always coexist with Input.

packages/spatial/src/renderer/components/DirectionalLightComponent.ts
```diff
@@ -154,8 +154,7 @@ export const DirectionalLightComponent = defineComponent({
       cameraFar: component.cameraFar.value,
       castShadow: component.castShadow.value,
       shadowBias: component.shadowBias.value,
-      shadowRadius: component.shadowRadius.value,
-      helper: component.helper.value
```
DanielBelmes, line 158: ooof nice find

packages/ecs/src/UUIDComponent.ts
```diff
@@ -82,6 +82,10 @@ export const UUIDComponent = defineComponent({
       setComponent(entity, UUIDComponent, uuid)
     }
     return state.value
+  },
+
+  generateUUID() {
+    return crypto.randomUUID() as EntityUUID
```
DanielBelmes: Can we revert using this since we've found issues with Node. We can still move generateEntityUUID to here but we should use 'uuid'

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -294,10 +266,10 @@ const createObjectFromSceneElement = (
     componentJson.find((comp) => comp.name === UUIDComponent.jsonID)?.props.uuid ?? generateEntityUUID()
 
   for (const [sceneID, entities] of Object.entries(scenes)) {
-    // gltf loader sanitizers names...
-    const name = getState(GLTFSourceState)[sceneID] ? 'New_Object' : 'New Object'
+    const name = 'New Object'
     if (getState(GLTFSourceState)[sceneID]) {
       const gltf = GLTFSnapshotState.cloneCurrentSnapshot(sceneID)
+      console.log(gltf, gltf.data.nodes)
```
DanielBelmes: leftover log

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -311,22 +283,35 @@ const createObjectFromSceneElement = (
...
+      } as GLTF.INode
+
+      gltf.data.nodes!.push(node)
+
+      if (extensions[TransformComponent.jsonID]) {
+        const comp = {
+          ...componentJsonDefaults(TransformComponent),
+          ...extensions[TransformComponent.jsonID]
+        }
+        const matrix = new Matrix4().compose(
```
DanielBelmes: optimize later

packages/engine/src/gltf/GLTFComponent.tsx
HexaField: Not yet, it will detect and load it through the threejs loader

packages/engine/src/scene/components/ColliderComponent.ts
HexaField: Yeah we need to figure something out here, probably a reactor to do this instead of in the deserializer

packages/ecs/src/UUIDComponent.ts
HexaField: Yes good point

DanielBelmes (approved):

DanielBelmes (approved):

PR #10104, opened 2024-04-30
fmt: Replace `v4` from 'uuid' with builtin `self.crypto.randomUUID()`

heysokam:
## Subtasks Checklist
- [x] Replace `v4` with `uuidv4`
- [x] Replace `uuidv4()` with `self.crypto.randomUUID()`
- [x] Remove all `from 'uuid'` imports
- [x] Remove `self.` prefix from `self.crypto`
- [x] Change `generateEntityUUID()` to use `crypto.randomUUID()`
- [x] `UUIDComponent.generateUUID()` return the type `EntityUUID`
- [x] Everywhere else use `crypto.randomUUID()` directly 
- [x] Add `generateEntityUUID()` to `UUIDComponent` as `UUIDComponent.generateUUID()`
- [x] Refactor every `generateEntityUUID()` call to use `UUIDComponent.generateUUID()`
- [x] Remove `generateEntityUUID()`
- [x] Remove leftover calls to `MathUtils.generateUUID()` _(two of them)_
- [x] Find `as EntityUUID` and replace with `UUIDComponent.generateUUID()`
- [ ] Fix `crypto is not defined` crash on `npm run test`
- [ ] Remove the `'uuid'` library/package from the list of installed packages

## References
closes #_insert number here_

## QA Steps

### DIFF
--- a/packages/client-core/src/common/components/AvatarPreview/index.tsx
+++ b/packages/client-core/src/common/components/AvatarPreview/index.tsx
@@ -37,7 +37,7 @@ import { SxProps, Theme } from '@mui/material/styles'
 
 import styles from './index.module.scss'
 
-import { createEntity, generateEntityUUID, setComponent, UndefinedEntity, UUIDComponent } from '@etherealengine/ecs'
+import { createEntity, setComponent, UndefinedEntity, UUIDComponent } from '@etherealengine/ecs'
 import { defaultAnimationPath, preloadedAnimations } from '@etherealengine/engine/src/avatar/animation/Util'
 import { LoopAnimationComponent } from '@etherealengine/engine/src/avatar/components/LoopAnimationComponent'
 import { AssetPreviewCameraComponent } from '@etherealengine/engine/src/camera/components/AssetPreviewCameraComponent'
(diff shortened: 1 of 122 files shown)


DanielBelmes:
Okay we're gonna need 'crypto' added to the require section in all needed .mocharc.js files @heysokam @HexaField

HexaField:
> Okay we're gonna need 'crypto' added to the require section in all needed .mocharc.js files @heysokam @HexaField 

Hmm, is this because the tests run cjs with fix-esm? What is the cause of this?

DanielBelmes:
> > Okay we're gonna need 'crypto' added to the require section in all needed .mocharc.js files @heysokam @HexaField
> 
> Hmm, is this because the tests run cjs with fix-esm? What is the cause of this?

I can't completely find out why. But I think the change is worth a try. It might be how mocha is ran within node. But it should be built into node. Maybe Mocha launches its own node instances with no libraries.

DanielBelmes:
@HexaField attempted to test other verisons of node and this was still not working on them when running Mocha. We will return to this a long time from now when we have the bandwith.

PR #10105, opened 2024-04-30
IR-1780 Interactables serializing and mounting cleanup

SamMazerIR:
## Summary
MountPoint/Grabbable/Link Components now add Interactables to the authoring layer, optimized useEffects to Systems/Components/Editors

proper serialization of InteractableComponents added by other components proper serialization of all remaining elements of Interactables proper editor callback selection on self for Interactables and Triggers

## References
[IR-1780](https://tsu.atlassian.net/browse/IR-1780)

## QA Steps

1. Add a new entity to a scene, add a component of one of the following: MountPoint, Link, Grabbable.
2. Observe that it will have also added an InteractableComponent
- note that the default callback is now properly selecting "Self" and the corresponding callback option from your chosen component is chosen as a dropdown option rather than a string (this is now also true for TriggerComponent if there are callbacks on the same entity you add it to) 
- image
4. Modify any properties of the InteractableComponent and save the scene
5. either view the scene source or publish and play the scene, your changes to the InteractableComponent should now be reflected properly
- image

[IR-1780]: https://tsu.atlassian.net/browse/IR-1780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/editor/src/components/properties/GrabbableComponentNodeEditor.tsx
+++ b/packages/editor/src/components/properties/GrabbableComponentNodeEditor.tsx
@@ -23,17 +23,35 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
-import { useComponent } from '@etherealengine/ecs'
+import { isClient } from '@etherealengine/common/src/utils/getEnvironment'
+import { getComponent, hasComponent, UUIDComponent } from '@etherealengine/ecs'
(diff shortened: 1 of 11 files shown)


AidanCaruso (approved):
Looks good to me. XRUI labels aren't appearing inside the activation threshold anymore, but that's not related to this PR. Probably should look into that at some point

the-reality-engineer (approved):

PR #10108, opened 2024-04-30
iR 594 Audit and add postprocessing effects

MbfloydIR:
Improve PostProcessing API

Added the missing post processing effects

GodRaysEffect needs light source that is not available, so it is left out

The following effects are aded but no difference can be seen when applied

SMAAEffect
SSAOEffect
SSREffect
DepthOfFieldEffect
BloomEffect
SSGIEffect
TRAAEffect
MotionBlurEffect
ShockWaveEffect


DanielBelmes:
@MbfloydIR you have some errors on the checks here btw. Double check your commit hooks are running.

packages/editor/src/components/properties/PostProcessingSettingsEditor.tsx
```diff
@@ -90,7 +97,7 @@ const EffectsOptions: Partial<EffectOptionsType> = {
     intensity: { propertyType: PropertyTypes.Number, name: 'Intensity', min: 0, max: 10, step: 0.01 },
     fade: { propertyType: PropertyTypes.Number, name: 'Fade', min: -1, max: 1, step: 0.01 },
     resolutionScale: { propertyType: PropertyTypes.Number, name: 'Resolution Scale', min: -10, max: 10, step: 0.01 },
-    kernelSize: { propertyType: PropertyTypes.Number, name: 'Kerne Size', min: 1, max: 5, step: 1 },
+    kernelSize: { propertyType: PropertyTypes.KernelSize, name: 'KernelSize' },
```
DanielBelmes: The min max and step for this was slightly correct. I actually think there is a KernelSelect we could use for this with values from 0-5. There are others that can use it too.

packages/spatial/src/renderer/effects/PostProcessing.ts
```diff
@@ -401,52 +443,52 @@ export type EffectPropsSchema = {
...
+  [Effects.FXAAEffect]: FXAAEffectProps
+  [Effects.TextureEffect]: TextureEffectProps
   [Effects.VignetteEffect]: VignetteEffectProps
-  // [Effects.LensDistortionEffect]: LensDistortionEffectProps
+  [Effects.LensDistortionEffect]: LensDistortionEffectProps
 }
 
 export type EffectPropsSchemaType = (typeof defaultPostProcessingSchema)[keyof typeof defaultPostProcessingSchema]
 
 export const defaultPostProcessingSchema: EffectPropsSchema = {
```
DanielBelmes, line 415: Were you able to check if the default values for the other effects matched the default values from the library/demos?

packages/spatial/src/renderer/functions/configureEffectComposer.ts
```diff
@@ -142,27 +161,96 @@ export const configureEffectComposer = (entity: Entity, schema?: typeof defaultP
...
+      composer[key] = eff
+      effects.push(eff)
+    } else if (key == Effects.TextureEffect) {
+      let texturePath = effectOptions.texturePath
+      if (texturePath == undefined) {
+        texturePath = null
+      }
+      let texture: Texture | null = null
+      if (texturePath != null) {
+        let textLoad = new TextureLoader()
```
DanielBelmes: @MichaelEstes Do you have anyway with the resourceManager to load something out of a reactor?

packages/spatial/src/renderer/functions/configureEffectComposer.ts
```diff
@@ -142,27 +161,96 @@ export const configureEffectComposer = (entity: Entity, schema?: typeof defaultP
...
+        //have to wait for the texture's image to load and then add the pass to the composer
+        texture = textLoad.load(texturePath, (textureLoaded) => {
+          effectOptions.texture = textureLoaded
+          const eff = new EffectClass(effectOptions)
+          composer[key] = eff
+          effects.push(eff)
+          AddPass()
```
DanielBelmes: I believe you're creating multiple duplicates of the EffectPass with `AddPass()`. We have to refactor this logic into PostProcessingComponent so we can use useEffects and the resource manager in order to load these textures.

DanielBelmes (changes requested):
Some small stuff at first but I believe we do have to refactor configure Effect composer into the PostProcessingComponent earlier than expected for the texture effects.

packages/spatial/src/renderer/components/PostProcessingComponent.tsx
```diff
@@ -63,10 +83,220 @@ export const PostProcessingComponent = defineComponent({
 
     useEffect(() => {
       if (!rendererEntity) return
-      configureEffectComposer(
-        rendererEntity,
-        postprocessingComponent.enabled.value ? postprocessingComponent.effects.get(NO_PROXY_STEALTH) : undefined
-      )
+      //configureEffectComposer(
```
HexaField: we should definitely keep this as a separate function, long useEffects are hard to read

HexaField (changes requested):

HexaField (changes requested):
we should add a supreactor for each of the effects, we can then react to them individually, as well as simplify the logic around the texture loading

packages/spatial/src/renderer/components/PostProcessingComponent.tsx
```diff
@@ -78,7 +78,7 @@ export const PostProcessingComponent = defineComponent({
   /** @todo this will be replaced with spatial queries or distance checks */
   reactor: () => {
     const entity = useEntityContext()
-    const rendererEntity = useScene(entity)
+    const rendererEntity = useScene(Engine.instance.viewerEntity)
```
HexaField: why this change? this means any asset panels will not have postprocessing and apply it to the main canvas

HexaField (changes requested):

HexaField:
closing in favour of continued work https://github.com/EtherealEngine/etherealengine/pull/10148

PR #10109, opened 2024-04-30
IR 465 ECS Materials

AidanCaruso:
## Summary

- Replaces material library hookstate object with components on the ECS

- Material Component is used on every entity that uses a material

  - Material component is broken up into MaterialInstance, MaterialState, MaterialPrototype and MaterialPlugin

  - MaterialInstance is placed on entities with GroupComponents that use materials, and contains UUIDs that point to material entities containing state and relevant data

  - MaterialState contains the material state, pointer to the MaterialPrototype and MaterialPlugin entities, and a list of all instance entities using the material

  - MaterialPrototype contains the material's prototype constructor, name and arguments

  - MaterialPlugin contains the OBC plugin and its parameters

  - Material names and sources are hashed to prevent duplicate materials from the use of duplicate assets

- Material names are unique across the entire scene, if a material name is in use, a numerical suffix is added to any subsiquent duplicate names

 - Setting material prototypes and plugins are done entirely through the ECS via setComponent, reactors propagate these changes to the threejs material state

- Material library studio panel accomodates these changes

  - Global material view to see all materials in a scene by querying material entities with material state

  - Local material view to see materials used by a selected entity by querying material entities associated with the selected entity

- Material instances are now automatically removed by the material component's onRemove call if it's a MaterialInstance

- Moves any material system code that does not make asset sourcing references to spatial package

## References
closes #_insert number here_


packages/engine/src/avatar/components/TransparencyDitheringComponent.tsx
```diff
@@ -69,39 +71,44 @@ export const TransparencyDitheringComponent = Array.from({ length: maxDitherPoin
     }
   })
 })
+const defaultParameters = {
```
HexaField: defaultParameters seems to be unused

packages/engine/src/assets/loaders/gltf/extensions/RegisterMaterialsExtension.ts
```diff
@@ -23,43 +23,33 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
-              material.plugins!.map((plugin) => plugin.toString()).reduce((x, y) => x + y, '')
-          }
-          const materialComponent = registerMaterial(material, { type, path })
-          material.userData?.plugins && materialComponent.plugins.set(material.userData['plugins'])
-        }
-        registerMaterialInstance(material, mesh.entity)
-        //iterate intersected object in the scene and set material
-      })
-  })
+  // const materialLibrary = getState(MaterialLibraryState)
```
HexaField: if registerMaterials is unused now, we should remove it

HexaField: actually looks like it is used in a few places still, but if RegisterMaterialsExtension is redundant it would be great to remove it

packages/engine/src/scene/materials/systems/MaterialLibrarySystem.tsx
```diff
@@ -25,86 +25,136 @@ Ethereal Engine. All Rights Reserved.
...
+  updateMaterialPrototype
+} from '@etherealengine/spatial/src/renderer/materials/materialFunctions'
+import { iterateEntityNode } from '@etherealengine/spatial/src/transform/components/EntityTree'
 
-function PluginReactor({ pluginId }: { pluginId: string }) {
-  const materialLibrary = useState(getMutableState(MaterialLibraryState))
-  const component = materialLibrary.plugins[pluginId]
-  return null
-}
+// function MaterialReactor({ materialId }: { materialId: string }) {
```
HexaField: remove commented code

packages/spatial/src/renderer/materials/MaterialComponent.ts
```diff
@@ -0,0 +1,166 @@
...
+export const MaterialPlugins = [NoiseOffsetPlugin, TransparencyDitheringPlugin]
+
+export enum MaterialComponents {
+  Instance,
+  State,
+  Prototype,
+  Plugin
+}
+
+export const materialByHash = {} as Record<string, EntityUUID>
```
HexaField, line 92: is there a reason these are not hyperflux states? should they be?

packages/spatial/src/renderer/materials/MaterialComponent.ts
```diff
@@ -0,0 +1,166 @@
...
+  [_: string]: {
+    type: string
+    default: any
+    min?: number
+    max?: number
+    options?: any[]
+  }
+}
+
+export const materialSuffix = ' (Material)'
```
HexaField: const exports should be capital case

packages/spatial/src/renderer/materials/constants/DefaultArgs.ts
```diff
@@ -43,6 +43,7 @@ export const ShaderArg = { default: '', type: 'shader' }
 export const ObjectArg = { default: {}, type: 'object' }
 
 export function getDefaultType(value) {
+  if (value == null) return ''
```
HexaField: pefer `if (!value)` over weak null equality check

packages/spatial/src/renderer/materials/constants/plugins/NoiseOffsetPlugin.ts
```diff
@@ -0,0 +1,150 @@
...
+import { PluginObjectType } from '../../../../common/functions/OnBeforeCompilePlugin'
+import { MaterialComponent, MaterialComponents } from '../../MaterialComponent'
+
+export type NoiseOffsetParameters = {
+  noiseTexture: DataTexture | null
+  textureSize: number
+  frequency: number
+}
+
+let time: IUniform | null = null
```
HexaField, line 40: should this be per shader instance instead of a global?

HexaField (changes requested):

packages/engine/src/assets/loaders/gltf/extensions/RegisterMaterialsExtension.ts
AidanCaruso: I agree I totally forgot to go through and replace remaining references

packages/spatial/src/renderer/materials/MaterialComponent.ts
AidanCaruso: My question as well, why should they be?

HexaField: in case we need to react to them? i guess it depends how/why they're used

HexaField: also allows for far easier inspection

AidanCaruso: I agree with the inspection bit. we should be using the ecs for reacting to them though

AidanCaruso: really I dont like having these to begin with but it just eliminates the need for excessive querying. the only one that's strictly necessary is the hashmap

packages/spatial/src/renderer/materials/MaterialComponent.ts
```diff
@@ -0,0 +1,173 @@
...
+  Prototype,
+  Plugin
+}
+
+export const materialByHash = {} as Record<string, EntityUUID>
+export const materialByName = {} as Record<string, EntityUUID>
+export const prototypeByName = {} as Record<string, Entity>
+export const pluginByName = {} as Record<string, Entity>
+
+export const MaterialComponent = Array.from({ length: 4 }, (_, i) => {
```
DanielBelmes, line 97: could also base the length of this off MaterialComponents

packages/spatial/src/renderer/components/GroupComponent.tsx
```diff
@@ -83,6 +83,7 @@ export function addObjectToGroup(entity: Entity, object: Object3D) {
   obj.matrix = transform.matrix
   obj.matrixWorld = transform.matrixWorld
   obj.layers = new Layer(entity)
+
```
DanielBelmes, line 86: sorry to nitpik but Random change

HexaField (approved):

DanielBelmes (approved):

PR #10118, opened 2024-05-01
updated the avatar radius and camera rotation

MbfloydIR:
## Summary
updated the camera rotation to use the world rotation
updated the avatar radius to be the sum of the eyeOffset and the camera Near values

## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/avatar/components/AvatarControllerComponent.ts
+++ b/packages/engine/src/avatar/components/AvatarControllerComponent.ts
@@ -102,7 +102,7 @@ export const AvatarControllerComponent = defineComponent({
     const avatarComponent = useComponent(entity, AvatarComponent)
     const avatarControllerComponent = useComponent(entity, AvatarControllerComponent)
     const isCameraAttachedToAvatar = useHookstate(getMutableState(XRControlsState).isCameraAttachedToAvatar)
-    const camera = getComponent(Engine.instance.cameraEntity, CameraComponent)
+    const camera = useComponent(Engine.instance.cameraEntity, CameraComponent)
 
     useEffect(() => {
       setAvatarColliderTransform(entity)
--- a/packages/engine/src/avatar/components/AvatarControllerComponent.ts
+++ b/packages/engine/src/avatar/components/AvatarControllerComponent.ts
@@ -43,6 +43,8 @@ import { TargetCameraRotationComponent } from '@etherealengine/spatial/src/camer
 import { PhysicsState } from '@etherealengine/spatial/src/physics/state/PhysicsState'
 import { XRControlsState } from '@etherealengine/spatial/src/xr/XRState'
 import { useEffect } from 'react'
+import { CameraComponent } from '../../../../spatial/src/camera/components/CameraComponent'
+import { setAvatarColliderTransform } from '../functions/spawnAvatarReceptor'
 import { AvatarComponent } from './AvatarComponent'
 
 export const AvatarControllerComponent = defineComponent({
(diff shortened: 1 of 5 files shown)


packages/spatial/src/camera/components/CameraComponent.ts
```diff
@@ -28,13 +28,14 @@ import { ArrayCamera, PerspectiveCamera } from 'three'
 import { defineComponent } from '@etherealengine/ecs/src/ComponentFunctions'
 import { addObjectToGroup, removeObjectFromGroup } from '../../renderer/components/GroupComponent'
 
+export const cameraNear = 0.1
```
HexaField: This is not a const, it is a default value that is configurable.

HexaField (changes requested):

packages/spatial/src/camera/components/CameraComponent.ts
MbfloydIR: It is being exported so it can be used to calculate the avatar radius in spawnAvatarReceptor.ts

MbfloydIR: removed the const and put the default value back in the array declaration 
spawnAvatarReceptor now updates the avatar radius by fetching the camera's near value from the camera component

packages/engine/src/avatar/functions/spawnAvatarReceptor.ts
```diff
@@ -62,12 +62,18 @@ import { AvatarAnimationComponent, AvatarRigComponent } from '../components/Avat
 import { AvatarComponent } from '../components/AvatarComponent'
 import { AvatarColliderComponent, AvatarControllerComponent } from '../components/AvatarControllerComponent'
 
-export const avatarRadius = 0.125
+import { CameraComponent } from '../../../../spatial/src/camera/components/CameraComponent'
```
HexaField, line 66: this needs to be a module import, not a relative import

packages/engine/src/avatar/functions/spawnAvatarReceptor.ts
```diff
@@ -62,12 +62,18 @@ import { AvatarAnimationComponent, AvatarRigComponent } from '../components/Avat
...
+import { eyeOffset } from '../systems/AvatarTransparencySystem'
+
+export let avatarRadius = 0 //0.125
 
 export const spawnAvatarReceptor = (entityUUID: EntityUUID) => {
   const entity = UUIDComponent.getEntityByUUID(entityUUID)
   if (!entity) return
 
+  const camera = getComponent(Engine.instance.cameraEntity, CameraComponent)
+  avatarRadius = eyeOffset + camera.near
```
HexaField: this will still need to be updated reactively, as the camera near can change

HexaField (changes requested):

packages/spatial/src/camera/components/CameraComponent.ts
MbfloydIR: Updated the avatar Radius to be calculated as part of the avatar collider's update

packages/engine/src/avatar/functions/spawnAvatarReceptor.ts
```diff
@@ -124,21 +125,33 @@ export const createAvatarCollider = (entity: Entity) => {
...
     shape: Shapes.Capsule,
     collisionLayer: CollisionGroups.Avatars,
     collisionMask: AvatarCollisionMask
   })
 }
 
+const avatarCapsuleOffset = 0.25
+export const setAvatarColliderTransform = (entity: Entity) => {
+  const avatarCollider = getComponent(entity, AvatarColliderComponent)
+  if (avatarCollider === undefined) {
```
HexaField: we should use getOptionalComponent if we are going to check existence, and we can use `!avatarCollider` instead of `=== undefined`

HexaField (changes requested):

packages/engine/src/avatar/components/AvatarControllerComponent.ts
```diff
@@ -100,19 +102,17 @@ export const AvatarControllerComponent = defineComponent({
     const avatarComponent = useComponent(entity, AvatarComponent)
     const avatarControllerComponent = useComponent(entity, AvatarControllerComponent)
     const isCameraAttachedToAvatar = useHookstate(getMutableState(XRControlsState).isCameraAttachedToAvatar)
+    const camera = getComponent(Engine.instance.cameraEntity, CameraComponent)
```
HexaField: should be useComponent

HexaField (changes requested):

packages/engine/src/avatar/functions/spawnAvatarReceptor.ts
MbfloydIR: implemented

packages/engine/src/avatar/components/AvatarControllerComponent.ts
MbfloydIR: implemented

PR #10119, opened 2024-05-02
IR-1820 playmode interactables

SamMazerIR:
## Summary
Allowing interactable usage in studio editor playmode (includes mountpoint, grabbables, link)

## References
[IR-1820](https://tsu.atlassian.net/browse/IR-1820)

## QA Steps
Enter studio, open a scene (ensure scene has interactables in it), hit the play button on the top bar to spawn an avatar and preview.

You should be able to interact with MountPointComponents, vanilla InteractableComponents, and LinkComponents

NOTES: 
- GrabbableComponents will work once their specific system is fixed
- A bug in play mode prevents exiting and re-entering play mode. you will need to refresh your page if you exit play mode before you can re-enter it

[IR-1820]: https://tsu.atlassian.net/browse/IR-1820?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -145,17 +180,33 @@ export const InteractableComponent = defineComponent({
...
+      if (isEditing.value || !input) return
       const canvas = getComponent(Engine.instance.viewerEntity, RendererComponent).canvas
       if (input.inputSources.length > 0) {
         canvas.style.cursor = 'pointer'
       }
       return () => {
         canvas.style.cursor = 'auto'
       }
-    }, [input?.inputSources.length])
+    }, [input?.inputSources.length, isEditing.value])
```
AidanCaruso, line 308: isEditing should not be unwrapped with .value, just use the actual state (in this case just isEditing) for reactor dependency arrays.

AidanCaruso (changes requested):

AidanCaruso:
other than the dependency issue, I tested and everything else looks good! we should probably address the serialized mount point being outdated/not working in the default project's apartment in a future PR

packages/engine/src/interaction/components/InteractableComponent.ts
HexaField: either is fine actually if the effect is reversable

HexaField: .value will re-render if the value changes, not .value will rerender any time a .set() is called on the state (even if the value doesnt change)

AidanCaruso (approved):

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -60,6 +68,33 @@ export enum XRUIActivationType {
...
+  InteractableTransitions.set(entity, transition)
+}
+
+const removeInteractableUI = (entity: Entity) => {
+  if (!isClient || !getMutableState(EngineState).isEditing.value) return //no xrui in editor
+  const interactable = getComponent(entity, InteractableComponent)
+  if (!interactable.label || interactable.label === '' || interactable.uiEntity == UndefinedEntity) return //null or empty label = no ui
+
+  removeEntity(interactable.uiEntity)
+  interactable.uiEntity = UndefinedEntity
```
HexaField: must set component values reactively

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -60,6 +68,33 @@ export enum XRUIActivationType {
...
+/**
+ * Adds an interactable UI to the entity if it has label text
+ * @param entity
+ */
+const addInteractableUI = (entity: Entity) => {
+  if (!isClient || getMutableState(EngineState).isEditing.value) return //no xrui in editor
+  const interactable = getComponent(entity, InteractableComponent)
+  if (!interactable.label || interactable.label === '' || interactable.uiEntity != UndefinedEntity) return //null or empty label = no ui
+
+  interactable.uiEntity = createUI(entity, interactable.label, interactable.uiInteractable).entity
```
HexaField: must set component values reactively

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -60,6 +68,33 @@ export enum XRUIActivationType {
   proximity = 0,
   hover = 1
 }
+
+/**
+ * Adds an interactable UI to the entity if it has label text
+ * @param entity
+ */
+const addInteractableUI = (entity: Entity) => {
+  if (!isClient || getMutableState(EngineState).isEditing.value) return //no xrui in editor
```
HexaField: unnecessary checks

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -60,6 +68,33 @@ export enum XRUIActivationType {
...
+  interactable.uiEntity = createUI(entity, interactable.label, interactable.uiInteractable).entity
+  setComponent(interactable.uiEntity, EntityTreeComponent, { parentEntity: Engine.instance.originEntity })
+
+  const transition = createTransitionState(0.25)
+  transition.setState('OUT')
+  InteractableTransitions.set(entity, transition)
+}
+
+const removeInteractableUI = (entity: Entity) => {
+  if (!isClient || !getMutableState(EngineState).isEditing.value) return //no xrui in editor
```
HexaField: unnecessary checks

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -145,17 +180,33 @@ export const InteractableComponent = defineComponent({
...
+    const isEditing = useMutableState(EngineState).isEditing
+
+    useEffect(() => {
+      setComponent(entity, DistanceFromCameraComponent)
+      setComponent(entity, DistanceFromLocalClientComponent)
+
+      addInteractableUI(entity)
+    }, [])
+
+    useEffect(() => {
```
HexaField: this should be refactored to use cleanup hooks
```ts
 useEffect(() => {
      if (isEditing.value) return
      addInteractableUI(entity)
      return () => {
        removeInteractableUI(entity)
      }
    }, [isEditing.value])    
```
this will also make the `onRemove` cleanup unnecessary

packages/engine/src/interaction/systems/InteractableSystem.ts
```diff
@@ -232,12 +190,12 @@ const execute = () => {
 
 export const InteractableSystem = defineSystem({
   uuid: 'ee.engine.InteractableSystem',
-  insert: { before: TransformSystem },
+  insert: { after: TransformSystem },
```
HexaField: why has this been moved to after the transform system?

packages/spatial/src/input/systems/ClientInputSystem.tsx
```diff
@@ -120,6 +120,7 @@ export function updateGamepadInput(eid: Entity) {
 const pointers = defineQuery([InputPointerComponent, InputSourceComponent, Not(XRSpaceComponent)])
 const xrSpaces = defineQuery([XRSpaceComponent, TransformComponent])
 const inputSources = defineQuery([InputSourceComponent, TransformComponent])
+const nonSpatialInputSources = defineQuery([InputSourceComponent])
```
HexaField: this isn't necessarily all non spatial input sources, rather ti is all input sources. this should be named `inputSourceQuery` and the one above should be called `spatialInputSourceQuery`

HexaField (dismissed):

packages/spatial/src/input/systems/ClientInputSystem.tsx
SamMazerIR: addressed

packages/engine/src/interaction/components/InteractableComponent.ts
SamMazerIR: changed

packages/engine/src/interaction/components/InteractableComponent.ts
SamMazerIR: changed

packages/engine/src/interaction/components/InteractableComponent.ts
SamMazerIR: can you elaborate? I figured we don't want to add an XRUI on the server, and we don't want to run this while editing

packages/engine/src/interaction/components/InteractableComponent.ts
SamMazerIR: removing the !isClient || !isEditing check.

packages/engine/src/interaction/components/InteractableComponent.ts
SamMazerIR: done, makes sense

packages/engine/src/interaction/systems/InteractableSystem.ts
SamMazerIR: reverted due to fixing the orbitcam issue

packages/engine/src/interaction/components/InteractableComponent.ts
SamMazerIR: nevermind, i think i see what you mean after removing the empty dependency useEffect. The remaining useEffect onMount already checked for editing.

Theorizing that isClient checks are redundant for xrUI anyway?

DanielBelmes (approved):
Looking good.

PR #10128, opened 2024-05-04
Add locationType menu

CITIZENDOT:
## References
closes [#IR-1899](https://tsu.atlassian.net/jira/software/c/projects/IR/boards/162?selectedIssue=IR-1899)

## QA Steps
image


HexaField:
Just a comment, if private is the default, we should make it public instead.

CITIZENDOT:
> Just a comment, if private is the default, we should make it public instead.

done!

HexaField (approved):

PR #10134, opened 2024-05-07
Added fsProjectSyncEnabled

MoizAdnan:


HexaField (changes requested):
i don't think this is quite right. config.disableFsProjectSync should be what replaces isDev when we are doing file operations, and for the inverse cases where we are creating invalidations, we should be doing that when an edge cache is enabled, which is another task that i think is worth doing here.

packages/server-core/src/appconfig.ts
```diff
@@ -418,7 +419,8 @@ const config = {
   testEnabled,
   /** @todo when project versioning is fully implemented, remove 'undefined' check here */
   allowOutOfDateProjects:
-    typeof process.env.ALLOW_OUT_OF_DATE_PROJECTS === 'undefined' || process.env.ALLOW_OUT_OF_DATE_PROJECTS === 'true'
+    typeof process.env.ALLOW_OUT_OF_DATE_PROJECTS === 'undefined' || process.env.ALLOW_OUT_OF_DATE_PROJECTS === 'true',
+  enableFsProjectSync: process.env.ENABLE_FS_PROJECT_SYNC ?? 'true'
```
hanzlamateen: This `?? 'true'` needs to be `false`, because our deployments would not contain this `ENABLE_FS_PROJECT_SYNC` env and thus for deployments, `enableFsProjectSync` will be true which we dont want.

> With it being false by default, we would need everyone having this running locally to update their `.env.local` with `ENABLE_FS_PROJECT_SYNC=true`

hanzlamateen: @HexaField your thoughts?

packages/server-core/src/assets/asset/asset.hooks.ts
```diff
@@ -88,12 +88,12 @@ const removeAssetFiles = async (context: HookContext<AssetService>) => {
 
     await storageProvider.deleteResources([assetIterationURL])
 
-    if (isDev) {
+    if (config.enableFsProjectSync === 'true') {
```
hanzlamateen: This should be boolean with name `fsProjectSyncEnabled`

> It seems like we have Enabled convention in appconfig. i.e. testEnabled, kubernetesEnabled, etc.

hanzlamateen (changes requested):

.env.local.default
```diff
@@ -221,3 +221,6 @@ NODE_ENV=production
 OPENSEARCH_USER=admin
 OPENSEARCH_PASSWORD=admin
 OPENSEARCH_HOST=http://localhost:9200
+
+# Switch to `true` to enable local file system operations
```
HexaField, line 225: Should be true by default

packages/server-core/src/appconfig.ts
HexaField: No it should be true by default, and we should add this to our deployments. Otherwise developers locally will have projects going out of sync.

packages/server-core/src/assets/asset/asset.hooks.ts
```diff
@@ -88,12 +88,12 @@ const removeAssetFiles = async (context: HookContext<AssetService>) => {
...
     await storageProvider.deleteResources([assetIterationURL])
 
-    if (isDev) {
+    if (config.fsProjectSyncEnabled) {
       const assetFilePath = path.resolve(projectPathLocal, assetIterationURL)
       if (fs.existsSync(assetFilePath)) {
         fs.rmSync(path.resolve(assetFilePath))
       }
-    } else {
+    } else if (config.server.edgeCachingEnabled) {
```
HexaField: This shouldn't be an else, just a separate conditional

packages/server-core/src/assets/asset/asset.hooks.ts
```diff
@@ -290,8 +290,9 @@ export const renameAsset = async (context: HookContext<AssetService>) => {
         true
       )
 
-      if (isDev) fs.renameSync(oldLocalFile, newLocalFile)
-      else await context.app.service(invalidationPath).create([{ path: oldDirPath }, { path: newDirPath }])
+      if (config.fsProjectSyncEnabled) fs.renameSync(oldLocalFile, newLocalFile)
+      else if (config.server.edgeCachingEnabled)
```
HexaField: This shouldn't be an else, just a separate conditional

packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -370,7 +369,7 @@ export class FileBrowserService
       )
     }
 
-    if (isDev && PROJECT_FILE_REGEX.test(key)) fs.rmSync(path.resolve(projectsRootFolder, key), { recursive: true })
+    if (config.fsProjectSyncEnabled) fs.rmSync(path.resolve(projectsRootFolder, key), { recursive: true })
```
HexaField, line 373: We seem to have a discrepancy, sometimes we are doing a truthy conditional, other times a strict equality to 'true'.

HexaField (changes requested):

hanzlamateen (approved):

HexaField (approved):

PR #10137, opened 2024-05-08
Refactoring MathConstants.ts Vector3 and folding in Axis3D.ts

SamMazerIR:
## Summary
refactoring our math constants, fixing a bug regarding RHS directionality and combining Axis3D.ts into MathConstants.ts and removing Axis3D.ts

fixed a bug with ObjectDirection.Right being the incorrect values (as we are in a RHS coord system)
[LHS vs RHS reference](https://www.evl.uic.edu/ralph/508S98/coordinates.html)

## References
closes [IR-1923](https://tsu.atlassian.net/browse/IR-1923)

[IR-1923]: https://tsu.atlassian.net/browse/IR-1923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


the-reality-engineer (approved):
A little bit more verbose than before, but it does seem to add clarity in most of the contexts where these constants are being used. 

I would only suggest following the convention for quaternion constants a bit closer. Something like:

V3_UP
V3_DOWN
V3_ONE
V3_ZERO

AidanCaruso (approved):

PR #10142, opened 2024-05-09
Admin search fixes

HexaField:
## References
closes #_insert number here_


CITIZENDOT:
Do we need search for ProjectsTable too? That's the only page without search bar.

CITIZENDOT (approved):

aditya-mitra (approved):
LGTM

HexaField:
> Do we need search for ProjectsTable too? That's the only page without search bar.

Maybe another time, it's fine for now.

PR #10144, opened 2024-05-10
Add support for project manifest.json

HexaField:
## Summary

- Adds manifest.json spec to supersede package.json and xrengine.config.ts as required project files
- Makes asset service generic instead of specific for scene GLTFs
- Adds helpful resolvers to asset and location services
- Improve git API unit tests
- Upload all resources to storage provider as part of seeding projects
- Studio file upload helper functions no longer hardcodes directory for uploaded assets
- A manifest.json automatically generates in projects that do not have one
- Scene CRUD operations update manifest.json scenes field

## References
closes #_insert number here_


packages/projects/template-project/manifest.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "default-project",
```
hanzlamateen: Should this be `ee-template`?

hanzlamateen (changes requested):

hanzlamateen (changes requested):
1. I added a new scene and it was not reflected in manifest.json.
2. I renamed apartment scene and it was not reflected in manifest.json

packages/projects/template-project/manifest.json
HexaField: Whoops, yes, but it shouldn't matter, it's overriden whenever it is used

DanielBelmes:
You can rename a scene into an already existing scene and it fails in SQL but pushes to manifest.

So default when no scenes are detected is no scenes key (edited)
but if you delete all scenes it's scenes: []
could lead to logic issues if the condition only checks for length

HexaField:
> You can rename a scene into an already existing scene and it fails in SQL but pushes to manifest.
> 
> So default when no scenes are detected is no scenes key (edited) but if you delete all scenes it's scenes: [] could lead to logic issues if the condition only checks for length

Restructuring into after hooks will take a bit of work, and I can just solve that one problem and we can use validators to ensure the data is compliant before it hits the mutation hooks. We can restructure the effects of hooks another time, but it is a good point to raise.

hanzlamateen (approved):
Looks good.

> Assuming Daniel's comments will be handled in a separate PR

packages/editor/src/functions/sceneFunctions.tsx
```diff
@@ -103,9 +104,7 @@ export const saveSceneGLTF = async (
 
     return
   }
-  const result = await Engine.instance.api
-    .service(assetPath)
-    .create({ name: sceneNameWithoutExtension, assetURL, project: projectName })
+  const result = await Engine.instance.api.service(assetPath).create({ assetURL, project: projectName })
```
MoizAdnan: This should have `isScene: true` in data argument. When saving a scene using `save as`, a new scene is created but before it can update manifest.json, it exits in `createSceneFiles` hook in `asset.hooks` on line 230. As `isScene` is null, resulting in the creation of a new scene without manifest.json updating

MoizAdnan (changes requested):

MoizAdnan (approved):

packages/server-core/src/assets/asset/asset.hooks.ts
```diff
@@ -193,65 +177,74 @@ export const createSceneInStorageProvider = async (context: HookContext<AssetSer
...
-        GLTF_ASSET_FILES.map((file) => {
-          return { path: `${context.directory}${data.name}${file}` }
-        })
-      )
-  } catch (e) {
-    logger.error(e)
-    logger.info(GLTF_ASSET_FILES)
+  // when seeding, no sourceURL is given, and the asset file exists, we just want to add the table entry
+  if (!data.sourceURL) {
+    const fileExists = await context.app.service(fileBrowserPath).get(data.assetURL!)
```
MoizAdnan, line 182: When QAing `save as`, file already gets uploaded in `saveSceneGLTF` and `fileExists` returns true. The hook is exited, because of this manifest.json does not update

MoizAdnan (changes requested):

PR #10147, opened 2024-05-13
fix update project submission

aditya-mitra:
## Summary

Update Project Submission was not working earlier due to `ProjectUpdateStatus` being undefined and not uninitialized. Fixed that by getting the state during submission.

## References
closes #_insert number here_


packages/client-core/src/admin/components/project/AddEditProjectModal.tsx
```diff
@@ -109,6 +109,7 @@ export default function AddEditProjectModal({
 
   const projectUpdateStatus = useHookstate(getMutableState(ProjectUpdateState)[project.name])
   useEffect(() => {
+    console.log('debug1 the project name was', project)
```
hanzlamateen: I think the consoles logs can be removed

packages/client-core/src/admin/components/project/ProjectTable.tsx
```diff
@@ -82,122 +82,119 @@ export default function ProjectTable() {
...
+        size="small"
+        className="bg-theme-blue-secondary mr-2 h-min whitespace-pre text-[#214AA6] disabled:opacity-50 dark:text-white"
+        disabled={project.name === 'default-project'}
+        onClick={() => {
+          PopoverState.showPopupover(
+            <AddEditProjectModal
+              update={true}
+              inputProject={project}
+              onSubmit={async () => {
+                const projectUpdateStatus = getMutableState(ProjectUpdateState)[project.name].value
```
hanzlamateen: I think we can move this onSubmit call back to a method instead of inline.

hanzlamateen (changes requested):

hanzlamateen (approved):

PR #10148, opened 2024-05-13
Refactor PostProcessingComponent into reactive registry of effects

MbfloydIR:
## Summary
Added a Highlight system to add the highlight component to selected items.
Based off of the branch michael/IR-594-postprocessing since it also effects the post processing changes

## References
closes #_insert number here_


packages/spatial/src/renderer/components/PostProcessingComponent.tsx
```diff
@@ -59,16 +104,532 @@ export const PostProcessingComponent = defineComponent({
...
   }
 })
+
+const RendererReactor = (props: { entity: Entity; rendererEntity: Entity }) => {
+  const { entity, rendererEntity } = props
+
+  const postprocessingComponent = useComponent(entity, PostProcessingComponent)
+  const camera = useComponent(rendererEntity, CameraComponent)
+  const renderer = useComponent(rendererEntity, RendererComponent)
+  const composer = new EffectComposer(renderer.value.renderer)
```
HexaField: this is creating a new composer and render pass every time the renderer reactor re-runs.

we should use a factory function for the initial state of this reactor like we do starting line 152 with the new scene

packages/editor/src/systems/HighlightSystem.ts
```diff
@@ -0,0 +1,55 @@
...
+import { HighlightComponent } from '@etherealengine/spatial/src/renderer/components/HighlightComponent'
+import { useEffect } from 'react'
+import { SelectionState } from '../services/SelectionServices'
+
+export const highlightQuery = defineQuery([HighlightComponent])
+
+const reactor = () => {
+  const selectedEntities = SelectionState.useSelectedEntities()
+
+  for (const entity of highlightQuery()) removeComponent(entity, HighlightComponent)
```
HexaField: why are we running a query in the body of a reactor? this is an anti pattern. if the intention is to clean up the highlight component, we can do it by returning a cleanup function in the use effect

packages/spatial/src/renderer/components/PostProcessingComponent.tsx
```diff
@@ -59,16 +104,532 @@ export const PostProcessingComponent = defineComponent({
...
+  const useDepthDownsamplingPass = useHookstate(false)
+
+  //const composer = useHookstate<EffectComposer>(() => new EffectComposer(renderer.value.renderer))
+  const effects = useHookstate<Record<string, Effect>>({})
+
+  useEffect(() => {
+    velocityDepthNormalPass.set(new VelocityDepthNormalPass(scene, camera))
+  }, [scene])
+
+  useEffect(() => {
```
HexaField: i think a cleaner pattern for specific effects would be to define a dictionary of effects to create/destroy functions that are invoked, rather than hardcoding each effect. we can then use sub-reactors to map each effect into it's own react component, and handle create/destroy/update events in a controlled and generic way.

HexaField (changes requested):

packages/editor/src/systems/HighlightSystem.ts
MbfloydIR: query removed, clean up now done in reaction to schema.isActive being changed to false

packages/engine/src/postprocessing/components/PostProcessingComponent.tsx
```diff
@@ -0,0 +1,572 @@
...
+    scene,
+    lut1DEffectTexture,
+    lut3DEffectTexture,
+    velocityDepthNormalPass,
+    useVelocityDepthNormalPass,
+    normalPass,
+    depthDownsamplingPass,
+    composer
+  } = props
+  const lookUp = {
```
HexaField, line 275: this mapping can be a hyperflux state or at least a module scope object (like the ones in PostProcessing.ts) such that it can be extensible

packages/spatial/tsconfig.json
```diff
@@ -41,5 +41,5 @@
     "../ecs/**/*",
     "./**/*.ts",
     "./**/*.tsx"
-  ]
+, "../engine/src/postprocessing/PostProcessing.ts"  ]
```
HexaField: vscode likes to do this when moving files, please fix

HexaField (changes requested):

packages/engine/src/postprocessing/components/PostProcessingComponent.tsx
```diff
@@ -0,0 +1,572 @@
...
+    VignetteEffect: VignetteEffectProcess
+  }
+
+  useEffect(() => {
+    if (!schema.isActive) {
+      effects[effectKey].set(none)
+    }
+  }, [schema.isActive])
+
+  useEffect(() => {
```
DanielBelmes, line 313: For the next changes we can probably just have schema.isActive as the dep and both create and destroy in one useEffect. I don't wanna break functionality if this works for now 😅

packages/engine/src/postprocessing/components/PostProcessingComponent.tsx
```diff
@@ -0,0 +1,572 @@
...
+        const val = effectsVal[key]
+        renderer.value.effectComposer[key] = val
+      }
+    }
+
+    //always have the smaa effect
+    const smaaPreset = getState(RenderSettingsState).smaaPreset
+    const smaaEffect = new SMAAEffect({
+      preset: smaaPreset,
+      edgeDetectionMode: EdgeDetectionMode.COLOR
```
DanielBelmes, line 190: seems like we forgot to add this to the effect composer. Gonna do it now

HexaField (changes requested):
We spoke on discord to again change the implementation to a full blown registry, making the component extensible and well scoped.

```ts
/** Spatial Package */
type EffectRegistryEntry = {
  reactor: (props: { parameters; isActive }) => void
  defaultValues: any // specifies the default values for the effect
  schema: any // specifies a schema for the editor to generate UI for each effect
}

export const PostProcessingEffectState = defineState({
  name: 'PostProcessingEffectState',
  initial: {} as Record<string, EffectRegistryEntry>
})

// PostProcessingComponent
const reactor = () => {
  const postProcessingComponent = useComponent(entity, PostProcessingComponent)
  const rendererEntity = useScene(entity) // get renderer entity
  const effects = useComponent(entity, PostProcessingEffectState)

  useEffect(() => {
    // create composer etc
  }, [])

  // for each effect specified in our postProcessingComponent, we mount a sub-reactor based on the effect registry for that effect ID
  return (
    <>
      {Object.entries(postProcessingComponent.effects.value).map(([key, value]) => {
        const effect = effects[key] // get effect registry entry
        return (
          <effect.reactor
            key={key}
            parameters={postProcessingComponent.effects[key]}
            isActive={value.isActive}
            rendererEntity={rendererEntity}
          />
        )
      })}
    </>
  )
}

/** Engine package */

// for effects using assets, the effect needs to exist in the engine package
const ChromaticAberrationEffectProcessReactor = (props: { parameters; isActive; rendererEntity }) => {
  const postProcessgComponent = useComponent(entity, PostProcessingComponent)
  const isLutPassActive = postProcessgComponent.effects[lutpass].isActive // if we want an effect to be active only if another effect is active
  const effectComposer = useComponent(props.rendererEntity, RendererComponent).effectComposer
  const mytexture = useAsset()

  useEffect(() => {
    if (!props.isActive || isLutPassActive) return
    const { schema, effects } = props
    const eff = new ChromaticAberrationEffect(schema)
    effects[Effects.ChromaticAberrationEffect].set(eff)
    effectComposer.addPass(eff)
    return () => {
      //cleanup
    }
  }, [props.isActive, isLutPassActive])
}

// registers the effect
getMutableState(PostProcessingEffectState)[Effects.ChromaticAberrationEffect].set({
  reactor: ChromaticAberrationEffectProcessReactor,
  defaultValues: {
    //...
  },
  schema: {
    hue: { propertyType: PropertyTypes.Number, name: 'Hue', min: -1, max: 1, step: 0.01 },
    saturation: { propertyType: PropertyTypes.Number, name: 'Saturation', min: -1, max: 1, step: 0.01 }
  }
})





/**
 * 1. create registry and move effect definitions to use registry 
 *    (removing Effects, EffectMap, defaultPostProcessingSchema etc in PostProcessing.ts)
 * 
 * 2. register each effect into the registry
 * 
 * 3. refactor PostProcessingComponent to use the registry definitions and mount sub-reactors for each effect
 * 
 * 4. refactor studio editor to use the registry definition's schemas for generating UI
 */
```

packages/engine/src/postprocessing/components/PostProcessingComponent.tsx
DanielBelmes: Forgot the reason for this. A lot of the new effects are convolution effects that can't mix with SMAA. Gonna just expose in editor and allow people to decide at scene authoring I think. Though we gotta consider error messaging or handling for this library. It's a headache.

PR #10156, opened 2024-05-14
Fixed bugs with builder commit fetching

barankyle:
## Summary

builder-info.class.ts was using the wrong value from privateECRRegexExec.

findBuilderTags in project-helper.ts was not using the new fromIni method to get credentials for AWS. Updated it to do so.

Made S3 and EKS SDKs save their credentials files to separate files, so that they don't collide with each other.

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/projects/project/project-helper.ts
+++ b/packages/server-core/src/projects/project/project-helper.ts
@@ -23,8 +23,9 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
-import { ECRClient } from '@aws-sdk/client-ecr'
+import { DescribeImagesCommand as DescribePrivateImagesCommand, ECRClient } from '@aws-sdk/client-ecr'
 import { DescribeImagesCommand, ECRPUBLICClient } from '@aws-sdk/client-ecr-public'
+import { fromIni } from '@aws-sdk/credential-providers'
 import * as k8s from '@kubernetes/client-node'
 import appRootPath from 'app-root-path'
 import { exec } from 'child_process'
@@ -99,6 +100,9 @@ export const privateECRTagRegex = /^[a-zA-Z0-9]+.dkr.ecr.([\w\d\s\-_]+).amazonaw
 
 const BRANCH_PER_PAGE = 100
 const COMMIT_PER_PAGE = 10
+
+const awsPath = './.aws/eks'
+const credentialsPath = `${awsPath}/credentials`
 
 const execAsync = promisify(exec)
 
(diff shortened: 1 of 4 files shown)


packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -778,63 +782,87 @@ export const findBuilderTags = async (): Promise<Array<ProjectBuilderTagsType>>
...
+
     const ecr = new ECRPUBLICClient({
-      credentials: {
-        accessKeyId: config.aws.eks.accessKeyId,
-        secretAccessKey: config.aws.eks.secretAccessKey
-      },
+      credentials: fromIni({
+        profile: config.aws.eks.roleArn ? 'role' : 'default',
+        filepath: credentialsPath
+      }),
```
lucas3900, line 794: Why are we creating a cred file each time, instead of just passing the values to client when we instantiate it?

packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -778,63 +782,87 @@ export const findBuilderTags = async (): Promise<Array<ProjectBuilderTagsType>>
...
+            tag,
+            commitSHA: tagSplit.length === 1 ? tagSplit[0] : tagSplit[1],
+            engineVersion: tagSplit.length === 1 ? 'unknown' : tagSplit[0],
+            pushedAt: imageDetails.imagePushedAt!.toJSON()
+          }
+        })
+    } catch (err) {
+      logger.error('Failure to get private ECR images')
+      logger.error('Command that was sent %o', result)
+      logger.error(err)
```
lucas3900, line 863: Do we have a way to suppress these log statements on certain deployments? For instance AWS error messages can contain sensitive data, so I doubt we want these in the console on production.

lucas3900 (approved):

packages/server-core/src/projects/project/project-helper.ts
barankyle: I don't think we have that right now.

packages/server-core/src/projects/project/project-helper.ts
barankyle: I updated it to not write the credentials file if it exists.

lucas3900: I was more asking about why we are using cred files, instead of passing those values to the client directly. Is it a paradigm across the app to write cred files, or will other parts of the applications expect credential files?

barankyle: When I was looking at authenticating the AWS SDK with Role ARNs, it seemed like the only way to do that was from a credentials file: https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html

lucas3900: You can use the assume_role command on the client. But it does seem like the syntax is a bit annoying for the JS SDK: https://docs.aws.amazon.com/code-library/latest/ug/sts_example_sts_AssumeRole_section.html

Since this is an admin only thing, probably not a huge deal to have this extra file I/O.

PR #10157, opened 2024-05-14
IR-1775-Split-dev-test-suite-into-Examples-Tests-and-Benchmarks

MichaelEstes:
## References
closes #_insert number here_


dinomut1 (approved):
only failing test is regarding project permissions -- there's no way this is related

PR #10158, opened 2024-05-14
Use component's true values in the node editor

CITIZENDOT:
## References
closes #_insert number here_


HexaField (changes requested):
why are we removing local/world support?

CITIZENDOT:
Accounted for transform space when any of the transforms changed!

HexaField:
duplicate work going on here https://github.com/EtherealEngine/etherealengine/pull/10166

DanielBelmes:
I'm not entirely sure why but with your changes when adding a child to a parent. Your useEffect is moving the child and parents position on any change or even reselect.

PR #10160, opened 2024-05-14
Fix/sync scene json assets insertion

barankyle:
## Summary

Fetching of existing assets was only getting the first page, so scenes with
more than that would mistakenly try to insert missed assets again and
trigger uniqueness errors. Now assets are fetched with paginate: false.

## References
closes #_insert number here_


HexaField:
This is addressed in the manifest json branch, so I think we can close this.

barankyle:
Addressed in #10144

PR #10161, opened 2024-05-14
[IR-1828] fix: Reparenting Transform applied based on the new Parent Transform

heysokam:


packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -623,6 +623,19 @@ const reparentObject = (entities: Entity[], before?: Entity | null, parent = get
...
 
+        // Ensure the entity Transform remains unmodified when reparented
+        const node = getGLTFNodeByUUID(gltf.data, entityUUID) // Get the GLTF Node for the entity
+        if (node) {
+          // Get the transforms for both entitites
+          const parentTransform = getComponent(parent, TransformComponent)
+          const entityTransform = getComponent(entity, TransformComponent)
+          // Calculate the new matrix relative to the new parent entity, and apply the matrix to its GLTF node.matrix
+          node.matrix = tempMatrix4
+            .copy(entityTransform.matrix)
```
HexaField: both of these might need to be `matrixWorld`

HexaField (changes requested):

packages/editor/src/functions/EditorControlFunctions.ts
heysokam: Changed at 9e40a4e2895fefe353d0842c67354ddaca67d436

HexaField (approved):

PR #10163, opened 2024-05-15
Reworked assetURL in saveSceneGLTF

MoizAdnan:
When using minio, bucket name would get appended to the assetURL and break saving scene flow. In this PR, the assetURL is now formed by replacing fileServer rather than using `url.pathname`. This is just reverse-engineering the key for GLTF Entity that is being set on line 146 of `sceneFunctions.ts`: https://github.com/EtherealEngine/etherealengine/blob/dev/packages/editor/src/functions/sceneFunctions.tsx#L146


hanzlamateen (approved):
Looks good

HexaField (approved):

PR #10165, opened 2024-05-15
Ir 2021 lookdev

JT00y:
## Summary
Add new .gltf subtype for postprocessing, skybox, and general look dev


dinomut1 (changes requested):
moving asset loader code into addmedianode

dinomut1 (approved):
great job

PR #10166, opened 2024-05-15
IR-2070-fix-transform-editor-reactivity

DanielBelmes:
## Summary
Refrence .value of pos,rot,scale. We don't proxify matrix or matrix world so they weren't triggering hookstate updates

## References
IR-2030
IR-2070


DanielBelmes:
annoying test failures occuring

packages/editor/src/components/properties/TransformPropertyGroup.tsx
```diff
@@ -61,15 +56,14 @@ const scale = new Vector3()
...
   const transformComponent = useComponent(props.entity, TransformComponent)
   const transformSpace = useHookstate(getMutableState(EditorHelperState).transformSpace)
 
-  transformSpace.value === TransformSpace.world
-    ? transformComponent.matrixWorld.value.decompose(position, rotation, scale)
-    : transformComponent.matrix.value.decompose(position, rotation, scale)
-
-  scale.copy(transformComponent.scale.value)
+  useEffect(() => {
+    transformSpace.value === TransformSpace.world
```
HexaField: we need to be listening for transformSpace in this useEffect too, as well as transformComponent.rotation

any time any dependent data updates, we must respond.

also, i'm not sure if this is actually going to reactively update since position, rotation and scale are plain maths objects

HexaField (changes requested):

packages/editor/src/components/properties/TransformPropertyGroup.tsx
DanielBelmes: It does reactively update. For some reason rotation also triggers the use effect without being a child. I shouldn't of missed it though. Honestly id rather revert it to how it was before which was directly reading from use component. But I expected you had a reason to refactor it this way.

HexaField (approved):

PR #10169, opened 2024-05-16
Refactor regex patterns for project keys

MoizAdnan:


barankyle:
From check-errors failing, looks like there's at least one more file that's still importing/using the regexes that were removed.

MoizAdnan:
> From check-errors failing, looks like there's at least one more file that's still importing/using the regexes that were removed.

Fixed errors

barankyle (approved):

PR #10173, opened 2024-05-16
Updated LinkedIn OAuth scopes for new app Product

barankyle:
## Summary

LinkedIn changed their app product and scopes.
Apps now need to have the
"Sign in with LinkedIn with OpenID Connect"
product added, which adds scopes 'profile' and 'email'. Updated authentication-setting.seed.ts and appconfig.ts to use these new scopes for OAuth flow.

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/setting/authentication-setting/migrations/20240516222541_change-linkedin-scopes.ts
+++ b/packages/server-core/src/setting/authentication-setting/migrations/20240516222541_change-linkedin-scopes.ts
@@ -1,0 +1,53 @@
+/*
+CPAL-1.0 License
+
+The contents of this file are subject to the Common Public Attribution License
+Version 1.0. (the "License"); you may not use this file except in compliance
+with the License. You may obtain a copy of the License at
+https://github.com/EtherealEngine/etherealengine/blob/dev/LICENSE.
+The License is based on the Mozilla Public License Version 1.1, but Sections 14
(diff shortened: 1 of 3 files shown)


lucas3900 (approved):
much cleaner :+1:

packages/server-core/src/setting/authentication-setting/migrations/20240516222541_change-linkedin-scopes.ts
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex: Knex): Promise<void> {
+  const authSettings = await knex.table(authenticationSettingPath).first()
+
+  if (authSettings) {
+    const oauthSettings = JSON.parse(authSettings.oauth)
+    oauthSettings.linkedin.scope = LINKEDIN_SCOPES
```
HexaField, line 41: @barankyle we shouldn't rely on outside data that might change in the future for migrations, can we change this to be a hardcoded copy of the correct data?

barankyle: https://github.com/EtherealEngine/etherealengine/pull/10181

PR #10183, opened 2024-05-17
fixing localStorage plugin

DanielBelmes:
## References
closes #_insert number here_

### DIFF
--- a/packages/hyperflux/functions/StateFunctions.ts
+++ b/packages/hyperflux/functions/StateFunctions.ts
@@ -143,11 +143,18 @@ export function syncStateWithLocalStorage<S, E extends Identifiable>(
         }
       },
       onSet: (state, desc) => {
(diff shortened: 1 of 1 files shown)


MichaelEstes (approved):
Working for me on local, looks good

the-reality-engineer (approved):

the-reality-engineer (approved):

packages/hyperflux/functions/StateFunctions.ts
```diff
@@ -142,12 +142,11 @@ export function syncStateWithLocalStorage<S, E extends Identifiable>(
           if (storedValue !== null && storedValue !== 'undefined') state[key].set(JSON.parse(storedValue))
```
speigg, line 142: We are still calling `.set` here synchronously, which seems to cause onSet to be called without a rootState set

PR #10185, opened 2024-05-17
[IR-2053] Unit Tests: Core All ECS Hooks/Reactors

heysokam:
## Summary
- [x] useQuery
- [x] ~~useAllComponents~~
- [x] useComponent
- [x] useOptionalComponent
- [x] UUIDComponent


packages/ecs/src/UUIDComponent.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+    it('should set/get the data of the component.', () => {
+      // Case1: set/get
+      setComponent(entity1, UUIDComponent, TestUUID)
+      const component1 = getComponent(entity1, UUIDComponent)
+      assert.ok(component1, 'The UUIDComponent did not get set correctly')
+      assert.equal(component1, TestUUID, 'The UUID value did not get set correctly')
+    })
+
+    it("shouldn't change the data when set multiple times with the same data", () => {
```
DanielBelmes, line 78: What happends if we set UUIDComponent with a different UUID? Do we throw an error if we try to change it?

heysokam: Nothing. It returns early

packages/ecs/src/UUIDComponent.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+      assert.equal(UndefinedEntity, UUIDComponent.entitiesByUUIDState[TestUUID].value)
+      assert.equal(getOptionalComponent(entity1, UUIDComponent), undefined)
+    })
+  })
+
+  describe('getEntityByUUID', () => {
+    it('should return the correct entity', () => {
+      setComponent(entity1, UUIDComponent, TestUUID)
+      const testEntity = UUIDComponent.getEntityByUUID(TestUUID)
+      assert.equal(testEntity, UUIDComponent.entitiesByUUIDState[TestUUID].value)
```
DanielBelmes, line 133: you can just pass entity1 into this condition as testEntity should be the same type and value as entity1.

heysokam: The idea of that `entitiesByUUIDState` check is to confirm that the underlying data is the same as the expected value, independent of what the function is doing internally
Added the comment as an extra check with c37140192ebfb47c3732e8c091ba64c751bd3638

packages/ecs/src/QueryFunctions.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+    it('should create a valid query', () => {
+      const query = defineQuery([component])
+      setComponent(entity1, component)
+      setComponent(entity2, component)
+      assertDefinedQuery(query, [entity1, entity2])
+    })
+  })
+
+  /**
+   *  @todo : Why does `query()` still work after calling `removeQuery(query)`
```
HexaField: This is a design decision of bitecs. Queries are defined at global scope, but only initialized lazily when they are first called. Removing a query simply releases it's instance inside bitecs until it is called again. All queries are automatically cleaned up when the bitecs world is destroyed (which happens when the hyper store / engine is destroyed)

HexaField: we can remove this commented out test

packages/ecs/src/QueryFunctions.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+    afterEach(() => {
+      removeEntity(entity1)
+      removeEntity(entity2)
+      return destroyEngine()
+    })
+
+    // Define the Reactor that will run the tested hook
+    const Reactor = () => {
+      const data = useQuery([component])
+      console.log('render', data)
```
HexaField: remove console logs

packages/ecs/src/QueryFunctions.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+      }, [data])
+      return null
+    }
+
+    it('should return an empty array when entities have the component', async () => {
+      const ExpectedValue = [] as Entity[]
+      const tag = <Reactor />
+      const { rerender, unmount } = render(tag)
+      await act(() => rerender(tag))
+      assertArrayNotEqual(
```
HexaField: i'm not a fan of asserting against some arbitrary value. we should just be asserting the data we expect. this is functionally equivalent as if `result` is undefined

packages/ecs/src/QueryFunctions.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+import { act, render } from '@testing-library/react'
+import assert from 'assert'
+import React, { useEffect } from 'react'
+import { ComponentMap, defineComponent, setComponent } from './ComponentFunctions'
+import { destroyEngine, startEngine } from './Engine'
+import { Entity, UndefinedEntity } from './Entity'
+import { createEntity, removeEntity } from './EntityFunctions'
+import { Query, defineQuery, useQuery } from './QueryFunctions'
+
+function assertArrayEqual<T>(A: Array<T>, B: Array<T>, err = 'Arrays are not equal') {
```
HexaField, line 36: are these helpers necessary? does `assert.deepEqual` not work for just two arrays?

HexaField (changes requested):

packages/ecs/src/QueryFunctions.test.tsx
heysokam: Yes, but you cannot check for undefined coming from the reactor if you don't initialize with some dummy data before doing so.

With this check, you can test if the reactor touched the result, and also check if the reactor applied an `undefined` value to the result (like with optional values, etc)

That's the reasoning behind it.

packages/ecs/src/QueryFunctions.test.tsx
heysokam: I don't know, honestly.
I tried with `equal` and it was failing because their references were different even if their data was equal.
I will try deepEqual :writing_hand:

heysokam: Well, apparently I did try it before.
It doesn't work with `assert.deepEqual` directly on the arrays without iterating :shrug:

packages/ecs/src/QueryFunctions.test.tsx
heysokam: Removed with 2cf1110c793eee66fbb7e8d3f634bdd31e8c9bd7

packages/ecs/src/QueryFunctions.test.tsx
heysokam: Applied on 2cf1110c793eee66fbb7e8d3f634bdd31e8c9bd7

packages/ecs/src/QueryFunctions.test.tsx
HexaField: we should be asserting both that the value is undefined and that the useEffect has run. something like a counter increment along with the result data would suffice

heysokam: Changed on 0320072
@HexaField Let me know if the `counter` variable checks are correct :writing_hand:

DanielBelmes:
Has conflicts @heysokam. Though I'm not sure how they got there since this PR creates the file

packages/ecs/src/UUIDComponent.test.tsx
DanielBelmes: oh yea reading this back two weeks later I was dumb.

DanielBelmes (approved):
Preliminary approval. Still pending integration with @speigg tests he wrote for immedieteUseEffect.

heysokam:
> Preliminary approval. Still pending integration with @speigg tests he wrote for immedieteUseEffect.

@DanielBelmes Should be good to go now :writing_hand:

PR #10189, opened 2024-05-18
IR-2012 Update file system operation to promise API (Partial)

jonathan-casarrubias:
## Summary

Due the extensive use of file system operations I decided to approach this story by sending portions of each fs api migration with the objective of avoiding extensive and overwhelming code reviews as well as to reduce risk.

For this pull request I introduced writeFileAsync, existsAsync and deprecated deleteFolderRecursive.

I have completely replaced any writeFileSyncRecursive call for the new writeFileAsync.
I have completely replaced any deleteFolderRecursive call for the native node.js api fps.rmdir(path, { recursive: true })
I have partially replaced some fs.existsSync calls for the new existsAsync function, will continue replacing these calls in subsequent PRs

## Subtasks Checklist
[x] Implement non-blocking writeFileAsync helper
[x] Implement non-blocking existsAsync helper
[x] Implement non-blocking cpAsync helper
[x] Replace deleteFolderRecursive helper in favour of native non-blocking fps.rmdir function

## Breaking Changes
N/A

## References
JIRA Ticket IR-2012

## QA Steps

Already modified tests using new non-blocking APIs should still pass
`npm run test`


jonathan-casarrubias:
@cla-bot check

packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -1426,8 +1426,8 @@ export const createExecutorJob = async (
   })
 }
 
-export const copyDefaultProject = () => {
-  deleteFolderRecursive(path.join(projectsRootFolder, `default-project`))
+export const copyDefaultProject = async () => {
```
HexaField, line 1430: usage of `copyDefaultProject` has not been changed to be awaited. it's called in `seeder,ts` and `project-helper.ts`

packages/common/src/utils/fsHelperFunctions.ts
```diff
@@ -57,7 +101,11 @@ export function writeFileSyncRecursive(filename, content, charset = undefined) {
   // -- write file
   fs.writeFileSync(root + filepath, content, charset)
 }
-
+*/
```
HexaField: if these are deprecated and unused, we should just removed them

packages/client/scripts/populate-assetlinks.ts
```diff
@@ -25,12 +25,16 @@ Ethereal Engine. All Rights Reserved.
...
 import fetch from 'node-fetch'
+import path from 'path'
 
-import { writeFileSyncRecursive } from '@etherealengine/common/src/utils/fsHelperFunctions'
+import { writeFileAsync } from '@etherealengine/common/src/utils/fsHelperFunctions'
 
-const envPath = appRootPath.path + `/packages/client/public/.wellknown/assetlinks.json`
+// Deprecated in favour of a safer approach, using path.join will resolve accounting the OS,
+// to concatente path strings is a well known bad practice in Node.js
+// --> const envPath = appRootPath.path + `/packages/client/public/.wellknown/assetlinks.json`
```
HexaField: these deprecation notices are unnecessary, we can just change it and not document the old code

HexaField (changes requested):

packages/server-core/src/projects/project/project-helper.ts
HexaField: this is likely what is causing checks to fail

jonathan-casarrubias: Good catch, I've already changed any call to copyDefaultProject to await for the promise to resolve, rebased, squached and pushed the updates, though while running the tests after addressing your comments I'm still getting the same bunch of errors I got yesterday, my assumption after analyzing the error stacks has been that these are unrelated to my modifications since in the stack hasn't include any file or function I created or modified, but after addressing your comments, I thought maybe these errors actually are very indirectly affected by my code, so I just tried running the tests from the dev branch and verify if there is any difference just to learn that I get the exact same errors I get while running the tests in my branch, that confirmed my assumption that if within the error stacks my code isn't mentioned, I also get the same errors on the dev branch, then these errors aren't caused by it.... Thoughts?

packages/client/scripts/populate-assetlinks.ts
jonathan-casarrubias: Documenting deprecated code is communication strategy that provides information such a the reason and the alternative, it's useful information that allow developers to better understand the changes and as a reference for future implementations, for now I'm simply addressing your change request but it's important to understand the importance of keeping deprecated code documented.

image

jonathan-casarrubias:
I'm looking at the failing checks here in github and contrary to what I get locally, I do see the project-helper file mentioned in the error stack.

[07:20:01.532] ERROR: ENOENT: no such file or directory, lstat '/home/runner/work/etherealengine/etherealengine/packages/projects/projects/default-project'
    component: "server-core:mysql"
    err: {
      "type": "Error",
      "message": "ENOENT: no such file or directory, lstat '/home/runner/work/etherealengine/etherealengine/packages/projects/projects/default-project'",
      "stack":
          Error: ENOENT: no such file or directory, lstat '/home/runner/work/etherealengine/etherealengine/packages/projects/projects/default-project'
      "errno": -2,
      "code": "ENOENT",
      "syscall": "lstat",
      "path": "/home/runner/work/etherealengine/etherealengine/packages/projects/projects/default-project"
    }
    
    it seems that the default-project isn't being copied, looking at the code I did modify that function, but only the part that removes the directory before trying to copy the default-project
    
image

the copyFolderRecursiveSync function still is the legacy one, I will focus on figuring out what is causing this, once finding the cause will push a fix

packages/client/scripts/populate-assetlinks.ts
HexaField: For our own code and endpoint APIs I agree, but this is a simple refactor. We should define what we should do in these situations in our standards / best practices documentation .

jonathan-casarrubias:
@HexaField 

I have finished coding this, it does correctly work locally but checks keep failing... So I added a bunch of logs for debugging where and why the flow is failing.

What I've found is that the problem is that the default-project source directory doesn't exist during the check process in the CI/CD, so there's nothing to be copied into the destination directory.

image

Also, it quite doesn't make sense why is needed to copy the default projects from

`etherealengine/packages/projects/projects/default-project`

to 

`etherealengine/packages/projects/default-project`

TBH it seems quite unnecessary IMO, can't the original source be used instead of copying 1 the exact same code into the parent directory?

scripts/create-project.ts
```diff
@@ -78,7 +78,7 @@ cli.main(async () => {
     }
 
     /** we used to use the project service create method here, but we shouldn't need to */
-    copyFolderRecursiveSync(templateFolderDirectory, projectsRootFolder)
+    cpAsync(templateFolderDirectory, projectsRootFolder)
```
HexaField: this needs to be awaited

HexaField (changes requested):

scripts/create-project.ts
jonathan-casarrubias: done... I'm still digging into the parent.remove error making the tests check to fail, following the destroyEngine path (first call in the error stack), even though my code isn't touching any of that by searching globally in the project, I found that sometimes it's being awaited and sometimes it isn't, this can be confirmed by using the IDE search tool for destroyEngine, will review in depth and see if that is the cause

packages/common/src/utils/fsHelperFunctions.ts
```diff
@@ -23,123 +23,137 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
-
-  // -- preparation to allow absolute paths as well
-  let root = ''
-  if (filepath[0] === '/') {
-    root = '/'
-    filepath = filepath.slice(1)
-  } else if (filepath[1] === ':') {
-    root = filepath.slice(0, 3) // c:\
-    filepath = filepath.slice(3)
+import logger from '../../../server-core/src/ServerLogger'
```
HexaField: the common package must not import from server-core. there is a common logger that should be used instead.

HexaField (changes requested):

HexaField:
Looking good now, I would like this QA'd on a deployment to make sure the builder and project installer works correctly

barankyle (approved):

barankyle:
> Looking good now, I would like this QA'd on a deployment to make sure the builder and project installer works correctly

Seemed to all work fine on QAT.

HexaField:
converted to a draft just while it doesnt have checks passing

PR #10190, opened 2024-05-18
Fixed bug in invalidation path encoding

barankyle:
## Summary

Only unsafe characters should be URL-encoded in Cloudfront invalidation paths. All others must be left as-is or else they invalidate the wrong path.

## References
closes #_insert number here_

### DIFF
--- a/scripts/run-batch-invalidation.ts
+++ b/scripts/run-batch-invalidation.ts
@@ -58,6 +58,24 @@ const options = cli.parse({
   startTime: [false, 'Timestamp of image', 'string']
 })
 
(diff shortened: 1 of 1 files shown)


DanielBelmes (approved):

HexaField (changes requested):
While this might be a good short term solution, we will want to support full UTF-8 at some point

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent#examples

Edit: ah, looks like this might be fine after all. worth testing thoroughly with a variety of characters

PR #10193, opened 2024-05-20
Move physics enter/exit back to reactors

HexaField:
## Summary

As part of the physics API refactor, the physics ECS component create/destroy functionality were moved from component reactors to enter/exit queries. This was due to insuficiencies in the gaurantees of scene loading.

This concern has since been addressed, and with the recent upgrades to hookstate that allow for a changing source to be hooked, we can simplify the component reactors. This includes changes to useAncestorWithComponent and useChildWithComponent that now react to changes in passed arguments.

## References
closes #_insert number here_


the-reality-engineer:
Seems to be a memory leak introduced here that is causing the heap to run out of memory.

HexaField:
> Seems to be a memory leak introduced here that is causing the heap to run out of memory.

Fixed. It was the change to useQuery.

PR #10198, opened 2024-05-20
[IR-2123] studio: smaller spinner

dtlehrer:
## Summary
[IR-2123] Make the loading spinner smaller when the files load in the file browser: https://tsu.atlassian.net/browse/IR-2123?atlOrigin=eyJpIjoiYzgyODRiNjcxNTI2NGQ2Mjk0ODdiMGM5N2NhZGM4OGYiLCJwIjoiaiJ9

## References
closes #_insert number here_

## QA Steps

[IR-2123]: https://tsu.atlassian.net/browse/IR-2123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dtlehrer:
@cla-bot check

packages/ui/src/primitives/tailwind/LoadingView/index.tsx
```diff
@@ -49,7 +49,7 @@ const LoadingView = ({
     <div role="status" className={twMerge('relative mx-auto my-0 block h-full w-full', className)}>
       <svg
         aria-hidden="true"
-        className="mx-auto my-auto h-full w-auto animate-spin fill-blue-600 text-gray-200 dark:text-gray-600"
+        className="mx-auto my-auto h-full max-h-12 w-auto max-w-12 animate-spin fill-blue-600 text-gray-200 dark:text-gray-600"
```
aditya-mitra: If we make changes to the LoadingView here, all the components using it in studio and admin will be affected.
I think we can change the [outer div's size](https://github.com/EtherealEngine/etherealengine/pull/10198/files#diff-070472ad4d673917a2667c6f60451dfc71dc14b3a8f1f3707075b9afdfd4ee21R49), the inner svg size will also change

aditya-mitra (changes requested):

packages/ui/src/primitives/tailwind/LoadingView/index.tsx
dtlehrer: I noticed that there are some other areas where this same component also appeared to be stretched (one example in video attached below).

Figured it may be worthwhile to just globally constrain the size to prevent this from happening

https://github.com/EtherealEngine/etherealengine/assets/19999194/c44e24a9-6474-4d28-8294-2f82337eeeca

HexaField: @dtlehrer the ui package is just a library of UI elements, they should not make assumptions about where they are used or what constaints they have, they should just have minimal styling to provide the functionality we need. Any specific constaints either need to be passed in and handled in some merge function, or handled in an external div.

PR #10205, opened 2024-05-21
refactor: Update Primus initialization to include pathname in server URL

MoizAdnan:


HexaField (approved):

HexaField (changes requested):
do we need to do this for instance server connections too?

MoizAdnan:
> do we need to do this for instance server connections too?

I am not sure, how to check?

HexaField:
> > do we need to do this for instance server connections too?
> 
> I am not sure, how to check?

Client connects here
https://github.com/EtherealEngine/etherealengine/blob/dev/packages/client-core/src/transports/SocketWebRTCClientFunctions.ts#L275

Instance server receives connection here
https://github.com/EtherealEngine/etherealengine/blob/dev/packages/instanceserver/src/channels.ts#L609

MoizAdnan:
> > > do we need to do this for instance server connections too?
> > 
> > 
> > I am not sure, how to check?
> 
> Client connects here https://github.com/EtherealEngine/etherealengine/blob/dev/packages/client-core/src/transports/SocketWebRTCClientFunctions.ts#L275
> 
> Instance server receives connection here https://github.com/EtherealEngine/etherealengine/blob/dev/packages/instanceserver/src/channels.ts#L609

We looked into it, and we do not need pathName in instance server connections

hanzlamateen (approved):

HexaField (approved):

PR #10208, opened 2024-05-21
IR-1887-Asset-Preview-breaking-if-you-click-gltf-that-is-in-the-scene

MichaelEstes:
## References
closes #_insert number here_


packages/engine/src/gltf/GLTFState.test.tsx
```diff
@@ -649,4 +655,73 @@ describe('GLTFState', () => {
...
+    }
+
+    Cache.add('/test.gltf', gltf)
+
+    const gltfEntity = GLTFSourceState.load('/test.gltf')
+
+    applyIncomingActions()
+
+    const sceneID = getComponent(gltfEntity, SourceComponent)
+    const GLTFSnapshotStateReactor = GLTFSnapshotState.reactor
```
HexaField: this reactor will run synchronously in `applyIncomingActions` as it is an event sourced state, so calls to run it manually are unnecessary

HexaField (approved):

PR #10213, opened 2024-05-21
IR-2080 CSM Fixes

DanielBelmes:
## References
closes #_insert number here_


DanielBelmes:
comment from @HexaField **"instead of passing the rendererSettingsEntity down, we should rearrange it such that rather than querying for the RendererSettingscomponent, and mounting a CSM reactor tree for each, it should be for each RendererComponent we want a CSM reactor tree, that uses useChildWithComponent to get an entity in the scene with a render settings component."**

dinomut1 (changes requested):
This does make CSM reactive at least to the properties, but it also breaks it further.

Here is loading electronics store scene on dev. You can see for a moment CSM is there with somewhat correct values before getting overridden or something:
[Screencast from 05-23-2024 07:01:56 PM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/3edfcba5-240c-4d37-aeaf-cb063af2b83a)

Here is the same scene when loaded in this branch:
[Screencast from 05-23-2024 07:03:13 PM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/97e4928c-daff-4e6b-9c77-7cf42ce5d0c5)

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -226,11 +229,13 @@ const PlainCSMReactor = (props: { rendererEntity: Entity }) => {
 }
 
 const ChildCSMReactor = (props: { rendererEntity: Entity }) => {
-  const entities = useTreeQuery(props.rendererEntity)
+  const entities = useHookstate(useTreeQuery(props.rendererEntity))
```
HexaField: wrapping useHookstate around this is an anti pattern. the data already updates reactively

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -246,17 +251,19 @@ const EntityChildCSMReactor = (props: { entity: Entity; rendererEntity: Entity }
...
 
-    const objs = [...groupComponent.value] as Mesh<any, Material>[]
-    for (const obj of objs) {
-      if (obj.material) {
-        csm.setupMaterial(obj)
-      }
-    }
-
-    return () => {
+    if (groupComponent) {
```
HexaField: let's do an early opt out rather than nested

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -266,16 +273,17 @@ const EntityChildCSMReactor = (props: { entity: Entity; rendererEntity: Entity }
 }
 
 function _CSMReactor() {
-  const renderSettingsEntity = useEntityContext()
-  const rendererEntity = useScene(renderSettingsEntity)
+  const rendererEntity = useEntityContext()
 
   if (!rendererEntity) return null
+  const renderSettingsEntity = useChildWithComponent(rendererEntity, RenderSettingsComponent)
```
HexaField: the same number of hooks must always exist in a reactor, move the early escape after all hooks

HexaField (changes requested):

packages/engine/src/scene/systems/ShadowSystem.tsx
DanielBelmes: Dang. I'll try again but this passes an UndefinedEntity to useComponent and causes some infintie suspend behavior.

packages/engine/src/scene/systems/ShadowSystem.tsx
DanielBelmes: Then I believe we have a bug within useTreeQuery. Specifically that we map through it's results with the Number constructor. I believe we might be derefrencing the results. I gotta test today.

packages/engine/src/scene/systems/ShadowSystem.tsx
DanielBelmes: Oh sorry misunderstood. Was confusing some changes I never pushed with this comment

packages/engine/src/scene/systems/ShadowSystem.tsx
DanielBelmes: nvm I was dumb with keys

dinomut1 (approved):
working better than what we have. Still have weird cascade offset though

PR #10123, opened 2024-05-03
IR-1830 - Add project id to location and instance table

hurairahmateen:
## Summary
- Add project id to location and instance table.
- Created migrations to add projectId column.
- Updated service calls to include projectId

PR #10214, opened 2024-05-21
IR-2153: replace scope checks with project permission checks

jose-galvan:
## Summary

IR-2153 & IR-2154

Replace scope checks in studio for project permission checks. 

## References
closes #_insert number here_

## QA Steps
- Log in with a user that doesn't have the location:write scope. 
- In the dashboard click on any scene in any project
- Once in the studio click on the Publish button on the upper right corner 
- Indicate a name for the location
- Click on submit. it should create a location.


packages/server-core/src/projects/project-permission/project-permission.hooks.ts
```diff
@@ -205,6 +206,41 @@ const makeRandomProjectOwner = async (context: HookContext<ProjectPermissionServ
...
+ * @param context
+ * @returns
+ */
+const resolveProjectId = async (context: HookContext<ProjectPermissionService>) => {
+  if (!context.params.query?.project) {
+    return
+  }
+
+  const projectResult = (await context.app.service(projectPath).find({
+    query: { name: context.params.query?.project, $limit: 1 }
```
hanzlamateen: I dont see any `project` field in query in project-permission schema

packages/server-core/src/projects/project-permission/project-permission.hooks.ts
```diff
@@ -205,6 +206,41 @@ const makeRandomProjectOwner = async (context: HookContext<ProjectPermissionServ
   if (context.id && context.result) await context.service.makeRandomProjectOwnerIfNone(result[0].projectId)
 }
 
+/**
+ * Filter permissions for logged in user
+ * @param context
+ * @returns
+ */
+const filterUserPermissions = async (context: HookContext<ProjectPermissionService>) => {
```
hanzlamateen: Do we need this hook? Can we use `@etherealengine/server-core/src/hooks/set-loggedin-user-in-query`?

hanzlamateen (commented):
Make sure to test this code using an incognito window with complete signup, project creation and then publishing from studio?

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
```diff
@@ -90,15 +91,17 @@ const LocationDrawer = ({ open, mode, selectedLocation, selectedScene, onClose }
...
-  const user = useHookstate(getMutableState(AuthState).user)
 
   const locationMutation = useMutation(locationPath)
 
-  const hasWriteAccess = user.scopes.get(NO_PROXY)?.find((item) => item?.type === 'location:write')
   const viewMode = mode === LocationDrawerMode.ViewEdit && !editMode.value
 
   const selectedSceneData = useGet(assetPath, selectedScene!)
 
+  const editorState = getState(EditorState)
```
hanzlamateen, line 100: Since this drawer is used in admin panel as well.

Maybe it would be better if you just have a prop in this component as disabledEditButton. And then from studio's `PublishLocation.tsx` you pass that boolean? As I think you are already checking this.

In this control, if value of `disabledEditButton` is undefined then the button should work as it normal works

hanzlamateen (changes requested):

packages/server-core/src/projects/project-permission/project-permission.hooks.ts
jose-galvan: per our conversation this has been added to projectPermissionQuerySchema in order to resolve the project id and will be dropped from the query when the project id resolved.

packages/server-core/src/projects/project-permission/project-permission.hooks.ts
jose-galvan: it's not needed, I replaced it with `set-loggedin-user-in-query`

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
jose-galvan: Done! it's now passed as a property

packages/server-core/src/social/location/location.hooks.ts
```diff
@@ -210,23 +213,49 @@ export default {
...
       persistData,
-      discard('locationSetting')
+      discard('locationSetting'),
+      discard('action')
     ],
     remove: [
-      iff(isProvider('external'), verifyScope('location', 'write')),
+      when(
+        isProvider('external'),
+        iff(
```
hanzlamateen: @HexaField it seems like we need to make projectId part of location in core engine instead of MT. We did initially in https://github.com/EtherealEngine/etherealengine/pull/10123

As in this scenario, it seems like we need projectId to check for project permissions. Passing it from client which is not desired.

Ref:
`packages/projects/projects/eepro-multitenancy/services/hooks/social/location.hooks.ts`

hanzlamateen (changes requested):

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
```diff
@@ -211,13 +212,27 @@ const LocationDrawer = ({ open, mode, selectedLocation, selectedScene, onClose }
 
     if (validateForm(state.value, state.formErrors.value)) {
       if (mode === LocationDrawerMode.Create) {
-        locationMutation.create(data).catch((error) => {
-          NotificationService.dispatchNotify(error.message, { variant: 'error' })
-        })
+        locationMutation
+          .create(data, {
+            query: {
+              action: 'studio',
```
hanzlamateen: As discussed below in https://github.com/EtherealEngine/etherealengine/pull/10214/files#r1616936843 we dont need to pass this action and project props here.

ProjectId is already part of location table.

Studio action is not required because the service should work in a combination of scope and permissions instead of 1 for admin and 1 for the studio.

hanzlamateen (changes requested):

packages/server-core/src/projects/project-permission/project-permission.hooks.ts
```diff
@@ -218,7 +221,13 @@ export default {
       () => schemaHooks.validateQuery(projectPermissionQueryValidator),
       schemaHooks.resolveQuery(projectPermissionQueryResolver)
     ],
-    find: [enableClientPagination(), iff(checkUserScopes, checkPermissionStatus)],
+    find: [
+      enableClientPagination(),
+      iff(checkUserScopes, checkPermissionStatus),
+      iff(isAction('studio'), setLoggedInUserInQuery('userId'), resolveProjectId()),
```
hanzlamateen: Again here instead of action and project. We should make this generic to any client request

hanzlamateen (changes requested):

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
jose-galvan: Done! removed action from query.

packages/server-core/src/projects/project-permission/project-permission.hooks.ts
jose-galvan: removed action from here as well

packages/client-core/src/user/useUserProjectPermission.tsx
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * @param path
+ * @param queryParams
+ * @param options
+ * @returns
+ */
+export const useProjectPermissions = (project: string, action: string): ProjectPermissionType => {
+  const { data } = useFind(projectPermissionPath, {
+    query: {
+      project,
```
hanzlamateen, line 40: Do we need to pass `project` name here? @HexaField do we have projectId anywhere in studio's state? That will remove having need to update project-permission service to use project name when it can natively work for project id?

packages/common/src/schemas/social/location.schema.ts
```diff
@@ -52,6 +52,9 @@ export const locationSchema = Type.Object(
     }),
     name: Type.String(),
     sceneId: Type.String(),
+    projectId: Type.String({
```
hanzlamateen: This wouldnt be required when you pull dev into this branch

packages/common/src/schemas/social/location.schema.ts
```diff
@@ -94,6 +97,7 @@ export const locationQueryProperties = Type.Pick(locationSchema, [
   'id',
   'name',
   'sceneId',
+  'projectId',
```
hanzlamateen: same for this

packages/editor/src/components/toolbar/tools/PublishLocation.tsx
```diff
@@ -82,7 +82,7 @@ export const PublishLocation = () => {
           <Button
             onClick={handleOpenLocationDrawer}
             className={styles.toolButton}
-            disabled={!sceneID.value || !hasWriteAccess}
+            disabled={!sceneID.value || (permission?.type !== 'owner' && permission?.type !== 'editor')}
```
hanzlamateen: I think we can keep the support of hasWriteAccess as well. As we want to support both project permissions or location:write

packages/server-core/src/hooks/verify-project-permission.ts
```diff
@@ -0,0 +1,70 @@
...
+export default (types: string[]) => {
+  return async (context: HookContext<Application>) => {
+    if (context.params.isInternal) return context
+
+    const loggedInUser = context.params.user as UserType
+
+    if (!loggedInUser) throw new NotAuthenticated('No logged in user')
+
+    const projectId =
+      context.service === 'project' ? context.id : context.params.query.projectId ?? context.data.projectId
```
hanzlamateen: What is the purpose of this `context.service === 'project` check here? I think `context.service` is an object rather than string.

packages/server-core/src/projects/project-permission/project-permission.hooks.ts
```diff
@@ -218,7 +219,12 @@ export default {
       () => schemaHooks.validateQuery(projectPermissionQueryValidator),
       schemaHooks.resolveQuery(projectPermissionQueryResolver)
     ],
-    find: [enableClientPagination(), iff(checkUserScopes, checkPermissionStatus)],
+    find: [
+      enableClientPagination(),
+      iff(checkUserScopes, checkPermissionStatus),
+      resolveProjectId(),
+      discardQuery('project')
```
hanzlamateen: If we have project id in studio then we dont need this overhead of passing project name and then getting id from it.

Otherwise it looks good.

packages/server-core/src/social/location/location.hooks.ts
```diff
@@ -200,6 +202,27 @@ const duplicateNameError = async (context: HookContext<LocationService>) => {
   }
 }
 
+/**
+ * Populate projectId from sceneId into location data
+ * @param context
+ */
+const populateProjectIdInData = async (context: HookContext) => {
```
hanzlamateen: This is not required as PR for this is already merged

packages/server-core/src/social/location/location.hooks.ts
```diff
@@ -210,23 +233,24 @@ export default {
     find: [discardQuery('action'), discardQuery('studio'), sortByLocationSetting],
     get: [],
     create: [
-      iff(isProvider('external'), verifyScope('location', 'write')),
       () => schemaHooks.validateData(locationDataValidator),
       schemaHooks.resolveData(locationDataResolver),
+      iff(isProvider('external'), verifyScope('editor', 'write'), verifyProjectPermission(['owner', 'editor'])),
```
hanzlamateen: Can this be made to work in hybrid with either location:write OR project permission with editor:write? Same goes for other methods in this file.

@HexaField what are your thoughts?

packages/server-core/src/social/location/location.resolvers.ts
```diff
@@ -106,6 +107,14 @@ export const locationDataResolver = resolve<LocationType, HookContext>({
       updatedAt: await getDateTimeSql()
     }
   },
+  projectId: async (value, location, context) => {
```
hanzlamateen: Not required as PR already merged

packages/server-core/src/social/location/migrations/20240529044849_project-id-column.ts
```diff
@@ -0,0 +1,58 @@
+/*
```
hanzlamateen: This file is not required as PR is already merged

hanzlamateen (changes requested):

packages/server-core/src/social/location/location.hooks.ts
jose-galvan: you cannot verify if a user has one scope or the other cause verifyScope will throw a Forbidden error the moment one fails not allowing the second check to take place. for that we'll have to modify the hook and that has impact on other hooks.

packages/server-core/src/projects/project-permission/project-permission.hooks.ts
jose-galvan: No we don't, I had already explained this, we only use project name in studio. 
image

packages/client-core/src/user/useUserProjectPermission.tsx
jose-galvan: No projectId in state. 
image

HexaField: we don't unfortunately

packages/server-core/src/social/location/location.hooks.ts
hanzlamateen: ~~Well there is a checkScope method which returns boolean instead of throw. You can make use of it. [Here](https://github.com/EtherealEngine/etherealengine/blob/4bf3b5166a5ad5cf7706eb58d8b43b52441fbf2c/packages/server-core/src/user/user-avatar/user-avatar.hooks.ts#L63) is the link of code where its used~~



~~Alternatively you can:~~

~~1. Create a new hook which does what verifyScope does, and rather than throwing it returns true false?
2. Update verifyScope hook to have an argument which has a param of throw = true by default. And if you change it to false, then it will return boolean?~~

hanzlamateen: I just created a new hook in following PR, you can use it in your code it returns true/false instead of throwing an error. Hope it helps?

https://github.com/EtherealEngine/etherealengine/pull/10279

packages/editor/src/components/toolbar/tools/PublishLocation.tsx
hanzlamateen: Resolved in https://github.com/EtherealEngine/etherealengine/pull/10214/commits/a6899a2bf32b291e1bbb3d003c2839935b59836b

packages/server-core/src/social/location/location.hooks.ts
hanzlamateen: Resolved in https://github.com/EtherealEngine/etherealengine/pull/10214/commits/112195766f42fe72364f2acdfaab9e6de1b9e189

hanzlamateen (approved):
I have made the changes in this PR to work with hybrid of:
- location:write scope OR
- editor:write scope with owner/editor project permission

I do made some changes in `project-permission` hooks as well.

I have manually tested all possible workflows through studio and they appear to be working fine.

@barankyle it would be great if you can have a quick look of these changes particularly `project-permission` hook ones.

barankyle (approved):

PR #10215, opened 2024-05-22
Improved GLTF Node Dictionary and refactored UUIDs

HexaField:
## References
closes #_insert number here_


HexaField:
some optimization can be made as a cleanup later

dinomut1 (changes requested):
In the studio, entering the "assets" folder in the file browser crashes the studio

HexaField:
> In the studio, entering the "assets" folder in the file browser crashes the studio

Is this bug pre-existing?

EDIT: hmm looks like it is in this branch unfortunately

dinomut1:
> > In the studio, entering the "assets" folder in the file browser crashes the studio
> 
> Is this bug pre-existing?
> 
> EDIT: hmm looks like it is in this branch unfortunately

i think it's to do with the asset preview

HexaField:
@dinomut1 @AidanCaruso @DanielBelmes The core refactor of this should be fairly solid - default scenes, studio authoring & multiplayer seems to work fine.

There are some things I don't know how to test around dererferncing & material library, so please test these if you have time. 

The main part left is changing how entities are referenced and resolved in the components that store references to other entities. So far the components identified that do this are Triggers, Render Settings, Interactables, Envmap, Spline, Media, Basic tutorial, Pong tutorial, LLM, XRUI tools

dinomut1 (changes requested):
This hard-breaks the concept of dereferencing, or any node moving between documents without having their simulation entity destroyed. We need to rethink our approach here

packages/engine/src/gltf/AuthoringHookState.tsx
the-reality-engineer, line 1: Is there a reason why the GLTFCallbackState.tsx version of this that Josh wrote is inadequate? The names and terminologies used in that one are much clearer to me, and avoid overloading existing terminologies (e.g. hooks), with different meanings.

the-reality-engineer (changes requested):

PR #10212, opened 2024-05-21
IR-1652 / IR-1986 Interactable Input Refactor

the-reality-engineer:
## Summary

IR-1652
IR-1986

ClientInputSystem - Routes input from InputSourceComponent, through InputComponent, to InputSinkComponent
  1. Each InputSource raycasts into the scene
  2. The hit results (meshes and/or rigidbodies) are traversed upwards to find the closest InputComponent
  3. The InputComponent explicitly routes input to itself and/or another entity

Various new hooks that allow entities to easily consume routed input:

`InputComponent.useExecuteWithInput(executeOnInput:(inputEid:Entity) => {}) : Entity`
 - walks up the entity tree, finding closest InputSinkComponent, and grabs inputEid from that

` InputComponent.useHasFocus()`
 - returns true if the entity input context has inputSources array length of non-zero.

PR #10219, opened 2024-05-22
Ir 1652 interactable input refactor

SamMazerIR:
## Summary
code suggestions and file location fixes for merged branch from PR:
[https://github.com/EtherealEngine/etherealengine/pull/10212](https://github.com/EtherealEngine/etherealengine/pull/10212)


packages/spatial/src/input/systems/ClientInputSystem.tsx
```diff
@@ -291,7 +285,7 @@ const execute = () => {
       let closestDistanceSquared = Infinity
 
       //use sourceEid if controller (one InputSource per controller), otherwise use avatar rather than InputSource-emulated-pointer
-      const selfAvatarEntity = AvatarComponent.getSelfAvatarEntity()
+      const selfAvatarEntity = UUIDComponent.getEntityByUUID((Engine.instance.userID + '_avatar') as EntityUUID) //would prefer a better way to do this
```
the-reality-engineer, line 288: For a future reference, one strategy (which I think @HexaField  suggested at some point), would be to inject these heuristics from higher level packages, so perhaps the `world` (currently `engine`) package injects the additional proximity heuristic.

the-reality-engineer (approved):

packages/spatial/src/input/systems/ClientInputSystem.tsx
SamMazerIR: I would love to learn more, not sure from the description what that would look like

HexaField: why does this system care about the avatar? what is the early scape in the for loop doing?

HexaField (approved):

packages/spatial/src/input/systems/ClientInputSystem.tsx
the-reality-engineer: Because one of the heuristics for focusing non-spatial input sources on entities that can receive input is some notion of proximity to those entities (esp when not pointing w/ the mouse cursor)

PR #10220, opened 2024-05-22
Optimized publishing duplication

internetcharles:
## Summary
Scene duplication added on publish.

## References
closes #_insert number here_


packages/client-core/src/admin/common/Location/LocationDrawer.tsx
```diff
@@ -181,8 +194,54 @@ const LocationDrawer = ({ open, mode, selectedLocation, selectedScene, onClose }
...
+
+    handleClose()
+  }
+
+  const publishScene = async (bake: boolean) => {
+    // TODO: Save current scene in place, and create new version to publish
+    const { sceneAssetID, projectName, sceneName, scenePath } = getState(EditorState)
+
+    if (projectName && sceneName && scenePath) {
+      const sceneQuery = useFind(assetPath, { query: { assetURL: scenePath ?? '' } }).data
```
dinomut1, line 229: This throws an invalid hook call. Must use useFind in a reactive context

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
```diff
@@ -181,8 +194,54 @@ const LocationDrawer = ({ open, mode, selectedLocation, selectedScene, onClose }
...
+      const sceneURL = sceneQuery?.[0]?.assetURL
+      const abortController = new AbortController()
+
+      // TODO: Bake implementation
+      if (bake) {
+        await saveSceneGLTF(sceneAssetID, projectName, sceneName + '-optimized', abortController.signal)
+      }
+
+      setCurrentEditorScene(sceneURL, sceneQuery[0].id! as EntityUUID)
```
dinomut1, line 238: This sets the current editor scene to the same scene it is currently in. Need to set current scene to the newly created copy

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
```diff
@@ -181,8 +194,54 @@ const LocationDrawer = ({ open, mode, selectedLocation, selectedScene, onClose }
...
+    }
+
```
dinomut1, line 240: Everything after this point needs to wait for the scene change to complete and the new scene to finish loading. Use a useEffect and wait on the root scene entity to complete loading before proceeding with mesh baking

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
dinomut1, line 238: Also, this needs to be in the "bake" conditional as well, because if we're not baking we have no need to change editor scene

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
```diff
@@ -181,8 +194,54 @@ const LocationDrawer = ({ open, mode, selectedLocation, selectedScene, onClose }
...
+    // TODO: In duplicated scene, perform mesh baking to de-reference all models in the scene (saving scene as GLTF & fuse/compress scenes)
+    // TODO: Publish the duplicated scene
+
```
dinomut1, line 243: Here is where the "bake" conditional should end

packages/client-core/src/admin/common/Location/LocationDrawer.tsx
```diff
@@ -202,28 +261,13 @@ const LocationDrawer = ({ open, mode, selectedLocation, selectedScene, onClose }
...
 
-    state.formErrors.merge({
-      name: state.name.value ? '' : t('admin:components.location.nameCantEmpty'),
-      maxUsers: state.maxUsers.value ? '' : t('admin:components.location.maxUserCantEmpty'),
-      scene: state.scene.value ? '' : t('admin:components.location.sceneCantEmpty'),
-      type: state.type.value ? '' : t('admin:components.location.typeCantEmpty')
-    })
+    const onError = (error) => NotificationService.dispatchNotify(error.message, { variant: 'error' })
 
-    if (validateForm(state.value, state.formErrors.value)) {
```
dinomut1, line 212: Don't delete the validateForm function, this removes notifications about incorrectly entered fields

dinomut1 (changes requested):

PR #10222, opened 2024-05-22
Dereference prefabs

dinomut1:
Moves prefab gltfs into assets folder of default project.
Creates simple authoring hook system which lets callbacks be assigned to operate on a given entity once it has been loaded into a document's authoring layer.
Uses authoring hook to dereference prefabs once they are loaded into the scene


JT00y (approved):
the prefab path is set to project/default-project/asset/prefab at #10242 
can you merge this pr first then #10242

PR #10231, opened 2024-05-23
corrected white wireframe not visible issue

MbfloydIR:
updated the background to use the color with when in wireframe render mode
updated the wireframe material to use the color black when in wireframe render mode

## References
closes #_insert number here_


HexaField (approved):

packages/spatial/src/renderer/functions/changeRenderMode.ts
```diff
@@ -82,6 +82,7 @@ export function changeRenderMode() {
       break
     case RenderModes.WIREFRAME:
       renderPass.overrideMaterial = new MeshBasicMaterial({
+        color: 0x000000,
```
DanielBelmes, line 85: Wonder why this didn't work when we tried it earlier

PR #10233, opened 2024-05-23
IR-1652 input sink fixes

SamMazerIR:
## Summary
fixing InputSink behavior, defaults and serialization, also fixing issue surrounding XrScenePlacement entity not being uniquely identifiable to exclude (now excluded from a specific query it needed to be - preventing issues with proximity detection of InputComponents at origin - or next to that entity)


packages/spatial/src/input/systems/ClientInputSystem.tsx
```diff
@@ -132,7 +133,13 @@ const worldPosInputComponent = new Vector3()
 const inputXRUIs = defineQuery([InputComponent, VisibleComponent, XRUIComponent])
 const inputBoundingBoxes = defineQuery([InputComponent, VisibleComponent, BoundingBoxComponent])
 const inputObjects = defineQuery([InputComponent, VisibleComponent, GroupComponent])
-const spatialInputObjects = defineQuery([InputComponent, VisibleComponent, TransformComponent, Not(CameraComponent)]) //TODO may be overkill if visible means it always has transform
+const spatialInputObjects = defineQuery([
+  InputComponent,
+  VisibleComponent,
+  TransformComponent,
+  Not(CameraComponent),
+  Not(XRScenePlacementComponent)
```
HexaField, line 141: ideally in the future we will figure out how to not have to specify many opt-outs here, but it's acceptable for now

SamMazerIR: I agree

the-reality-engineer (approved):

PR #10234, opened 2024-05-24
Updated Cloudfront function to pass most things as routes

barankyle:
## Summary

Updated Cloudfront function to pass most things as routes
    
Project-defined routes were not being handled properly by the Cloudfront function;
the function was not properly handling subroutes of the registered route, e.g.
/newroute/subroute would throw a 403 because the CF function didn't know how to
handle anything other than /newroute.

Rather than make the function conditionally handle subroutes, it was determined that
it would be easier to instead have /projects and /recordings paths be passed through
as-is, and have everything else that wasn't a static client file be passed as if it
were a route, i.e. get directed to client/index.html and let the router there determine
what to do with it (or nothing, if it's not a registered route).

## References
[Enhances IR 2260](https://tsu.atlassian.net/browse/IR-2260)


packages/client/src/route/customRoutes.tsx
```diff
@@ -44,8 +44,12 @@ const CustomRoutes = ({ customRoutes }) => {
     <Suspense fallback={<LoadingCircle message={t('common:loader.loadingCustom')} />}>
       <Routes>
         {matchedRoutes.map((route, i) => {
-          const { route: r, component, props: p } = route
+          let { route: r, component, props: p } = route
           const Element = component as any
+          // e.g. /places/* will get truncated to /places, /places* will get truncated to /places
```
hanzlamateen: I wonder if line 56 will have any impact with this?

as its looking for `r.split('/')[1]` and what if the path in project is like `/dashboard/*`

image

hanzlamateen (changes requested):

packages/client/src/route/customRoutes.tsx
barankyle: I read a bit more about how React-Router works and realized we didn't need much of the split section on what was line 56 any more. I think that was done as a cruder way of matching sub-routes, but with this new pattern, we only need to rewrite splats without slashes. `/dashboard/*` is actually the [splat](https://www.diaart.org/exhibition/exhibitions-projects/robert-smithson-exhibition) format the React-Router uses, so that can be used as-is, and while it'll handle `/dashboard*`, it throws a warning, so I decided to add the slash for that situation. And now we can say that anyone using `/dashboard` is literally only wanting that exact route, and not any subroutes, so we can use that without adding a splat to the end.

hanzlamateen: @barankyle or what the way forward as @kattsushi is waiting for this PR to fix routing on MT.

barankyle: I fixed some issues with this yesterday, it should be working. Just want to make sure the new route schema is what we want.

hanzlamateen:
@barankyle I am just wondering why we have this cloudfront function in place which restricts routing and rather why not allow all requests and then have our react-route/vite decide which page exists and redirect the ones that does not exist.

We can have an exception of /static folder which will serve assets. Maybe we need favicon, etc files to be in static folder instead of root public folder.

Because there can be essentially undefined amount of routes we can support once we implement this comment https://tsu.atlassian.net/browse/IR-2130?focusedCommentId=39265

barankyle:
> @barankyle I am just wondering why we have this cloudfront function in place which restricts routing and rather why not allow all requests and then have our react-route/vite decide which page exists and redirect the ones that does not exist.
> 
> We can have an exception of /static folder which will serve assets. Maybe we need favicon, etc files to be in static folder instead of root public folder.
> 
> Because there can be essentially undefined amount of routes we can support once we implement this comment https://tsu.atlassian.net/browse/IR-2130?focusedCommentId=39265

Because we serve both the static client files and front-end project files from the same bucket & CF distro, and because of the directory structure we use to do so, we need this CF function to correctly route traffic. There's a couple of different categories of traffic that CF will handle, and they need to be handled in their own way.

We store all of the static files that vite builds in the `/client` folder so that the bucket is better organized. project files have their paths unchanged, which means they all end up in `/projects`, and recordings are placed in `/recordings`. We could dump all of the client files in the top level of the bucket, but I find that messy to navigate through.

Requests for client files therefore need to be rerouted by this function. e.g., if the client is requesting /assets/index-BHFRHRFR.js, the function will reroute the request to /client/assets/index-BHFRHRFR.js. I feel like this rule is simple enough; when building the front-end files, we just grab all the top-level paths, including folders, in client/dist and say that those need to have `client/` prepended on the path.

Requests for engine routes all need to be handled by client/index.html, as our routing logic that is located in the root page is what determines which front-end components get rendered based on the route. Ordinarily, if someone hits /location/admin, CF would just try to pass that to a file located at /location/admin in S3, which doesn't exist. At one point I'd considered copying the index.html file to all of the route paths in S3, but this doesn't work for things like /location/<location>, since having a copy of the file for every location seems convoluted. So, what seemed like the best option was to explicitly reroute route-related paths to client/index.html.

Requests for project files don't need to be changed, since by convention they're all already in /projects/<projectName>.

Under the current function structure, we can't just let anything that's not a static file go to client/index.html since all the project files should not be handled by the front-end client page. An alternative this has brought up is that maybe we can just pass-through anything starting with /projects or /recordings, and have the default function behavior be to pass it to client/index.html.

barankyle:
I ended up implementing the alternative I'd mentioned in the previous comment, as it makes things a lot simpler (at the slight expense of not being able to specify e.g. subroute-only routes, e.g. /places/*). 

The CF function now operates like this:

* If the path is to a static file, point to `/client/<original path>` (Unchanged from before)
* If the path is to /projects or /recordings, do nothing to the path, and it'll resolve to whatever it was going to
* If neither of the above is true, assume it's something to be handled by a route, and pass it to client/index.html

I checked that this worked on QAT, which it did. No changes need to be made to any of the routes in any project's xrengine.config.ts, nor should any other changes need to be made to existing projects.

hanzlamateen:
That sounds great. Do we still need this PR?

hanzlamateen (approved):

hanzlamateen:
This new approach sounds great.

PR #10235, opened 2024-05-24
fix model transform in model component editor

dinomut1:
## References
closes #_insert number here_


packages/editor/src/components/properties/ModelTransformProperties.tsx
```diff
@@ -140,7 +140,7 @@ export default function ModelTransformProperties({ entity, onChangeModel }: { en
 
       for (const variant of variants) {
         if (clientside) {
-          await clientSideTransformModel(variant as ModelTransformParameters)
+          nuPath = await clientSideTransformModel(variant as ModelTransformParameters)
```
HexaField, line 143: what does `nu` mean? can we use full words?

dinomut1: haven't you ever heard of nu metal?

PR #10244, opened 2024-05-24
Update prefab gltf paths

CITIZENDOT:
## Summary

Sets the correct gltf paths in prefabs.

## References
closes #_insert number here_


dinomut1 (commented):
we're putting the prefabs into the assets folder instead of updating these paths

dinomut1:
see #10222

PR #10246, opened 2024-05-24
IR-1439-Scene-Render-Quality-Dropping-Unexpectedly-with-Automatic-Render-Settings

MichaelEstes:
## Summary

- Makes render scale a part of the performance state quality settings
- Smooths out timings between automatic quality adjustments
- Locks quality changes behind automatic flag
- Changes render resolution slider to change the quality preset instead of just resolution
- Updates sync with local storage to store keys that are set to false 

## References
closes #_insert number here_


packages/hyperflux/functions/StateFunctions.ts
```diff
@@ -147,8 +147,10 @@ export function syncStateWithLocalStorage<S, E extends Identifiable>(
       onSet: (state, desc) => {
         for (const key of keys) {
           const storageKey = `${stateNamespaceKey}.${rootState.identifier}.${key}`
-          if (!rootState[key] || !rootState[key].get(NO_PROXY)) localStorage.removeItem(storageKey)
-          else localStorage.setItem(storageKey, JSON.stringify(rootState[key].get(NO_PROXY)))
+          const value = rootState[key]?.get(NO_PROXY)
+          // We should still store flags that have been set to false
+          if (value === null || value === undefined) localStorage.removeItem(storageKey)
```
HexaField: we should only check for undefined and still store null, such that values that have a default value that is not null but is nullable still are set. undefined should never be used in hookstate so we can ensure that key is removed

packages/spatial/src/renderer/PerformanceState.ts
```diff
@@ -143,56 +200,78 @@ export const PerformanceState = defineState({
     }, [performanceState.gpuTier])
 
     useEffect(() => {
-      if (isEditing) return
+      performanceState.isSmoothing.set(true)
+      recreateEMA()
+      setTimeout(() => {
```
HexaField: instead of a timeout could we use an accumulator or drain, maybe a useExecute would suffice? can do that in a followup if this is urgent

HexaField (approved):

PR #10247, opened 2024-05-24
[IR-2055] Unit Test - Camera

heysokam:
## Summary
- [x] `CameraEntityState` event sourced state
- [x] `SpectateEntityState` event sourced state


heysokam:
@HexaField @DanielBelmes
There must be something wrong with the networking setup for this event sourced unit test, but I don't know how to start finding what could be incorrect.
Could one of you take a look to see why it is failing?

HexaField:
we might need to wrap `flushSync` around everything in `applyIncomingActions` to ensure all event sources are resolved to completion, otherwise it could be taking two `applyIncomingActions` as the ordering of event sources registration order may be causing the problem here

HexaField (approved):

PR #10251, opened 2024-05-26
studio: center dropdown arrow

achen5671:
## References
closes #_insert number here_


packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -145,7 +145,7 @@ const Select = <T extends OptionValueType>({
       <MdOutlineKeyboardArrowDown
         size="1.5em"
         className={`text-theme-primary absolute right-3 transition-transform ${showOptions.value ? 'rotate-180' : ''} ${
-          label ? 'top-8' : 'top-2'
+          label ? 'top-8' : 'top-1'
```
aditya-mitra: our dropdown arrow is absolutely positioned. can we reposition it as in a `block` manner?
(it should be such that the dropdown arrow is vertically centred with any height)

aditya-mitra (changes requested):

packages/ui/src/primitives/tailwind/Select/index.tsx
achen5671: I ran into some issues with spacing while using position block. I did it with absolute positioning with transform. This should also work. (centered with any height)

aditya-mitra:
Closing in favour of #10278 so that we can merge this to dev

PR #10252, opened 2024-05-27
Guarantee immediate scheduling of async hook effects 

the-reality-engineer:
## Summary

Okay, so using useImmediateEffect is critical whenever we want to write a hook that returns a value that will be updated based on some kind of asynchronous task that this hook is intended to initiate. The scheduling of that asynchronous side-effect must be guaranteed, in order to ensure any subsequent suspending effects that might consume the state managed by the prior asynchronous side-effect has the opportunity to initiate and complete and resolve itself so that the higher-level task can continue when ready. 
So any time we see this pattern:
```ts
const useSomething = () => {
  const x = useHookstate(*:any)
  // schedule something to update `x` at some point in the future
  return x
}
```

It is critical that it is implemented like this:
```ts
const useSomething = (params, deps) => {
  const x = useHookstate(*:any)
  const useImmediateEffect(() => { // not useLayoutEffect or useEffect!!!!
    // schedule something to update `x` at some point in the future
    // we *just* have to schedule it immediately, we don't have to 
    // actually execute the task immediately / synchronously
    // this can totally be a whole `startReactor()` chain of effects
    return () => {
      // here we cleanup our data bindings to anything we bound to in the effect!
      // if we made a reactor in the effect, we destroy it here, 
      // or schedule it's destruction here!
    } 
  }, [params, deps])
  return x
}
```

And in doing so, we can eliminate the possibility that any of our hooks will be directly responsible for never-ended suspense locks. 

This is important for a number of our hooks. Without this fix, the following apis have the potential to produce unintended unresolvable suspends under specific edge case conditions:

useQuery
useChildWithComponent
useAncestorWithComponent
useTreeQuery
useLoader
useBatchLoader
useScene

## References
closes #_insert number here_


HexaField:
I thought we wanted to support changing entity values for useComponent?

the-reality-engineer:
> I thought we wanted to support changing entity values for useComponent?

@HexaField  We do indeed. How do you perceive that this change precludes that support?

HexaField:
> > I thought we wanted to support changing entity values for useComponent?
> 
> @HexaField  We do indeed. How do you perceive that this change precludes that support?
> 
> 
 Hmm I suppose it makes sense to use useOptionalComponent in the case where a component doesn't exist, but conceptually passing in UndefinedEntity should be valid to useComponent. It should just suspend until there is a valid entity/component, rather than throwing.

Though we have also spoken about not suspending a context at all, so perhaps useOptionalComponent and useComponent should just be combined.

the-reality-engineer:
> Hmm I suppose it makes sense to use useOptionalComponent in the case where a component doesn't exist, but conceptually passing in UndefinedEntity should be valid to useComponent. It should just suspend until there is a valid entity/component, rather than throwing.
> 
> Though we have also spoken about not suspending a context at all, so perhaps useOptionalComponent and useComponent should just be combined.

I really appreciate being able to suspend. I do think it leads to the cleanest patterns at higher levels of abstraction in the long run. I believe that not being able to suspend w/ the feathers hooks leads to more complex reactive patterns than necessary, for example (though it definitely would be preferable in that particular use case to be able to suspend optionally). 

That said, you're right, it should be possible to suspend w/ useComponent when given UndefinedEntity, and then properly unsuspend when the context should be providing a different entity. That's a great point. Let me consider that before we merge this.

the-reality-engineer:
This should be ready now. The useQuery hook was breaking several rules of hooks (conditionally calling useOptionalComponent) and React started complaining about this after using `setImmediate` for whatever reason.  Rewrote it using valid react patterns, including reactivity to bitecs query modifiers (e.g. `Not(*)`), which we were pretty much ignoring before (ugh).

Adam-Zimmer (approved):

PR #10253, opened 2024-05-27
Optimize install projects find query

HexaField:
## References
closes #_insert number here_

### DIFF
--- a/scripts/install-projects.js
+++ b/scripts/install-projects.js
@@ -56,9 +56,9 @@ async function installAllProjects() {
     if (!fs.existsSync(localProjectDirectory)) fs.mkdirSync(localProjectDirectory, { recursive: true })
     logger.info('running installAllProjects')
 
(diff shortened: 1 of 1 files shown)


hanzlamateen (approved):

scripts/install-projects.js
```diff
@@ -56,9 +56,9 @@ async function installAllProjects() {
     if (!fs.existsSync(localProjectDirectory)) fs.mkdirSync(localProjectDirectory, { recursive: true })
     logger.info('running installAllProjects')
 
-    const projects = await app.service(projectPath).find({ paginate: false })
+    const projects = await app.service(projectPath).find({ paginate: false, assetsOnly: false })
```
hanzlamateen, line 59: @HexaField I wonder if this should be 

```
const projects = await app.service(projectPath).find({ paginate: false, query: { assetsOnly: false }})
```

HexaField: oh yeah probably...

PR #10255, opened 2024-05-28
IR: 1609: Handle scene renaming with existing scene name

MoizAdnan:


packages/editor/src/components/assets/ScenesPanel.tsx
```diff
@@ -138,7 +138,7 @@ export default function ScenesPanel() {
     setRenaming(false)
     const currentURL = loadedScene!.assetURL
     const newURL = currentURL.replace(currentURL.split('/').pop()!, newName + '.gltf')
-    const newData = await renameScene(id, newURL)
+    const newData = await renameScene(id, newURL, loadedScene!.projectName)
```
jonathan-casarrubias, line 141: By using the TypeScript's bang operator, we are ignoring an issue that the compiler is trying to warn, we are basically telling the compiler "shut up and trust me, I know", after reviewing the entire file for context, I see that loadedScene is being validated prior calling finishRenaming which btw doesn't handle the case where loadedScene is null or undefined which would cause an unexpected behavior for the end user having to click the button again, instead I recommend to get rid of the loadedScene validation prior calling finishRenaming (_functions must be self contained and not to depend on other functions' logic for its correct execution, depending on other functions logic it's a bad practice known as coupling, this function success depends on the parent function to validate the variable, if the function is called anywhere else without validating, the function will fail and that's what the compiler is trying to warn_), so getting rid of the bang operators and moving the finishRenaming logic within a conditional block handling the case where loadedScene might be null or undefined will guarantee to the compiler that there are no potential errors, therefore the compiler will stop complaining and then the bang operator won't be required.

Option A
```ts
  const finishRenaming = async (id: string) => {
    if (loadedScenes) {
       // perform renaming logic
    } else {
       throw new Error(`loadedScenes is ${typeof loadedScenes}`)
    }
  }
``` 

Option B (preferred)
```ts
  const finishRenaming = async (id: string) => {
    if (loadedScenes) {
       // perform renaming logic, compiler won't complain about nullable variable, because we demonstrated that
       // loadedScenes isn't null or undefined, therefore bang operator isn't required to work around what the compiler
       // is trying to avoid as potential issue
    } else {
       // set timeout to gracefully handle the error retrying the operation until loadedScene is available for X amount
       // of tries, after failing X amount of times, throw an error like in option A
    }
  }
``` 

The former it's a self contained function which result only depends on its own logic and handles its own errors, the latest is a more robust version in which result also only depends on its own logic but that also will gracefully handle errors before throwing the exception.

In conclusion, when we feel the need to use a bang operator to work around complains from the compiler it actually means that there's a design flaw in the implementation and the compiler is trying to warn us about it, like yes the variable is being validated prior calling the function, then why the compiler keeps complaining? because the function then is coupled to another function and that's the bad practice in this specific case that the compiler is trying to warn, even if we know that the variable is being validated before calling the function and will never be null or undefined, using the bang operator isn't the solution, the solution is to understand what the compiler is trying to warn and to fix it.

jonathan-casarrubias (changes requested):

packages/editor/src/components/assets/ScenesPanel.tsx
HexaField: While I agree with what you're saying @jonathan-casarrubias this is premature. We are doing this all over the place, and are focusing on architectural design rather than the smallest details, which is part of a longstanding effort to achieve what you are suggesting since the project began 4 years ago, trading off architectural design and immediate requirements over strict type safety.

jonathan-casarrubias (approved):

HexaField (approved):

PR #10256, opened 2024-05-28
Update Scene Assets search

CITIZENDOT:
## Summary

- Currently, Scene Asset search will work only if a category (Model, Image, Material etc..) is selected. This PR broadens the search to all assets, when no category is selected.
- Also adds `WAV` and `JSON` AssetTypes.

## References
closes #_insert number here_


packages/editor/src/components/assets/SceneAssetsPanel.tsx
```diff
@@ -251,20 +251,26 @@ const SceneAssetsPanel = () => {
...
-        return (
-          <>
-            {searchedStaticResources
-              .filter((resource) => {
+    if (searchedStaticResources.value) {
+      return (
+        <>
+          {searchedStaticResources
+            .filter((resource) => {
+              if (selectedCategory.value !== null) {
```
aditya-mitra: This case will be true for empty string as well.
We can do this:  `if (selectedCategory.value?.name && selectCategory.value?.object)`

packages/editor/src/components/assets/SceneAssetsPanel.tsx
```diff
@@ -251,20 +251,26 @@ const SceneAssetsPanel = () => {
...
-        )
+              } else if (searchText.value === '') {
+                return false
+              } else {
+                return true
+              }
+            })
+            .map((resource) => {
+              return <ResourceFile key={resource.value.id} resource={resource.get(NO_PROXY) as StaticResourceType} />
+            })}
```
aditya-mitra: we can make this shorter
```tsx
.map((resource) => (<ResourceFile key={resource.value.id} resource={resource.get(NO_PROXY) as StaticResourceType} />
            ))}

aditya-mitra (changes requested):
Looks good! Just a few changes and let's merge it.

packages/editor/src/components/assets/SceneAssetsPanel.tsx
CITIZENDOT: Done!

packages/editor/src/components/assets/SceneAssetsPanel.tsx
CITIZENDOT: Done!

aditya-mitra (approved):
Great!

PR #10257, opened 2024-05-28
Implement gizmo presets and optimize code

SYBIOTE:
This pull request includes the following changes:

- Added gizmo presets for better customization

- Removed unnecessary cloning operations

- Optimized code for better performance

These changes improve the functionality and efficiency of the gizmo feature in the software.


HexaField (changes requested):
we need meshes to be unique between instances of the transform gizmo, thus not module scoped, but materials and geometries can be

SYBIOTE (commented):
fixed it

packages/editor/src/constants/GizmoPresets.ts
```diff
@@ -0,0 +1,448 @@
...
+  return geometry
+}
+
+// Creates an Object3D with gizmos described in custom hierarchy definition.
+
+// Gizmo definitions - custom hierarchy definitions for setupGizmo() function
+
+const gizmoTranslate = {
+  X: [
+    [new Mesh(arrowGeometry, matRed), [0.5, 0, 0], [0, 0, -Math.PI / 2]],
```
HexaField, line 262: the meshes here still need to be created for each instances of the transform gizmo, otherwise their position/rotation/scale etc will be common between all instances (since we want to support multiple gizmos at a time). It's just the materials/geometries that we want to deduplicate so save on render calls & memory

HexaField (dismissed):

packages/editor/src/constants/GizmoPresets.ts
SYBIOTE: the setup gizmo function clones these

HexaField: Aha okay. It does look like the geometry is cloned. I understand it's to apply a rotation, and I suppose we probably would have to use a custom shader to optimize completely. Let's double check that the material is not cloned, and this should be fine.

SYBIOTE: The materials are not cloned when assigned to the meshes, the meshes are cloned later though

DanielBelmes (changes requested):
Sorry after some QAing it appears this PR breaks the transforms :(

SYBIOTE:
Hmm ,strange I'll take a look

SYBIOTE:
i fixed the issue with the transforms, it should be good to merge now

DanielBelmes (approved):

PR #10267, opened 2024-05-29
Fix infinite loading in `VariantComponent`

CITIZENDOT:
## Summary

Use `useOptionalComponent` instead of `useComponent` in the NodeEditor (for optional components)

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/components/properties/VariantNodeEditor.tsx
+++ b/packages/editor/src/components/properties/VariantNodeEditor.tsx
@@ -27,7 +27,7 @@ import DeblurIcon from '@mui/icons-material/Deblur'
 import React, { useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
 
-import { useComponent, useOptionalComponent } from '@etherealengine/ecs/src/ComponentFunctions'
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/properties/VariantNodeEditor.tsx
```diff
@@ -96,11 +96,12 @@ export const VariantNodeEditor: EditorComponentType = (props: { entity: Entity }
   const { t } = useTranslation()
   const entity = props.entity
   const variantComponent = useComponent(entity, VariantComponent)
-  const modelComponent = useComponent(entity, ModelComponent)
   const previewIndex = useHookstate(0)
 
   const setPreview = (index: number) => {
     previewIndex.set(index)
+    const modelComponent = useOptionalComponent(entity, ModelComponent)
```
HexaField: should be getOptionalMutableComponent instead of use if it is not in the body of a react functional component

HexaField (dismissed):

packages/editor/src/components/properties/VariantNodeEditor.tsx
CITIZENDOT: Done!

the-reality-engineer (approved):

PR #10268, opened 2024-05-29
Move studio registries to appropriate directory

HexaField:
## Summary

These registry states should be centrally located in the editor file structure, and not included in the files of the UI that consumes them.

## References
closes #_insert number here_


aditya-mitra (approved):
Thanks for refactoring this!!!

PR #10272, opened 2024-05-29
Add File Browser Search

CITIZENDOT:
Filter entries based on $like query in file-browser service

## References
closes #_insert number here_


hanzlamateen (changes requested):
Can you add some tests in `packages/server-core/src/media/file-browser/file-browser.test.ts` to test this?

hanzlamateen (approved):

PR #10274, opened 2024-05-29
adding helper function hook for in-frustum

SamMazerIR:
## Summary
adding helper function hook for whether an entity context is in frustum of the viewer

closes #_insert number here_
[https://tsu.atlassian.net/browse/IR-1652](https://tsu.atlassian.net/browse/IR-1652)

### DIFF
--- a/packages/spatial/src/transform/components/TransformComponent.ts
+++ b/packages/spatial/src/transform/components/TransformComponent.ts
@@ -24,12 +24,14 @@ Ethereal Engine. All Rights Reserved.
 */
 
 import { Types } from 'bitecs'
-import { Euler, Matrix4, Quaternion, Vector3 } from 'three'
+import { Euler, Frustum, Matrix4, Quaternion, Vector3 } from 'three'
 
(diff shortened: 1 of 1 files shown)


packages/spatial/src/transform/components/TransformComponent.ts
```diff
@@ -227,6 +233,15 @@ export const TransformComponent = defineComponent({
     transform.matrixWorld.compose(vec3, quat, scale)
   },
 
+  useInViewingFrustum: () => {
```
HexaField: can we pass in the entity and camera instead? using useEntityContext here is indirectional, and this should be extensible to any camera

HexaField (changes requested):

HexaField (changes requested):
also this hook will not update reactively. we'd need to probably either need to store a hookstate boolean that is updated in a useExecute system or something

packages/spatial/src/transform/components/TransformComponent.ts
SamMazerIR: changed this to just be a function call with params as suggested here. I don't think it being a hook is necessary in this case

PR #10275, opened 2024-05-29
Updated prune_ecr_images to correctly handle private ECR repos.

barankyle:
## Summary
Converted that script to .ts

Made findBuilderTags filter on tagged status.

## References
closes #_insert number here_

### DIFF
--- a/scripts/build_and_publish_package.sh
+++ b/scripts/build_and_publish_package.sh
@@ -128,9 +128,9 @@ fi
 
 if [ $PRIVATE_REPO == "true" ]
 then
-  node ./scripts/prune_ecr_images.js --repoName $DESTINATION_REPO_NAME_STEM-$PACKAGE --region $REGION --service $PACKAGE --releaseName $STAGE
+  npx ts-node ./scripts/prune_ecr_images.ts --repoName $DESTINATION_REPO_NAME_STEM-$PACKAGE --region $REGION --service $PACKAGE --releaseName $STAGE
 else
-  node ./scripts/prune_ecr_images.js --repoName $DESTINATION_REPO_NAME_STEM-$PACKAGE --region us-east-1 --service $PACKAGE --releaseName $STAGE --public
+  npx ts-node ./scripts/prune_ecr_images.ts --repoName $DESTINATION_REPO_NAME_STEM-$PACKAGE --region us-east-1 --service $PACKAGE --releaseName $STAGE --public
 fi
 
 BUILD_END_TIME=`date +"%d-%m-%yT%H-%M-%S"`
--- a/scripts/build_docker_builder.sh
+++ b/scripts/build_docker_builder.sh
@@ -50,8 +50,8 @@ if [ $REPO_PROVIDER === "aws" ]
 then
   if [ $PRIVATE_REPO == "true" ]
   then
-    node ./scripts/prune_ecr_images.js --repoName $REPO_NAME-builder --region $REGION --service builder --releaseName $STAGE
+    npx ts-node ./scripts/prune_ecr_images.ts --repoName $REPO_NAME-builder --region $REGION --service builder --releaseName $STAGE
   else
-    node ./scripts/prune_ecr_images.js --repoName $REPO_NAME-builder --region us-east-1 --service builder --releaseName $STAGE --public
+    npx ts-node ./scripts/prune_ecr_images.ts --repoName $REPO_NAME-builder --region us-east-1 --service builder --releaseName $STAGE --public
   fi
 fi
--- a/scripts/prune_ecr_images.ts
+++ b/scripts/prune_ecr_images.ts
@@ -1,4 +1,3 @@
-
 /*
 CPAL-1.0 License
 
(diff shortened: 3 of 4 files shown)


scripts/prune_ecr_images.ts
```diff
@@ -41,88 +47,127 @@ const options = cli.parse({
...
         )
       }
       currentImages = [...new Set(currentImages)]
     }
-    const ecr = (
+
+    console.log('EKS config', config.aws.eks)
+    const awsCredentials = `[default]\naws_access_key_id=${config.aws.eks.accessKeyId}\naws_secret_access_key=${config.aws.eks.secretAccessKey}\n[role]\nrole_arn = ${config.aws.eks.roleArn}\nsource_profile = default`
+
+    console.log('awsCredentials', awsCredentials)
```
lucas3900: If we don't have log level control, we shouldn't log the AWS access credentials to the console. And maybe `config.aws.eks` depending on what is stored there.

scripts/build_docker_builder.sh
```diff
@@ -50,8 +50,8 @@ if [ $REPO_PROVIDER === "aws" ]
 then
   if [ $PRIVATE_REPO == "true" ]
   then
-    node ./scripts/prune_ecr_images.js --repoName $REPO_NAME-builder --region $REGION --service builder --releaseName $STAGE
+    npx ts-node ./scripts/prune_ecr_images.ts --repoName $REPO_NAME-builder --region $REGION --service builder --releaseName $STAGE
   else
-    node ./scripts/prune_ecr_images.js --repoName $REPO_NAME-builder --region us-east-1 --service builder --releaseName $STAGE --public
+    npx ts-node ./scripts/prune_ecr_images.ts --repoName $REPO_NAME-builder --region us-east-1 --service builder --releaseName $STAGE --public
```
lucas3900, line 55: Public repos can be in any region. Should we use `$REGION` here as well?

scripts/build_and_publish_package.sh
```diff
@@ -128,9 +128,9 @@ fi
 
 if [ $PRIVATE_REPO == "true" ]
 then
-  node ./scripts/prune_ecr_images.js --repoName $DESTINATION_REPO_NAME_STEM-$PACKAGE --region $REGION --service $PACKAGE --releaseName $STAGE
+  npx ts-node ./scripts/prune_ecr_images.ts --repoName $DESTINATION_REPO_NAME_STEM-$PACKAGE --region $REGION --service $PACKAGE --releaseName $STAGE
 else
-  node ./scripts/prune_ecr_images.js --repoName $DESTINATION_REPO_NAME_STEM-$PACKAGE --region us-east-1 --service $PACKAGE --releaseName $STAGE --public
+  npx ts-node ./scripts/prune_ecr_images.ts --repoName $DESTINATION_REPO_NAME_STEM-$PACKAGE --region us-east-1 --service $PACKAGE --releaseName $STAGE --public
```
lucas3900, line 133: Public repos can be in any region. Should we use `$REGION` here as well?

scripts/prune_ecr_images.ts
```diff
@@ -41,88 +47,127 @@ const options = cli.parse({
...
-        ? new ECRPUBLICClient({ region: 'us-east-1' })
-        : new ECRClient({ region: options.region || 'us-east-1' })
-    )
-    const images = await getAllImages(ecr, options.repoName || 'etherealengine', undefined, [])
+        ? new ECRPUBLICClient({
+            credentials: fromIni({
+              profile: config.aws.eks.roleArn ? 'role' : 'default',
+              filepath: credentialsPath
+            }),
+            region: 'us-east-1'
```
lucas3900, line 154: here for public region as well

lucas3900:
Wow the way the TS SDK makes your jump through hoops just to sort and delete on an ECR is insane.

scripts/prune_ecr_images.ts
barankyle: These were debugging statements, I'll remove them when it's 100% ready.

scripts/build_docker_builder.sh
barankyle: I swear that every time I've tried to access a public repo that's located in a non-us-east-1 region using that region, it hasn't worked, but that using us-east-1 as the region has worked. I will re-check this to confirm.

barankyle: Yeah, the SDK for ECR Public works a little weird when combined with how public repos are addressed. I think public repos are region-less, or are always in us-east-1; if I switch regions in the web console when looking at public repos, I always see every one.

If I set up ECRPUBLICClient with a region other than us-east-1, then try to call a DescribeImagesCommand, I get e.g. 
```
Uncaught Error: getaddrinfo ENOTFOUND api.ecr-public.us-west-1.amazonaws.com
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:107:26)
    at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:128:17) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: 'api.ecr-public.us-west-1.amazonaws.com',
  '$metadata': { attempts: 1, totalRetryDelay: 0 }
}
```
I don't know why it lets you change the region if anything other than us-east-1 is invalid, but that's why I hard-coded it to us-east-1.

lucas3900: that's very weird. always fun exploring the hidden functionality of AWS

lucas3900 (approved):
Approved. Just leaving a note here to make sure we delete those logging statements. Might be good to set up a flow where we test those sort of AWS/K8s specific things on QAT, prior to merging into dev?

barankyle:
> Approved. Just leaving a note here to make sure we delete those logging statements. Might be good to set up a flow where we test those sort of AWS/K8s specific things on QAT, prior to merging into dev?

I'd already deleted them. It is very useful to test them on QAT, which I did for this, making it explicit would be a good thing.

PR #10282, opened 2024-05-30
[IR-2164] Unit Test - Physics

heysokam:
## Subtasks Checklist
- [x] All `Physics.ts` Rapier->ECS APIs
- [x] ~~All `Physics.ts` External APIs~~
- [x] `RigidbodyComponent`
- [x] `ColliderComponent`
- [x] `CollisionComponent`
- [x] `TriggerComponent`
- [x] `TriggerSystem`
- [x] `PhysicsSystem`


heysokam:
@HexaField @DanielBelmes
1. Are these tests going in the right direction?
2. I have tests for like 35-40% of a single file, and its already at 700sloc. Should I subdivide this task into multiple PRs? It feels massive for a single one :thinking:

packages/spatial/src/physics/classes/Physics.test.ts
```diff
@@ -361,3 +362,714 @@ describe('Physics', () => {
...
+          body.bodyType(),
+          RigidBodyType.KinematicPositionBased,
+          "The RigidBody's type was not changed to KinematicPositionBased"
+        )
+      })
+    })
+
+    describe('setRigidbodyPose', () => {
+      const position = new Vector3(1, 2, 3)
+      const rotation = new Quaternion(4, 5, 6, 1)
```
HexaField: we should only be using valid values, we should normalize this value

packages/spatial/src/physics/classes/Physics.test.ts
```diff
@@ -361,3 +362,714 @@ describe('Physics', () => {
...
+        const body = Physics._Rigidbodies.get(testEntity)!
+        Physics.enabledCcd(testEntity, true)
+        assert.equal(body.isCcdEnabled(), true)
+        Physics.enabledCcd(testEntity, false)
+        assert.equal(body.isCcdEnabled(), false)
+      })
+    })
+
+    /**
+  // @todo How to check for the impulse of an entity?
```
HexaField: impulse will have a resultant velocity the next frame, we can calculate what the velocity is, or just assert it's beforehand and assert it's not zero or close to an expected value afterwards

packages/spatial/src/physics/classes/Physics.test.ts
```diff
@@ -361,3 +362,714 @@ describe('Physics', () => {
...
+      const testImpulse = new Vector3(1,2,3)
+      const body = Physics._Rigidbodies.get(testEntity)!
+      Physics.applyImpulse(testEntity, testImpulse)
+    })
+  })
+  */
+
+    /**
+  // @todo How to check rotations?
+  describe("lockRotations", () => {
```
HexaField: apply an impulse and assert it has not rotated along the locked axis

packages/spatial/src/physics/classes/Physics.test.ts
```diff
@@ -361,3 +362,714 @@ describe('Physics', () => {
...
+  describe("setEnabledRotations", () => {})
+  describe("updatePreviousRigidbodyPose", () => {})
+  describe("updateRigidbodyPose", () => {})
+  describe("setKinematicRigidbodyPose", () => {})
+  */
+  }) // << Rigidbodies
+
+  describe('Colliders', () => {
+    /**
+    // @todo How to check for `CollisionGroups` behavior?
```
HexaField: you can use the collisionGroups() and isSensor() functions on the collider object

HexaField (changes requested):

packages/spatial/src/physics/classes/Physics.test.ts
heysokam:
```ts
it("should apply the impulse to the RigidBody of the entity", () => {
  const testImpulse = new Vector3(1,2,3)
  const body = Physics._Rigidbodies.get(testEntity)
  assert.ok(body)
  
  console.log(JSON.stringify(body.linvel()))
  console.log(JSON.stringify(body.angvel()))
  
  Physics.applyImpulse(testEntity, testImpulse)
  
  console.log(JSON.stringify(body.linvel()))
  console.log(JSON.stringify(body.angvel()))
})
```
```ts
// Output
{"x":0,"y":0,"z":0}
{"x":0,"y":0,"z":0}
{"x":0,"y":0,"z":0}
{"x":0,"y":0,"z":0}
```
Am I missing something to "sync" the physics so that they are applied right away?

packages/spatial/src/physics/classes/Physics.test.ts
heysokam: After applying the reference rotation values on creation, the resulting Quat does not remain the same compared to the starting reference. It somehow changes to float values with many decimals that have a lot of variance from run to run due to floating point imprecision.

So, this "how to check rotations" note is for asking about asserting Quaternion/Rotation results, as in how to assert the contained float values themselves in a reliable way :thinking:

packages/spatial/src/physics/classes/Physics.test.ts
HexaField: you need to run the physics simulation, looks at PhysicsSystem.ts

packages/spatial/src/physics/classes/Physics.test.ts
heysokam: ty. Solved in 9e63efe3bee15f048d18d911eec260ef5f12eccb

packages/spatial/src/physics/classes/Physics.test.ts
heysokam: The passed quaternion was being normalized internally by Rapier.
So the issue was that I couldn't check the Expected and the Result, because they were different no matter the expected value.
Calling `.normalize()` on the input/expected quaternion, no matter their values, solved the problem :writing_hand:

packages/spatial/src/physics/classes/Physics.test.ts
heysokam: Fixed somewhere in the commit list of the PR
_Forgot to track the commit where I applied it, and now I cant find the sha_

packages/spatial/src/physics/classes/Physics.test.ts
heysokam: Fixed somewhere in the commit list.   _(forgot to track the commit's sha)_

The mistake was that Quaterions were being normalized internally in Rapier, but I wasn't sending a normalized value.
It was solved by normalizing the input value and the checking the resulting data against that input.

HexaField (approved):

PR #10283, opened 2024-05-30
Fix Image Compression Settings

CITIZENDOT:
## References
closes #_insert number here_


packages/editor/src/components/assets/ImportSettingsPanel.tsx
```diff
@@ -83,7 +86,7 @@ const UASTCFlagOptions = [
   { label: 'Disable Flip and Individual', value: 256 }
 ]
 
-const ImageCompressionBox = ({ compressProperties }) => {
+const ImageCompressionBox = ({ compressProperties }: { compressProperties: State<KTX2EncodeArguments> }) => {
```
HexaField: rather than passing compressProperties down as a state dependency, can we just call `const compressProperties = useMutableState(ImportSettingsState).imageSettings` again in this component?

HexaField (changes requested):

packages/editor/src/components/assets/ImportSettingsPanel.tsx
CITIZENDOT: Yea, Good Idea.

CITIZENDOT: Actually, we have to pass the state.

`handleSaveChanges()` is called in the parent component (`ImportSettingsPanel`). It needs to know the changes done to `imageSettings`. (Hence passing the state).

But we can not move the `handleSaveChanges()` into `ImageCompressionBox` because `handleSaveChanges()` deals with some other things too, so it's not just associated with `ImageCompressionBox`.

HexaField: But the `ImportSettingsState.imageSettings` is global. Any changes made in `ImageCompressionBox` will cause a rerender of `ImportSettingsPanel`, and if `handleSaveChanges()` is called, it will have those new values.

HexaField (changes requested):

packages/editor/src/components/assets/ImportSettingsPanel.tsx
CITIZENDOT: Done!

HexaField (approved):

PR #10286, opened 2024-05-30
Change ECS exporter to get gltf data from authoring layer

dinomut1:
## References
closes #_insert number here_


dinomut1:
this should actually be optional. I think we want to decouple I/O from the authoring layer anyways as we refactor layers

PR #10287, opened 2024-05-30
IR-2365 Fix erroneous material updates and prototype switching errors

AidanCaruso:
## References
closes #_insert number here_


JT00y (approved):
thanks it fixed the material problem

PR #10295, opened 2024-05-31
optimize spline helper

HexaField:
## References
closes #_insert number here_


packages/engine/src/scene/components/debug/SplineHelperComponent.ts
```diff
@@ -75,39 +64,43 @@ export const SplineHelperComponent = defineComponent({
...
-    const [redMat] = useResource(redMeshMaterial, entity)
-    const [redSphere] = useDisposable(
-      Mesh,
-      entity,
-      sphereGeometry.value as SphereGeometry,
-      redMat.value as MeshBasicMaterial
-    )
+
+    const [lineGeometry] = useResource(createLineGeom, entity)
+    /** @todo these are probably unnecessary and were just used for debugging the implementation */
```
MichaelEstes, line 76: Everything looks good without this, do we want to keep it in as a comment?

HexaField: yeah we can keep this as a todo to optimize or just come back to in the future

MichaelEstes (approved):

PR #10296, opened 2024-05-31
IR-2225 Asset Loading Tech Debt

MichaelEstes:
## Summary
- Moves asset related enums to common package
- Adds custom TextureLoader with scaling logic
- All required post load asset transformations have been moved into the loaded callbacks in ResourceState  

## Breaking Changes
- AssetType.ts and AssetClass.ts files have been removed, replaced with AssetType.ts in common package
- AssetLoader.load and AssetLoader.loadAsync have been removed, replaced with AssetLoader.loadAsset

## References
closes #_insert number here_


HexaField (approved):
there is also `CommonKnownContentTypes.ts` and `fileTypes.tsx` that may have commonalities that could be combined, but we can do that another time. Looks great!

PR #10299, opened 2024-05-31
IR-2153:  Disable/Enable Publish button based on permissions & Scopes

jose-galvan:
## Summary

This disables/enables the publish button in the tool bar in the new studio UI. 
It is only enable if a user in the owner or an editor, or if the user has location:write scope. 

## References
closes #_insert number here_

## QA Steps

### DIFF
--- a/packages/editor/src/components/Editor2Container.tsx
+++ b/packages/editor/src/components/Editor2Container.tsx
@@ -27,7 +27,7 @@ import { PopoverState } from '@etherealengine/client-core/src/common/services/Po
 import { useRemoveEngineCanvas } from '@etherealengine/client-core/src/hooks/useRemoveEngineCanvas'
 import { assetPath } from '@etherealengine/common/src/schema.type.module'
 import { EntityUUID } from '@etherealengine/ecs'
-import { getMutableState, useHookstate, useMutableState } from '@etherealengine/hyperflux'
+import { NO_PROXY, getMutableState, useHookstate, useMutableState } from '@etherealengine/hyperflux'
 import { useFind } from '@etherealengine/spatial/src/common/functions/FeathersHooks'
 import { AssetsPanelTab } from '@etherealengine/ui/src/components/editor/panels/Assets'
 import { FilesPanelTab } from '@etherealengine/ui/src/components/editor/panels/Files'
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/Editor2Container.tsx
```diff
@@ -123,6 +125,11 @@ const EditorContainer = () => {
 
   useHotkeys(`${cmdOrCtrlString}+s`, () => PopoverState.showPopupover(<SaveSceneDialog />))
 
+  const user = useHookstate(getMutableState(AuthState).user)
```
aditya-mitra: this has been changed to `useMutableState`

packages/editor/src/components/Editor2Container.tsx
```diff
@@ -123,6 +125,11 @@ const EditorContainer = () => {
...
+  const hasLocationWriteScope = user.scopes.get(NO_PROXY)?.find((item) => item?.type === 'location:write')
```
aditya-mitra: Using [`useUserHasAccessHook`](https://github.com/EtherealEngine/etherealengine/blob/9cdea6a91825c4e1b075d7bfa4ef7480724801f3/packages/client-core/src/user/userHasAccess.ts#L30-L34) we can combine line 128 and 129 into a shorter single line

packages/editor/src/components/toolbar/Toolbar2.tsx
```diff
@@ -105,7 +105,11 @@ const generateToolbarMenu = () => {
 
 const toolbarMenu = generateToolbarMenu()
 
-export default function Toolbar() {
+interface Props {
+  publishSceneDisabled: boolean
+}
+
```
aditya-mitra: If this prop is only used in the toolbar, we can directly have the [added lines](https://github.com/EtherealEngine/etherealengine/pull/10299/files#diff-1ba1511d694d6d92370f194f1133e5a389146fff09830725402671afd2988e4cR128-R131) from `Editor2Container.tsx` to `Toolbar2.tsx` file

aditya-mitra (changes requested):

packages/editor/src/components/toolbar/Toolbar2.tsx
jose-galvan: @aditya-mitra  Moved logic to check for publish access to the toolbar component

aditya-mitra (approved):
Great work! 🚀

PR #10302, opened 2024-06-01
Add uniform scaling UI and fix collider resizing

HexaField:
## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/components/editor/input/Vector3/index.tsx
+++ b/packages/ui/src/components/editor/input/Vector3/index.tsx
@@ -28,6 +28,7 @@ import { Vector3_Zero } from '@etherealengine/spatial/src/common/constants/MathC
 import React from 'react'
 import { twMerge } from 'tailwind-merge'
 import { Vector3 } from 'three'
+import Icon from '../../../../primitives/mui/Icon'
 import Scrubber from '../../layout/Scrubber'
 import NumericInput from '../Numeric'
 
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/editor/input/Vector3/index.tsx
```diff
@@ -93,7 +96,13 @@ export const Vector3Input = ({
   onRelease,
   ...rest
 }: Vector3InputProp) => {
-  const uniformEnabled = useHookstate(uniformScaling)
+  const id = uniqueId++
+  const uniformEnabled = useHookstate(false)
```
aditya-mitra: We are using ids for accessibility purposes which should remain constant. Due to re-rendering, ids will change. React provides a [useId](https://react.dev/reference/react/useId) hook for generating unique ids.

image

In this case, we can wrap the input inside label somewhat like this:
```tsx
            <label title="Uniform Scale" htmlFor={checkboxId}>
              <input
              className="hidden"
              id={checkboxId}
              type="checkbox"
              checked={uniformEnabled.value}
              onChange={onToggleUniform}
              />
              <Icon type={uniformEnabled.value ? 'Link' : 'LinkOff'} />
            </label>
```

I will replace this with a tailwind button with the icon set conditionally

aditya-mitra (changes requested):
Need a few changes with the id usage.
We should discourage id usage unless required with exhausted alternatives.

packages/ui/src/components/editor/input/Vector3/index.tsx
HexaField: Cool thanks, I just copied it from the old implementation that was lost

aditya-mitra (approved):

Issue #9749, opened 2024-02-13
[Epic] Project Versioning


HexaField:
- [x] Add script to install projects from manifest
  - [x] Checks out to specific branch
  - [x] npm run install-manifest -- --manifestURL="https://raw.githubusercontent.com/EtherealEngine/project-manifest/main/manifest.json" --branch="dev"
- [x] Add git checkout dev script
- [x] Default all scripts to main branch
- [x] Add main & dev branches for every project
- [x] Default all projects to dev branch
- [x] Lock down perms on all projects
- [x] Disable projects that having incorrect engine version number
- [ ] Rename `etherealEngine` property in project package.json to `engineVersion`
- [ ] Add CLI & GHA scripts to help automate updates & releases
  - [ ] Change updating dev version to after the release is made
  - [ ] Auto-merge engine release update PR
  - [ ] Upon tag being created in etherealengine repo, run workflow in all projects
    - [ ] Launch project update workflow for release event for engine https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release 
  - [ ] Add a PR with the changes to the version number
- [ ] How can we automatically create main & dev branches for every project?
  - [x] npm run create-project – –name=”ee-new-project”
  - [ ] Can we automate branch protection?
- [ ] Add CI/CD for all projects, and include all open source projects in engine CI/CD with branch name matching
  - [ ] Once engine PR is closed or merged, close or merge all related PRs
- [ ] Test tag flow & create tag releases
- [ ] Force projects to have org namespace in project name (yet still allow esm imports…)
  - [ ] Symlinking with org namespaces seems to work
  - [ ] We need to change the folder structure to be /packages/projects/projects/<@namespace>/<project-name>

PR #10304, opened 2024-06-03
Static Resource, Asset, CMS Backend Refactor

HexaField:
## Summary

- Remove sid field from static-resource table
- move getCachedURL and getCacheDomain to storage provider interface
- Add new fields to static-resource table
- Change structure of resources.json
- Remove `asset` service, in favour of querying for `static-resource` service with `type: "scene"`
- File browser service API now makes project path implicit, and supports projects with slash in the name
- File browser disallows non-project paths, and non /asset/ or /public/ paths from being accessed
- Static resources service overhauled to 
- Rewrites file browser tests to be a bit cleaner and less procedural
- refactored API for projects seeding avatars

## References
closes #_insert number here_


packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -181,7 +165,9 @@ export class FileBrowserService
     const storageProvider = getStorageProvider(params?.query?.storageProviderName)
     if (directory[0] === '/') directory = directory.slice(1)
 
-    checkDirectoryInsideNesting(directory, params?.nestingDirectory)
+    if (!directory.startsWith('projects/')) throw new Error('Not allowed to access this directory')
```
aditya-mitra: this can also written as `ensureProjectDirectory(directory)`

HexaField:
Requested provisional code review to address any architectural decisions. Still need to test all admin, console & studio functionality.

Some of the file browser functions have taken on a bit of bloat necessary for MT, and will be simplified in the future once the last point in https://github.com/EtherealEngine/etherealengine/issues/9749 is solved (i believe I might have a short term solution worked out in my mind)

packages/client-core/src/components/World/LoadLocationScene.tsx
```diff
@@ -80,18 +80,18 @@ export const useLoadLocation = (props: { locationName: string }) => {
...
+    const sceneURL = scene.url
     return GLTFAssetState.loadScene(sceneURL, scene.id)
   }, [locationState.currentLocation.location.sceneId, scene])
 }
 
 export const useLoadScene = (props: { projectName: string; sceneName: string }) => {
-  const sceneURL = `projects/${props.projectName}/${props.sceneName}`
-  const assetID = useFind(assetPath, { query: { assetURL: sceneURL } })
+  const sceneKey = `projects/${props.projectName}/${props.sceneName}`
+  const assetID = useFind(staticResourcePath, { query: { key: sceneKey } })
```
aditya-mitra: adding `type: 'scene` here will speed up the query since static-resources will be bigger table

packages/projects/createLocations.ts
```diff
@@ -23,62 +23,5 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
-        query: {
-          slugifiedName: cleanedLocationName
-        }
-      })) as Paginated<LocationType>
-      if (existingLocation.total === 0) await app.service(locationPath).create(location)
-    })
-  )
-}
+import { createLocations } from '@etherealengine/server-core/src/social/location/location-helper'
+export { createLocations }
```
aditya-mitra: how about re-export?
`export { createLocations } from '@etherealengine/server-core/src/social/location/location-helper'`

packages/server-core/src/assets/ktx2-encode/ktx2-encode.service.ts
```diff
@@ -41,44 +41,49 @@ declare module '@etherealengine/common/declarations' {
...
+      const outPath = inPath.replace(/\.[^\.]+$/, '.ktx2')
+      const outURIDir = path.dirname(inURI)
+      const projectRelativeDirectoryPath = outURIDir.split('/').slice(2).join('/')
+      const fileName = /[^\\/]*$/.exec(outPath)![0]
+      const args = `-ktx2${data.mode === 'UASTC' ? ' -uastc' : ''}${data.mipmaps ? ' -mipmap' : ''}${
+        data.flipY ? ' -y_flip' : ''
+      }${data.srgb ? ' -linear' : ''} -q ${data.quality} ${inPath}`
+      console.log(args)
+      console.log(execFileSync(BASIS_U, args.split(/\s+/)))
+      console.log(execFileSync('mv', [fileName, outPath]))
```
aditya-mitra, line 65: do we need these console.logs? if so, we can have a distinguisher to read the logs when printed

packages/server-core/src/media/storageprovider/storageprovider.interface.ts
```diff
@@ -202,12 +198,14 @@ export interface StorageProviderInterface {
...
   doesExist(fileName: string, directoryPath: string): Promise<boolean>
 
+  getCacheDomain(internal?: boolean): string
+
   /**
-   * Get the object from edge cache, otherwise returns getObject.
+   * Get the object's URL on the edge cache.
    * @param key Key of object.
-   * @returns {StorageObjectInterface}
+   * @returns {string}
```
aditya-mitra, line 206: this typedoc comment can be removed as return type is inferred by typescript

packages/server-core/src/seeder.ts
```diff
@@ -55,7 +57,7 @@ export async function seeder(app: Application, forceRefresh: boolean, prepareDb:
       if (fs.existsSync(uploadPath)) fs.rmSync(uploadPath, { recursive: true })
     }
     copyDefaultProject()
-    await app.service(projectPath)._seedProject('default-project')
+    if (config.kubernetes.enabled || config.testEnabled) await app.service(projectPath)._seedProject('default-project')
```
aditya-mitra, line 60: we should also reseed on dev-reinit

aditya-mitra (changes requested):
Really liked the parts of shortening!

packages/client-core/src/admin/components/resources/ResourceTable.tsx
```diff
@@ -79,22 +77,6 @@ export default function ResourceTable({ search }: { search: string }) {
...
-                  text={`${t('admin:components.resources.confirmResourceDelete')} '${el.key}'?`}
-                  onSubmit={async () => {
-                    await Engine.instance.api.service(staticResourcePath).remove(el.id)
-                  }}
-                />
-              )
-            }}
-          >
-            <HiTrash className="place-self-center text-theme-iconRed" />
-          </Button>
```
aditya-mitra: why are we removing the ability to delete in admin?

packages/server-core/src/seeder.ts
HexaField: this will happen in `_fetchDevLocalProjects`

packages/server-core/src/assets/ktx2-encode/ktx2-encode.service.ts
HexaField: these were already there and the service isn't used anywhere, its a placeholder, so i'm not worried about it

packages/client-core/src/admin/components/resources/ResourceTable.tsx
HexaField: oh whoops, i need to add this back.

packages/server-core/src/media/storageprovider/storageprovider.interface.ts
HexaField: the rest of the functions here are the same, we can do this another time

aditya-mitra (approved):

packages/server-core/src/media/static-resource/migrations/20240603013245_static-resource-cms.ts
```diff
@@ -0,0 +1,215 @@
...
+import { createStaticResourceHash } from '../../upload-asset/upload-asset.service'
+
+const assetPath = 'asset'
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+export async function up(knex: Knex): Promise<void> {
+  const trx = await knex.transaction()
```
barankyle: Per what Daniel researched last week in #10331, we shouldn't be making knex transactions in our migrations, because the migration itself is a transaction. You can just remove the `const trx = await knex.transaction()` and `await trx.commit()` and replace every `trx` with `knex`.

packages/server-core/src/mysql.ts
```diff
@@ -121,14 +121,28 @@ export default (app: Application): void => {
...
           logger.info('Knex migration rollback started')
-          await knexClient.migrate.rollback(config.migrations, true)
+
+          const allTables = (
+            await db.raw(
+              `select table_name from information_schema.tables where table_schema = '${appConfig.db.database}'`
+            )
+          )[0].map((table) => table.table_name)
+
+          const trx = await knexClient.transaction()
```
barankyle, line 131: Not sure if we should be getting rid of the transaction here since it's outside a test. But maybe see if this can also be done with straight `knex.raw` commands.

packages/editor/src/functions/sceneFunctions.tsx
```diff
@@ -53,17 +52,29 @@ const logger = multiLogger.child({ component: 'editor:sceneFunctions' })
...
+  resource: StaticResourceType,
+  newKey: string,
+  projectName: string,
+  params?: Params
+) => {
+  const oldPath = resource.key
+  const newPath = newKey
+  const oldName = resource.key.split('/').pop()!
+  const newName = newKey.split('/').pop()!
+  console.log({ resource, newKey, projectName, params, oldPath, newPath, oldName, newName })
```
barankyle: Can be removed

packages/server-core/src/media/static-resource/static-resource.hooks.ts
```diff
@@ -69,6 +76,45 @@ const ensureResource = async (context: HookContext<StaticResourceService>) => {
...
+  const results =
+    'data' in context.result ? context.result.data : Array.isArray(context.result) ? context.result : [context.result]
+
+  const projectNames = results.reduce((acc: string[], result) => {
+    if (result.project && result.key.startsWith('projects/') && !acc.includes(result.project)) acc.push(result.project)
+    return acc
+  }, [])
+
+  for (const projectName of projectNames) {
+    await updateProjectResourcesJson(context.app, projectName)
```
dinomut1: We shouldn't fully regenerate the resources.json with every CRUD operation. This means every edit to the static resources involves querying the entirety of a project's static resources and writing it into a json file. This will get out of hand quickly with larger projects + mass uploads. We need specific patch and delete functions which perform singular operations on the resource.json

packages/server-core/src/media/static-resource/static-resource.hooks.ts
```diff
@@ -164,10 +234,10 @@ export default {
     all: [],
     find: [],
     get: [],
-    create: [],
-    update: [],
-    patch: [],
-    remove: []
+    create: [updateResourcesJson],
+    update: [updateResourcesJson],
```
dinomut1, line 256: need individual functions for patch + remove on resources.json

dinomut1 (changes requested):

barankyle (changes requested):
When I applied the migration via `prepare-database`, it seemed to work, but when I created a new scene, it didn't immediately appear in the scene list in the studio; I had to refresh. And when I made further new scenes, I did not see anything even after refreshing. The backend might not be doing that thing where it will append e.g. (1) to the name if it detects it's a duplicate name. If I manually rename New-Scene.gltf, the updated name doesn't appear in the studio unless I refresh. And if I then make a new scene, it gets made as New-Scene.gltf, which is expected, so it's probably the lack of de-duplication that's the problem.

When I ran dev-reinit, all static-resources, even scene files, had type 'file', so no scenes showed up in the studio. And are some of those files supposed to be of type 'asset'? Making a new scene worked, and New-Scene was of type 'scene'.

packages/server-core/src/media/static-resource/static-resource.hooks.ts
the-reality-engineer: ^ seems like a reasonable optimization

HexaField: I actually disagree here. Batch uploads can/should be supported at the service level when needed, and that kind of optimisation I don't feel is appropriate at this stage. The difference between a partial write and a complete write is negligible, as the query/read/write requests themselves are the heavy part, not the size of the query.

packages/server-core/src/mysql.ts
DanielBelmes: Depends on if this is meant to be atomic or not. Like if a part of the code in the transaction fails do we want all of it to do nothing or do we want partial updates. (I did google what transactions are for for this comment. But I think I am right :) )

dinomut1 (approved):

barankyle:
I'm still seeing these issues:

> When I applied the migration via `prepare-database`, it seemed to work, but when I created a new scene, it didn't immediately appear in the scene list in the studio; I had to refresh. And when I made further new scenes, I did not see anything even after refreshing. The backend might not be doing that thing where it will append e.g. (1) to the name if it detects it's a duplicate name. If I manually rename New-Scene.gltf, the updated name doesn't appear in the studio unless I refresh. And if I then make a new scene, it gets made as New-Scene.gltf, which is expected, so it's probably the lack of de-duplication that's the problem.
> 
> When I ran dev-reinit, all static-resources, even scene files, had type 'file', so no scenes showed up in the studio. And are some of those files supposed to be of type 'asset'? Making a new scene worked, and New-Scene was of type 'scene'.

I also noticed that after dev-reinit, the thumbnail image for scenes in the 'scenes' tab is pointing to the GLTF of the scene, e.g. `projects/default-project/public/scenes/New-Scene.gltf`. This of course does not render a thumbnail.

HexaField:
> I'm still seeing these issues:
> 
> > When I applied the migration via `prepare-database`, it seemed to work, but when I created a new scene, it didn't immediately appear in the scene list in the studio; I had to refresh. And when I made further new scenes, I did not see anything even after refreshing. The backend might not be doing that thing where it will append e.g. (1) to the name if it detects it's a duplicate name. If I manually rename New-Scene.gltf, the updated name doesn't appear in the studio unless I refresh. And if I then make a new scene, it gets made as New-Scene.gltf, which is expected, so it's probably the lack of de-duplication that's the problem.
> > When I ran dev-reinit, all static-resources, even scene files, had type 'file', so no scenes showed up in the studio. And are some of those files supposed to be of type 'asset'? Making a new scene worked, and New-Scene was of type 'scene'.
> 
> I also noticed that after dev-reinit, the thumbnail image for scenes in the 'scenes' tab is pointing to the GLTF of the scene, e.g. `projects/default-project/public/scenes/New-Scene.gltf`. This of course does not render a thumbnail.

Due to changes in how dev-reinit works in this branch, you may need to remove and prune the docker image of the db, checkout dev & reinit, and then checkout this branch and prepare-database. If this is what you have been doing, then we may need to sync up to figure this out.

barankyle:
> > I'm still seeing these issues:
> > > When I applied the migration via `prepare-database`, it seemed to work, but when I created a new scene, it didn't immediately appear in the scene list in the studio; I had to refresh. And when I made further new scenes, I did not see anything even after refreshing. The backend might not be doing that thing where it will append e.g. (1) to the name if it detects it's a duplicate name. If I manually rename New-Scene.gltf, the updated name doesn't appear in the studio unless I refresh. And if I then make a new scene, it gets made as New-Scene.gltf, which is expected, so it's probably the lack of de-duplication that's the problem.
> > > When I ran dev-reinit, all static-resources, even scene files, had type 'file', so no scenes showed up in the studio. And are some of those files supposed to be of type 'asset'? Making a new scene worked, and New-Scene was of type 'scene'.
> > 
> > 
> > I also noticed that after dev-reinit, the thumbnail image for scenes in the 'scenes' tab is pointing to the GLTF of the scene, e.g. `projects/default-project/public/scenes/New-Scene.gltf`. This of course does not render a thumbnail.
> 
> Due to changes in how dev-reinit works in this branch, you may need to remove and prune the docker image of the db, checkout dev & reinit, and then checkout this branch and prepare-database. If this is what you have been doing, then we may need to sync up to figure this out.

Wouldn't this permanently minorly break local development if dev-reinit no longer works as intended?

HexaField:
> Wouldn't this permanently minorly break local development if dev-reinit no longer works as intended?

Hanzla and I agreed these changes are correct. Currently, the knex migrations are not wiped when reinitting, leading to migrations not running as expected. This nukes all tables and ensures all migrations always run as needed.

HexaField:
@barankyle addressed all your feedback

barankyle (approved):

PR #10305, opened 2024-06-03
Add networking changes for production grade deployment to IR engine

muradkhateeb78:


barankyle:
This solution turned out to be impossible with how nginx handles UDP forwarding and how UDP addressing works.

PR #10306, opened 2024-06-03
IR-2441 AddComponent search autofocus

SamMazerIR:
## Summary
adding inputRef argument to StringInputProps and StringInput element for new studio. adding auto-focus for search field in elementList/index.tsx

## References
closes [https://tsu.atlassian.net/browse/IR-2441](https://tsu.atlassian.net/browse/IR-2441)

## QA Steps
Click "AddComponent" on an entity in the new studio, the search bar for components should get cursor focus now without requiring the user to click it


aditya-mitra (approved):
Works fine on local!!!

PR #10313, opened 2024-06-04
Michael/ir 1983 registry - Post Processing and Highlight

MbfloydIR:
## Summary
Updated the post processing to use individual reactors that can be injected.
Added Highlight system to outline selected items using post processing.
Created unit tests for Post Processing Component

## References
closes #_insert number here_


packages/spatial/src/renderer/components/PostProcessingComponent.tsx
```diff
@@ -23,29 +23,75 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
+import { CameraComponent } from '../../camera/components/CameraComponent'
+import { HighlightState } from '../HighlightState'
+import { RendererState } from '../RendererState'
+import { RendererComponent } from '../WebGLRendererSystem'
+import { ObjectLayers } from '../constants/ObjectLayers'
+import { PostProcessingEffectState } from '../effects/EffectRegistry'
+import { LinearTosRGBEffect } from '../effects/LinearTosRGBEffect'
 import { useScene } from './SceneComponents'
 
+declare module 'postprocessing' {
```
HexaField, line 47: this effect type override should be moved to each of the respective modules

packages/spatial/src/renderer/components/PostProcessingComponent.tsx
```diff
@@ -23,29 +23,75 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
+
 export const PostProcessingComponent = defineComponent({
   name: 'PostProcessingComponent',
   jsonID: 'EE_postprocessing',
 
   onInit(entity) {
     return {
       enabled: false,
-      effects: defaultPostProcessingSchema
+      effects: {} as Record<string, any> // effect name, parameters
```
HexaField: can we add a better type than `any` for the record value here? we could a base type without effect-specific props perhaps for at least the context of this module

packages/spatial/src/renderer/effects/EffectRegistry.ts
```diff
@@ -0,0 +1,66 @@
...
+import { Entity } from '@etherealengine/ecs'
+import { defineState } from '@etherealengine/hyperflux'
+import { EffectComposer } from 'postprocessing'
+import React from 'react'
+import { Scene } from 'three'
+
+export type EffectReactorProps = {
+  isActive: boolean
+  rendererEntity: Entity
+  effectData
```
HexaField: let's provide types for `effectData` and `effects` here, even if they are just `any`

packages/spatial/tsconfig.json
```diff
@@ -41,5 +41,5 @@
     "../ecs/**/*",
     "./**/*.ts",
     "./**/*.tsx"
-  ]
```
HexaField: please check this is correct as per our indentation

HexaField (changes requested):

packages/spatial/src/renderer/components/PostProcessingComponent.tsx
MbfloydIR: implemented

packages/spatial/src/renderer/components/PostProcessingComponent.tsx
MbfloydIR: implemented

packages/spatial/src/renderer/effects/EffectRegistry.ts
MbfloydIR: implemented

packages/spatial/tsconfig.json
MbfloydIR: implemented

HexaField (approved):

PR #10315, opened 2024-06-04
added the missing confirm dialog so the delete confirmation will show

MbfloydIR:
## Summary
added the missing confirm dialog so the delete confirmation will show

## References
closes #_insert number here_


aditya-mitra (approved):
Thanks for fixing!

PR #10319, opened 2024-06-05
Fix Thumbnail Generation

dinomut1:
Fixes thumbnail generator:
- changes jobs state to load and render one asset at a time to minimize memory usage + reuse one webGL render context
- add lighting to thumbnail render
- fix camera framing + angle
- add support for rigged meshes
- add support for rendering nested gltfs
- change thumbnail resolution to 256x256
- automatically update resources.json with thumbnail URLs
- add progress bar
- fix thumbnail URLs getting populated in static resources when installing project
- make thumbnail URLs relative in resources.json
- fix thumbnail generation registering thumbnails themselves as static resources and trying to give thumbnails thumbnails
- fix race condition in shadow system where shadows are trying to be rendered before the renderer has initialized

Closes https://tsu.atlassian.net/browse/IR-2408
Unblocks https://tsu.atlassian.net/browse/IR-1981


packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -1751,7 +1751,7 @@ export const uploadLocalProjectToProvider = async (
   const resourceDBPath = path.join(projectRootPath, 'resources.json')
   const hasResourceDB = fs.existsSync(resourceDBPath)
 
-  const files = getFilesRecursive(projectRootPath)
+  const files = getFilesRecursive(projectRootPath, false, [path.join(projectRootPath, 'thumbnails')])
```
HexaField, line 1754: rather than overloading getFilesRecursive with an exclusion, we should just filter the results once we have them. the overhead is negligible

aditya-mitra (approved):
Looks Good ✅

PR #10326, opened 2024-06-05
Removed font.css, replaced with meta tag imports in components

barankyle:
## Summary
Import of Google fonts in font.css was breaking first line of later .css file imports because of lack of semicolon at the end. But we should be importing fonts in components via <link>s hoisted into the <head> via <MetaTags>.

## References
closes #_insert number here_


DanielBelmes (approved):

barankyle:
Closing this in favor of one with a more linkable branch name.

PR #10328, opened 2024-06-05
Removed font.css, replaced with meta tag imports in components

barankyle:
## Summary

Import of Google fonts in font.css was breaking first line of later .css file imports because of lack of semicolon at the end. But we should be importing fonts in components via <link>s hoisted into the <head> via <MetaTags>. This fixes IR-2428.

When the user menu is rendered in a modal, e.g. at /studio,its inclusion in an MUI modal was adding some undesired styling, specifically a white background. Made the menu stylings all !important to override that behavior. This fixes IR-2452.

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/components/Editor2Container.tsx
+++ b/packages/editor/src/components/Editor2Container.tsx
@@ -23,6 +23,7 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
+import MetaTags from '@etherealengine/client-core/src/common/components/MetaTags'
 import { PopoverState } from '@etherealengine/client-core/src/common/services/PopoverState'
 import { useRemoveEngineCanvas } from '@etherealengine/client-core/src/hooks/useRemoveEngineCanvas'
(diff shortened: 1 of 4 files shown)


packages/editor/src/components/Editor2Container.tsx
```diff
@@ -149,6 +149,13 @@ const EditorContainer = () => {
 
   return (
     <main className="pointer-events-auto">
+      <MetaTags>
+        <link
+          href="https://fonts.googleapis.com/css2?family=Figtree:ital,wght@0,300..900;1,300..900&display=swap"
+          rel="stylesheet"
+          type="text/css"
+        />
+      </MetaTags>
```
aditya-mitra, line 158: we should add this in client folder
will merge put in this a later PR

aditya-mitra (approved):
Looks good 🚀

packages/editor/src/components/Editor2Container.tsx
SYBIOTE: this doesnt cover the entire project, we need to put it in the head

PR #10330, opened 2024-06-06
Project settings everywhere

the-reality-engineer:
## Summary

Allow all project settings to be used in any project

image

## References
closes #_insert number here_


HexaField (changes requested):
@hanzlamateen is going to refactor project settings to be generic, rather than constrained to a specific schema, so this is unnecessary with that change.

This also will not work with our project permissions / scopes, nor is it a good idea security wise.

PR #10334, opened 2024-06-06
fix bug in getProjectEnv

CITIZENDOT:
## Summary

Project Settings are not read correctly before. Fixed it in this PR.

## References
closes #_insert number here_


packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -410,8 +406,12 @@ export const getProjectEnv = async (app: Application, projectName: string) => {
 
   const projectSetting = project.data?.[0]?.settings || []
 
-  const settings: ProjectSettingType[] = []
-  Object.values(projectSetting).map(({ key, value }) => (settings[key] = value))
+  const settings = {}
```
HexaField, line 409: we can use Object.fromEntries here instead to preserve types implicitly

HexaField (changes requested):

CITIZENDOT:
Just tested again and the bug doesn't actually exist. It was a mistake while testing with different project settings.

Hence closing this PR.

PR #10337, opened 2024-06-06
Changes to move project-setting into a separate table

hanzlamateen:
## Summary

This PR does following things:

1. Move project settings from a json serialize column in project table to a new table named project-setting
2. All old settings will be migrated using a migration file
3. Removes settings from xrengine.config.ts. Now they need to be set in onInstall callback through API call
4. Introduced type column in project-setting which can be public, private
5. private project-setting can be accessed by guests
6. Other settings access requests projects:read/write OR editor:write with project-permissions.
7. Updated admin UI
8. Added tests for project-setting service
9. Updated InstallFunctionType to include project type object as param

Below is how projects should initialized their project settings now instead of using `xrengine.config.ts`:

```
onInstall: async (app: Application, project: ProjectType) => {
    await app.service(projectSettingPath).create({
      key: 'applicationId',
      value: '',
      type: 'private',
      projectId: project.id
    })
}
```
Or multiple as:
```
onInstall: async (app: Application, project: ProjectType) => {
    await app.service(projectSettingPath).create([{
      key: 'applicationId',
      value: '',
      type: 'private',
      projectId: project.id
    }, {
      key: 'clientKey',
      value: '',
      type: 'public',
      projectId: project.id
    }])
}
```

Admin Panel:
image

Database:
image

## References
closes #_insert number here_


barankyle (approved):
Looks good. I had one question - why does GET not have any hooks on it, and allows access to private settings? I'm sure there's a specific reason, just curious what it is.

HexaField (changes requested):
> Looks good. I had one question - why does GET not have any hooks on it, and allows access to private settings? I'm sure there's a specific reason, just curious what it is.

Same with `update`, I think they were just overlooked. We should get these addressed.

barankyle:
> > Looks good. I had one question - why does GET not have any hooks on it, and allows access to private settings? I'm sure there's a specific reason, just curious what it is.
> 
> Same with `update`, I think they were just overlooked. We should get these addressed.

update isn't enabled in project-setting.schema.ts -> projectSettingMethods, so it's not a problem, but maybe the hooks should be added there in case someone ever decides to enabled update.

hanzlamateen:
> Looks good. I had one question - why does GET not have any hooks on it, and allows access to private settings? I'm sure there's a specific reason, just curious what it is.

@barankyle @HexaField The reason why I dont put security on GET endpoint was because previously settings were stored in project table and in project service we dont have any hook restrictions on GET method

https://github.com/EtherealEngine/etherealengine/blob/1c68aaa125ad23ba0288d35bee8541eb50174c08/packages/server-core/src/projects/project/project.hooks.ts#L590

Edit: I have disabled GET endpoint for project-setting service as it wasnt being used anywhere in client.

PR #10339, opened 2024-06-06
Fix directional and spot light addition

SYBIOTE:
This pull request fixes the addition of directional and spot lights in the code. It addresses issues related to the light components and their properties. The changes include fixing the shadow properties for the directional light and updating the color, intensity, penumbra, angle, range, and decay properties for the spot light.


aditya-mitra (approved):
Looks good ✅

PR #10342, opened 2024-06-06
Fix font size in list file view

SYBIOTE:
This pull request fixes the font size issue in the index file. The font size was not being applied correctly to certain elements, causing them to appear larger than intended. With this fix, the font size is now properly applied to all relevant elements, ensuring consistent and visually pleasing typography.


DanielBelmes (approved):
QA'd locally. Looks good

PR #10344, opened 2024-06-06
IR-2377 spin products

SamMazerIR:
## Summary
- Ignore drag and only use click in Interactables
- Helper functions for TransformComponent (right, up, forward)
- More comprehensive options for default interaction aliases (spatial, nonspatial, combined - clicks/triggers/E key)
- no longer setting drag to false on mouse up, should reflect state similarly to pressed

- camera orbit system now requires dragging to be true before it captures input and begins orbiting
- InteractableComponent now requires dragging to be false to capture (this allows orbiting if you started mouseDown over an interactable)
- EditorControlSystem no longer needs accumulator or checking mouse down vs up entity, just checks for no dragging on mouse up instead
- no longer starting GizmoPlane with visibleComponent, this is managed in the TransformGizmoControlComponent when a control entity is dragged
- simplified TransformGizmoControlComponent logical flow, now properly updates VisibleComponent on gizmoPlane

## References
paving the way for ecommerce ticket:
[https://tsu.atlassian.net/browse/IR-2377](https://tsu.atlassian.net/browse/IR-2377)

fixes for single-click selection (gizmo plane was interfering)
[https://tsu.atlassian.net/browse/IR-1293](https://tsu.atlassian.net/browse/IR-1293)

## QA Steps
Interactables should only react to a click, not a click & drag (there is a threshold for cursor travel before it is considered a drag)


packages/spatial/src/transform/components/TransformComponent.ts
```diff
@@ -227,10 +233,32 @@ export const TransformComponent = defineComponent({
     transform.matrixWorld.compose(vec3, quat, scale)
   },
 
+  forward: (entity: Entity) => {
+    const matrixElements = getComponent(entity, TransformComponent).matrix.elements
```
HexaField, line 232: we should have an optional vec3 passed in to these functions that the output is put onto, rather than a single hidden global

HexaField (changes requested):

packages/spatial/src/transform/components/TransformComponent.ts
```diff
@@ -71,7 +74,10 @@ export const TransformComponent = defineComponent({
         new Vector3(1, 1, 1)
       ) as Vector3,
       matrix: new Matrix4(),
-      matrixWorld: new Matrix4()
+      matrixWorld: new Matrix4(),
+      forward: () => TransformComponent.forward(entity),
```
HexaField: this is an anti pattern. component definitions should just store data, not any utility functions

HexaField (dismissed):

packages/spatial/src/transform/components/TransformComponent.ts
SamMazerIR: removed per our discussion

packages/spatial/src/transform/components/TransformComponent.ts
SamMazerIR: changed

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -288,18 +295,54 @@ export const InteractableComponent = defineComponent({
...
         InputState.setCapturingEntity(entity)
+
+        //on releasing continued SpatialInteract click
+        if (buttons.SpatialInteract?.up) {
+          if (inputPointerEntity !== UndefinedEntity) {
+            const inputPointer = getOptionalComponent(inputPointerEntity, InputPointerComponent)
+
+            //TODO store this threshold of 0.001 somewhere (drag threshold)
+            if (inputPointer && inputPointerPosition.distanceToSquared(inputPointer.position) < 0.001) {
+              console.log('distanceSquared = ' + inputPointerPosition.distanceToSquared(inputPointer.position))
```
the-reality-engineer: Please remove the log here

the-reality-engineer (changes requested):

packages/engine/src/interaction/components/InteractableComponent.ts
```diff
@@ -288,18 +295,54 @@ export const InteractableComponent = defineComponent({
...
+              callInteractCallbacks(entity)
+            } else {
+              console.log('not firing interact')
```
the-reality-engineer: And here

packages/spatial/src/input/components/InputComponent.ts
```diff
@@ -47,7 +47,9 @@ import { InputSourceComponent } from './InputSourceComponent'
 export type InputAlias = Record<string, (string | number)[]>
 
 export const DefaultInputAlias = {
-  Interact: [MouseButton.PrimaryClick, XRStandardGamepadButton.Trigger, KeyboardButton.KeyE]
+  MergedInteract: [MouseButton.PrimaryClick, XRStandardGamepadButton.Trigger, KeyboardButton.KeyE],
+  SpatialInteract: [MouseButton.PrimaryClick, XRStandardGamepadButton.Trigger],
+  NonSpatialInteract: [KeyboardButton.KeyE]
```
the-reality-engineer: Not a huge fan of this additional complexity; is this necessary?

SamMazerIR: I'm open to updating this, I do want the ability to use a merged spatial interact though

the-reality-engineer: I still find this confusing as an API surface. I think these aliases should represent categories of user intentions, not categories of input.

the-reality-engineer (changes requested):

the-reality-engineer (approved):

PR #10347, opened 2024-06-07
scene name validation on save as and rename

aditya-mitra:
## Summary

user input scene name validation

## References

https://tsu.atlassian.net/browse/IR-2453

https://tsu.atlassian.net/browse/IR-2531

https://tsu.atlassian.net/browse/IR-1446

## QA Steps

image

image


achen5671 (approved):

SYBIOTE (approved):
looks good, is it worth giving the valid scene name conditions its own variable so its easily reusable in the future

HexaField (changes requested):
we should make the same restrictions on the backend

aditya-mitra:
> we should make the same restrictions on the backend

done and added the validation in common package

packages/common/src/utils/validateSceneName.ts
```diff
@@ -0,0 +1,28 @@
...
+
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Ethereal Engine team.
+
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023 
+Ethereal Engine. All Rights Reserved.
+*/
+
+export default function isValidSceneName(sceneName: string) {
+  return sceneName.length >= 3 && sceneName.length <= 64 && sceneName.indexOf('_') === -1
```
hanzlamateen: Is it checking for valid file name? Because I think some charaters like `/`, etc are also not allowed in file names.

Ref:
https://www.npmjs.com/package/filename-reserved-regex
https://github.com/sindresorhus/filename-reserved-regex/blob/main/index.js

Maybe we can have all common regex in `/packages/common/src/constants/regex.ts` for reuseability?

hanzlamateen (changes requested):

packages/common/src/utils/validateSceneName.ts
aditya-mitra: Thanks a lot for pointing this out. Added these and updated the regex to include `_`
currently since we do not have common file for putting all our regex in, we can refactor all our regex used across packages to such a regex file.

HexaField (approved):

PR #10348, opened 2024-06-07
studio:  polish existing prop components

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-2521

This PR polishes existing components to match the figma design. There are more components that needs to be cleaned up in this section but I wanted to break it out to a separate PR since those component need to be migrated in its entirety using tailwind i.e `Post Processing`, `Media Setting`, `Scene Setting`, etc. This PR also already touches 23 files so it might be better to break it up

This PR polish
`Camera`, `Transform`, `EnvMap Bake`, `Scene Preview Camera`, `Volumetric`, etc

## References
closes #_insert number here_


packages/ui/src/components/editor/input/Vector3/index.tsx
```diff
@@ -124,13 +124,13 @@ export const Vector3Input = ({
   const vz = value.z
 
   return (
-    <div className="flex flex-wrap justify-start gap-1.5">
+    <div className="flex flex-row flex-nowrap justify-start gap-1.5">
```
aditya-mitra: why was `flex-nowrap` added?
on smaller screens the vector input is overflowing
image

packages/ui/src/primitives/tailwind/Select/index.tsx
```diff
@@ -144,10 +144,11 @@ const Select = <T extends OptionValueType>({
         endComponent={
           <MdOutlineKeyboardArrowDown
             size="1.5em"
-            className={`text-theme-primary transition-transform ${showOptions.value ? 'rotate-180' : ''}`}
+            className={`text-[#444444] transition-transform ${showOptions.value ? 'rotate-180' : ''}`}
             onClick={toggleDropdown}
           />
         }
+        containerClassname="rounded-md"
```
aditya-mitra: why did we add border radius here?
[in designs](https://www.figma.com/design/ln2VDACenFEkjVeHkowxyi/iR-Engine-Design-Library-File?node-id=2508-3527&t=FC7FlAaAzLUoVy6o-0), select primitive does not have a border radius

packages/ui/src/primitives/tailwind/Input/index.tsx
```diff
@@ -97,7 +97,7 @@ const Input = forwardRef<HTMLInputElement, InputProps>(
     }
 
     const twcontainerClassName = twMerge(
-      'flex w-full flex-col items-center ',
+      'flex flex-col items-center overflow-hidden',
```
aditya-mitra: There are some distortions like the focus in the admin panel. Primitive behaviour like full width needs to be changed with every component using it.
image

packages/ui/src/components/editor/properties/envMapBake/index.tsx
```diff
@@ -162,7 +162,9 @@ export const EnvMapBakeNodeEditor = (props) => {
       description="For Adding EnvMap bake in your scene"
       icon={<IoMapOutline />}
     >
-      <Button onClick={() => uploadBPCEMBakeToServer(props.entity)}>Bake</Button>
+      <Button className="my-1 ml-auto mr-6 px-10" onClick={() => uploadBPCEMBakeToServer(props.entity)}>
+        Bake
```
aditya-mitra: could you add this to translation file?

aditya-mitra (changes requested):
Great work on the detailing 🔥 They look nicer now!
A few amendments like translations and reverting the changes to primitives would get this PR ready to merge.

packages/ui/src/components/editor/input/Vector3/index.tsx
achen5671: resolved

packages/ui/src/primitives/tailwind/Select/index.tsx
achen5671: oh I didnt realized we had this file for primitive. will resolve

aditya-mitra (approved):

PR #10351, opened 2024-06-07
[IR-2561] studio: avoid locking studio with generating thumbnails indicator

dtlehrer:
# Summary
resolves [IR-2561] by setting pointer-events to 'none' for the ProgressBar.. also reduces height of the component and adds some light styling to avoid overlaying the entire studio with the black bg color

eventually we may want to add additional styling or get rid of this useEffect to avoid showing the ProgressBar altogether:
```
useEffect(() => {
    if (jobState.length > 0) {
      const newJob = jobState[0].get(NO_PROXY)
      currentJob.set(JSON.parse(JSON.stringify(newJob)))
      getMutableState(ProgressBarState)['thumbnail-generation'].set(<ProgressBar />)
    } else {
      getMutableState(ProgressBarState)['thumbnail-generation'].set(none)
      cleanupState()
    }
  }, [jobState.length])
```

[IR-2561]: https://tsu.atlassian.net/browse/IR-2561?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/client-core/src/common/services/FileThumbnailJobState.tsx
+++ b/packages/client-core/src/common/services/FileThumbnailJobState.tsx
@@ -129,19 +129,12 @@ const ProgressBar = () => {
   return (
     <div
       key="thumbnail-generation-progress-bar"
-      style={{
-        position: 'fixed',
-        width: '100%',
-        height: '100%',
-        backgroundColor: 'rgba(0, 0, 0, 0.5)',
-        display: 'flex',
-        justifyContent: 'center',
-        zIndex: 1000,
-        padding: '1em'
-      }}
+      className="pointer-events-none fixed inset-x-0 top-0 z-[1000] m-2 flex justify-center"
     >
-      <div>Generating Thumbnails</div>
-      <div>{thumbnailJobState.length} remaining</div>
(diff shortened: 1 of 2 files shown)


packages/client-core/src/common/services/FileThumbnailJobState.tsx
```diff
@@ -129,19 +129,12 @@ const ProgressBar = () => {
...
-        justifyContent: 'center',
-        zIndex: 1000,
-        padding: '1em'
-      }}
+      className="pointer-events-none fixed inset-x-0 top-0 z-[1000] m-2 flex justify-center"
     >
-      <div>Generating Thumbnails</div>
-      <div>{thumbnailJobState.length} remaining</div>
+      <div className="flex max-w-[500px] flex-col items-center rounded-md bg-gray-700 bg-opacity-50 px-4 py-2">
+        <div>Generating Thumbnails</div>
```
HexaField: can we add i18n for these too

HexaField (dismissed):

dinomut1 (approved):

PR #10353, opened 2024-06-10
IR-2549: create static resource using project name from request data

jose-galvan:
## Summary

When static resources are uploaded through the file browser service the project field is assigned by splitting the file path 
and using the first segment after project. i.e  for a file with the path `projects/default-project/drop-shadow.png`
the project would be `default-project`, however this approach doesn't work for multi-tenancy projects where the first segment of the path after `projects/` only identifies the account name. To support both single-tenant and multi-tenancy projects the project name is now attached to the request, rather than being extracted from the file path.

## References
closes #_insert number here_

## QA Steps

- Go into the studio. 
- Click on Files in the assets sections located at the bottom of the screen 
- Click on Upload Assets 
- Select a new file. 
- Verify in the db the static resource has been created with the appropriate project name.

### DIFF
--- a/packages/editor/src/components/assets/FileBrowser/FileBrowserContentPanel.tsx
+++ b/packages/editor/src/components/assets/FileBrowser/FileBrowserContentPanel.tsx
@@ -66,7 +66,6 @@ import LoadingView from '@etherealengine/ui/src/primitives/tailwind/LoadingView'
 import { SupportedFileTypes } from '../../../constants/AssetTypes'
 import { downloadBlobAsZip, inputFileWithAddToScene } from '../../../functions/assetFunctions'
 import { bytesToSize, unique } from '../../../functions/utils'
-import { EditorState } from '../../../services/EditorServices'
 import BooleanInput from '../../inputs/BooleanInput'
 import InputGroup from '../../inputs/InputGroup'
 import StringInput from '../../inputs/StringInput'
@@ -84,7 +83,7 @@ import { FilePropertiesPanel } from './FilePropertiesPanel'
 type FileBrowserContentPanelProps = {
   onSelectionChanged: (assetSelectionChange: AssetSelectionChangePropsType) => void
   disableDnD?: boolean
-  projectName?: string
+  projectName: string
   selectedFile?: string
   folderName?: string
   nestingDirectory?: string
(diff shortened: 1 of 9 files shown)


hanzlamateen:
@HexaField can you please have a look of this

packages/ui/src/components/editor/panels/Files/container/index.tsx
```diff
@@ -140,6 +140,8 @@ const FileBrowserContentPanel: React.FC<FileBrowserContentPanelProps> = (props)
   const filesViewMode = useHookstate(getMutableState(FilesViewModeState).viewMode)
   const [anchorPosition, setAnchorPosition] = React.useState<any>(undefined)
 
+  const { projectName } = useMutableState(EditorState)
```
HexaField: editor state is not the correct data to use here. we need to somehow get it from the file browser's state

HexaField (changes requested):
I didn't realise there was already for a task for this, aditya did it here based off the PR I have that changes a lot of the backend services

https://github.com/EtherealEngine/etherealengine/pull/10354/files

packages/editor/src/components/assets/FileBrowser/FileBrowserContentPanel.tsx
```diff
@@ -130,7 +131,7 @@ const FileBrowserContentPanel: React.FC<FileBrowserContentPanelProps> = (props)
   const selectedDirectory = useHookstate(originalPath)
   const nestingDirectory = useHookstate(props.nestingDirectory || 'projects')
   const fileProperties = useHookstate<FileType | null>(null)
-
+  const { projectName } = useMutableState(EditorState)
```
HexaField: editor state is not the correct data to use here. we need to somehow get it from the file browser's state

HexaField (changes requested):

packages/editor/src/components/assets/FileBrowser/FileBrowserContentPanel.tsx
jose-galvan: Changed to a component prop.

packages/ui/src/components/editor/panels/Files/container/index.tsx
jose-galvan: Changed to a component prop.

PR #10358, opened 2024-06-11
Added code to prevent default scene files being moved/deleted

barankyle:
## Summary

Since new scene creation runs off of default-project/public/scenes/default.gltf, if this file has been moved, scene creation will break. The backend now disallows moving or deleting this scene and its related files.

Resolves IR-2566.

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/util/upload.tsx
+++ b/packages/client-core/src/util/upload.tsx
@@ -72,12 +72,12 @@ export const uploadToFeathersService = (
             resolve(JSON.parse(request.responseText))
           } else {
             if (aborted) return
+            const errorResponse = JSON.parse(request.responseText)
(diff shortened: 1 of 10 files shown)


HexaField (changes requested):
Instead of specific paths, I think we should just disallow any operations to the default project completely, if `config.syncFileSystem` or whatever it is is false.

barankyle:
> Instead of specific paths, I think we should just disallow any operations to the default project completely, if `config.syncFileSystem` or whatever it is is false.

That's fine with me too. Good idea about tying it to syncFileSystem.

HexaField:
needs to be reimplemented around new CMS

DanielBelmes:
@HexaField did you mean to close this one?

PR #10360, opened 2024-06-11
Fixed Project Settings table

MoizAdnan:


hanzlamateen (changes requested):
It seems like this PR sends update request for all settings for a project. Even if just 1 is updated. Which is not desired. We just want to send request for the updated ones, so that the database updatedAt and userId field reflects correct values.

hanzlamateen (approved):

PR #10365, opened 2024-06-11
prettifies resource json on write

dinomut1:
Makes resources.json human readable on write


HexaField (changes requested):
We use 2 not 4 elsewhere, we should standardize on 2

Rezmason (approved):

PR #10366, opened 2024-06-11
IR-2554 Facer Component

dinomut1:
Adds a Facer Component, which makes the component entity look at either the viewer, or a specific entity in the scene.
Implements a NodeInput UI component which allows nodes to be dragged and dropped onto it.
Changes the behavior of hierarchy nodes such that they are not selected immediately on mouse down but on click instead. Also changes their drag behavior to allow dragging of non-selected nodes.

Closes https://tsu.atlassian.net/browse/IR-2554

### DIFF
--- a/packages/ui/src/components/editor/properties/facer/index.tsx
+++ b/packages/ui/src/components/editor/properties/facer/index.tsx
@@ -1,0 +1,68 @@
+/*
+CPAL-1.0 License
+
+The contents of this file are subject to the Common Public Attribution License
+Version 1.0. (the "License"); you may not use this file except in compliance
+with the License. You may obtain a copy of the License at
+https://github.com/EtherealEngine/etherealengine/blob/dev/LICENSE.
+The License is based on the Mozilla Public License Version 1.1, but Sections 14
+and 15 have been added to cover use of software over a computer network and 
+provide for limited attribution for the Original Developer. In addition, 
+Exhibit A has been modified to be consistent with Exhibit B.
+
+Software distributed under the License is distributed on an "AS IS" basis,
+WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for the
+specific language governing rights and limitations under the License.
+
+The Original Code is Ethereal Engine.
+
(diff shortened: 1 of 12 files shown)


packages/ui/src/components/editor/properties/facer/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const facerComponent = useComponent(props.entity, FacerComponent)
+
+  return (
+    <NodeEditor
+      entity={props.entity}
+      component={FacerComponent}
+      name={t('editor:properties.facer.name')}
+      description={t('editor:properties.facer.description')}
+    >
+      <InputGroup name="Target" label={t('editor:properties.facer.target')}>
```
DanielBelmes, line 56: there is an example in RenderSettingsNodeEditor

```ts
  const directionalLightOptions = [
    {
      label: 'None',
      value: '' as EntityUUID
    }
  ].concat(
    useQuery([DirectionalLightComponent]).map((entity) => {
      return {
        label: getComponent(entity, NameComponent),
        value: getComponent(entity, UUIDComponent)
      }
    })
  )
```

It's for getting a list of options with the name of entity and its uuid automatically tied to it. Could be a better option for this input as it's more usable.

dinomut1: I'm wanting to avoid that type of selection since it gets confusing with large scenes (this dropdown in this context would include every single entity with a transform component in the entire scene). Trying to follow Unreal + Unity in allowing drag and drop as the primary method of setting scene objects as inputs

DanielBelmes: Sounds good

HexaField (dismissed):
Can we rename this to something like `LookAtComponent` and make it generic to look at any entity? It will be much more reusable that way.

dinomut1:
> Can we rename this to something like `LookAtComponent` and make it generic to look at any entity? It will be much more reusable that way.

It already does this

the-reality-engineer:
LookAtComponent would be a much better name for this, imo.

PR #10367, opened 2024-06-12
studio: quit to dashboard for mt (breaks on single tenant)

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-2621

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/components/projects/ProjectsPage.tsx
+++ b/packages/editor/src/components/projects/ProjectsPage.tsx
@@ -182,6 +182,7 @@ const ProjectsPage = ({ studioPath }: { studioPath: string }) => {
   const editPermissionsDialogOpen = useHookstate(false)
   const projectDrawerOpen = useHookstate(false)
   const changeDestination = useHookstate(false)
+  const selfUser = useHookstate(getMutableState(AuthState)).user
 
   const navigate = useNavigate()
   const hasWriteAccess =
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/projects/ProjectsPage.tsx
```diff
@@ -212,7 +213,11 @@ const ProjectsPage = ({ studioPath }: { studioPath: string }) => {
     }
   })
 
-  const installedProjects = projectFindQuery.data.filter(() => projectCategoryFilter.installed.value)
+  const isOwner =
+    projectPermissionsFindQuery.data?.find((permission) => permission.userId === selfUser.id.value)?.type === 'owner'
```
DanielBelmes, line 217: I'm assuming we also want to show projects that you have `editor` permissions on. I am unsure what to do about `reviewer` permissions

HexaField (changes requested):
This solution breaks the base engine. it must be implemented in the MT project instead

HexaField:
cc @hanzlamateen my proposed solution for this is here https://github.com/theinfinitereality/eepro-multitenancy/pull/250

DanielBelmes:
@HexaField ah I was told this was for single tenant as well. That in single tenant we wanted to hide projects you don't have permissions for. From reading the ticket I think this is for MT only and we should close the PR.

hanzlamateen:
> cc @hanzlamateen my proposed solution for this is here https://github.com/theinfinitereality/eepro-multitenancy/pull/250

Sounds good. I just requested some changes in that PR

PR #10369, opened 2024-06-12
fix(IR-2630): fix modal component

kattsushi:
## References

Screenshot 2024-06-11 at 11 25 28 p m

linked to #_insert number here_


packages/ui/src/components/editor/panels/Assets/container/index.tsx
```diff
@@ -222,7 +222,7 @@ const AssetPanel = () => {
 
   const CategoriesList = () => {
     return (
-      <div className="mb-8 h-[100%] w-[200px] overflow-y-auto bg-[#0E0F11] pb-8">
+      <div className="mb-8 h-[100%] w-[200px] overflow-y-auto bg-neutral-950 pb-8">
```
aditya-mitra: can you please give me a link to this figma design?

aditya-mitra (changes requested):

packages/ui/src/primitives/tailwind/Modal/index.tsx
```diff
@@ -43,26 +43,41 @@ export interface ModalProps {
...
+  actionsOnHeader?: boolean
+  submitButtonText?: string
+  onSubmit?: () => void
   onClose?: (isHeader: boolean) => void
 }) => {
   // sticky top-0 z-10 bg-theme-surface-main
+  const { t } = useTranslation()
   return (
-    <div className="relative flex items-center justify-center border-b border-b-theme-primary px-6 py-5">
+    <div className="relative flex items-center justify-between border-b border-b-theme-primary px-6 py-5">
```
aditya-mitra, line 71: according to the admin figma designs, text should be centered in the modal

https://www.figma.com/design/9F1J9ZuxkCbimWUzjzDzTm/Ready-for-dev%3A-Dashboard-%26-Admin?node-id=1-11949&t=UF1WY267SMoHc2V6-0

currently we are having it like this
image

aditya-mitra (changes requested):

packages/ui/src/primitives/tailwind/Modal/index.tsx
```diff
@@ -87,20 +103,23 @@ export const ModalFooter = ({
...
   submitButtonText?: string
   showCloseButton?: boolean
+  startSubmitIcon?: ReactNode
 }) => {
   const { t } = useTranslation()
   return (
     <div className="grid grid-flow-col border-t border-t-theme-primary px-6 py-5">
       {showCloseButton && (
-        <Button variant="outline" disabled={closeButtonDisabled} onClick={() => onCancel && onCancel(false)}>
+        <Button variant="secondary" disabled={closeButtonDisabled} onClick={() => onCancel && onCancel(false)}>
```
aditya-mitra, line 112: this button should be `outline` as per figma designs on the admin

https://www.figma.com/design/9F1J9ZuxkCbimWUzjzDzTm/Ready-for-dev%3A-Dashboard-%26-Admin?node-id=1-11949&t=UF1WY267SMoHc2V6-0

packages/ui/src/primitives/tailwind/Modal/index.tsx
```diff
@@ -43,26 +43,41 @@ export interface ModalProps {
...
+  submitButtonText,
+  onSubmit,
   onClose
 }: {
   closeIcon?: boolean
   title?: string
+  startSubmitIcon?: ReactNode
+  actionsOnHeader?: boolean
+  submitButtonText?: string
+  onSubmit?: () => void
```
aditya-mitra, line 65: we are adding too many props to the header component which is an anti-pattern
in these cases, we should create our own derived component for the specific use case
For example, the ui/src/components/editor folder does this derivations from the primitives for its components.

aditya-mitra (changes requested):
Are the changes to the modal a one-off or two-off case case?
If so, we should make a separate component since these props will be rarely used

packages/ui/src/components/editor/panels/Assets/container/index.tsx
hanzlamateen: https://www.figma.com/design/Ds2NEvvz9nqddltP3lsKoP/Design-Changes?node-id=0-1&t=YLp6WBatFXncmfY7-0

packages/ui/src/primitives/tailwind/Modal/index.tsx
hanzlamateen: You can ask this from UI team on figma as to what should be the unified standard.

https://www.figma.com/design/Ds2NEvvz9nqddltP3lsKoP/Design-Changes?node-id=0-1&t=YLp6WBatFXncmfY7-0

packages/ui/src/primitives/tailwind/Modal/index.tsx
kattsushi: yes maybe we need to rethink this because right now the ThemeService works for every project that use the talwind components, we need to create a theme for each project

packages/ui/src/primitives/tailwind/Modal/index.tsx
kattsushi: yes maybe in this case should work better using a new variant, i don't think so we need to create another component because how the primitives are build doesn't have very much extensibility probably the primitives should be headless and have separation of concerns using named children pattern to achieve that extensibility

packages/ui/src/primitives/tailwind/Modal/index.tsx
aditya-mitra: yes we can extend the tailwind config in that case

packages/ui/src/primitives/tailwind/Modal/index.tsx
aditya-mitra: The `Modal` is a high level component and made to be configurable but not extendable (since there can be entirely different ways it can be implemented). This [is a one-off case](https://www.figma.com/design/Ds2NEvvz9nqddltP3lsKoP/Design-Changes?node-id=0-1&t=YLp6WBatFXncmfY7-0) which can be put in the multitenancy project with a specific modal title or in this file with a specific name like `HeaderActionsModal`.

PR #10371, opened 2024-06-12
making file list multi select

Adam-Zimmer:
## Summary
File list is now selectable and multi selectable
## References
closes #_insert number here_


DanielBelmes (dismissed):
While I can visually see multiple items being selected. I can not do anything with multiselect. I can't drag multiple files to another folder. I can't drag multiple into scene.

Adam-Zimmer:
@DanielBelmes yeah that is what we decided on for MVP functionality in standup on Monday, in future tickets we will implement actions that can be taken against the selected items. Happy to open up the discussion again though if we want to include actions in this PR as well.

packages/ui/src/components/editor/panels/Files/browserGrid/index.css
```diff
@@ -0,0 +1,4 @@
+.selected-file {
+  background: rgba(64, 92, 169, .2);
+  border-radius: 6px;
+}
```
aditya-mitra: since we are using tailwind, we are not using external css files anymore

packages/ui/src/components/editor/panels/Files/container/index.tsx
```diff
@@ -385,6 +385,7 @@ const FileBrowserContentPanel: React.FC<FileBrowserContentPanelProps> = (props)
       drop: (dropItem) => dropItemsOnPanel(dropItem as any),
       collect: (monitor) => ({ isFileDropOver: monitor.canDrop() && monitor.isOver() })
     })
+    const [selectedKeys, setSelectedKeys] = useState<string[]>([])
```
aditya-mitra: we should use `useHookstate` for state usage

packages/ui/src/components/editor/panels/Files/container/index.tsx
```diff
@@ -407,6 +408,32 @@ const FileBrowserContentPanel: React.FC<FileBrowserContentPanelProps> = (props)
       staticResourceModifiedDates.set(modifiedDates)
     }, [staticResourceData.data])
 
+    const handleItemClick = useCallback(
```
aditya-mitra: since this function is used in the same component and is unused in dependency arrays, we should avoid putting it in useCallback as it will be a unnecessary memoization.

packages/ui/src/components/editor/panels/Files/container/index.tsx
```diff
@@ -407,6 +408,32 @@ const FileBrowserContentPanel: React.FC<FileBrowserContentPanelProps> = (props)
...
+      (e, key) => {
```
aditya-mitra: please add the type for MouseEvent `e`

aditya-mitra (changes requested):
Good work on the new functionality 🔥 We need a few changes though!

aditya-mitra (changes requested):
in hookstate, to get the value we use `selectedKeys.value` instead of `selectedKeys.get()`
also there is no way to deselect all the selected files

aditya-mitra (approved):

PR #10374, opened 2024-06-12
Michael/ir 2056 render unit tests

MbfloydIR:
## Summary
mock renderer and write tests for WebGLRendererSystem rendering loop
ensure background, environment, fog affect the scene correctly
ensure only visible entities are rendered and non-visible children are snubbed
ensure resizing works correctly
WebGLRendererSystem reactors work as expected
PostProcessingComponent has unit tests
FogSystem works as expected
## References
closes #_insert number here_


packages/spatial/src/renderer/WebGLRendererSystem.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+    await act(() => rerender(tag))
+
+    const camera = getComponent(rootEntity, CameraComponent)
+    const rendererComp = getComponent(rootEntity, RendererComponent)
+    const effectComposer = rendererComp.effectComposer
+    const passes = effectComposer?.passes.filter((p) => p.name === 'RenderPass') as any
+    const renderPass: RenderPass = passes ? passes[0] : undefined
+
+    const performanceState = getMutableState(PerformanceState)
```
MichaelEstes: I'd remove this, PerformanceState has it's own tests and that cpuTier value as a constant number is subject to change

packages/spatial/src/renderer/WebGLRendererSystem.test.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+  afterEach(() => {
+    return destroyEngine()
+  })
+
+  it('test', async () => {
+    const { background, environment, fog, children } = getSceneParameters([rootEntity])
+    SystemDefinitions.get(WebGLRendererSystem)?.execute()
+    globalThis._scene
+    console.log('test')
```
DanielBelmes: extra console log here and the name of the test is not super descriptive

MbfloydIR: corrected

packages/spatial/src/renderer/components/FogSettingsComponent.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    fogSettingsComponent.type.set(FogType.Height)
+    fogSettingsComponent.color.set('#ff0000')
+    fogSettingsComponent.far.set(2000)
+    fogSettingsComponent.near.set(2)
+    fogSettingsComponent.density.set(0.02)
+    const { rerender, unmount } = render(<></>)
+
+    await act(() => {
+      rerender(<></>)
+    })
```
DanielBelmes, line 94: So after fixing fog with @dinomut1  and team there's one thing we should add into here and that is making sure the ShaderChunks from three.js are overrided. FogShader.ts I believe holds the functions that we use to override the shaders when brownian and height fog is selected. So we just need to assert that ShaderChunk.fog_fragment etc... are equal to our overrides. I can help with further context.

DanielBelmes (changes requested):
Webgl rendering looked good to me. Fog is mostly good but see comment. Im willing to do the change as well if I have time over the weekend or on monday. But I wanna update it before merge for sure.

packages/spatial/src/renderer/WebGLRendererSystem.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+    assert(camera.layers.isEnabled(ObjectLayers.PhysicsHelper), 'enable physicsDebug')
+    assert(camera.layers.isEnabled(ObjectLayers.AvatarHelper), 'enable avatarDebug')
+    assert(camera.layers.isEnabled(ObjectLayers.Gizmos), 'enable gridVisibility')
+    assert(camera.layers.isEnabled(ObjectLayers.NodeHelper), 'enable nodeHelperVisibility')
+
+    webGLRendererSystem?.execute()
+
+    assert(!rendererComp.needsResize, 'resize updated')
+    assert(
+      globalThis._scene.children.length == 1 && globalThis._scene.children[0].entity == visibleEntity,
```
HexaField: we should not rely on `globalThis._scene`, rather than assert the data put onto the _scene object, we should either change the implementation of the function that modifies it to be more functionally oriented (smaller composed functions that do less, and are more pure, such that they are more easily testable), or test the data some other way (such as overriding/mocking the render function and asserting the data passed to it)

HexaField (changes requested):

DanielBelmes (approved):

HexaField (approved):

PR #10375, opened 2024-06-12
IR-2555: Add controls for alpha masking and custom uv offset + scaling for Video Component

dinomut1:
Closes https://tsu.atlassian.net/browse/IR-2555

### DIFF
--- a/packages/client-core/i18n/en/editor.json
+++ b/packages/client-core/i18n/en/editor.json
@@ -867,6 +867,7 @@
       "description": "Display a video",
       "lbl-projection": "Projection",
       "lbl-side": "Side",
+      "lbl-side-info": "The side of the video to display",
       "lbl-use-alpha": "Use Alpha",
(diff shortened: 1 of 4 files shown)


AidanCaruso:
In testing with the sample video in the default project, it appears the video texture value is always null resulting in the shader color being red. The media component does have the resource url set, on dev it seems to render properly.
image

DanielBelmes:
image
Updates no longer red and alpha works

dinomut1:
I'm seeing a problem with our implementation for looping videos. One of our reactors is running every frame and repeatedly setting the image of the video, which causes the video texture to have a new "version" for every frame. When I changed the reactor not to happen every frame videos don't loop anymore

dinomut1:
ok i'm just merging this in after convo with daniel

DanielBelmes (commented):
Post approval

PR #10378, opened 2024-06-13
Added cache busting to studio asset requests

barankyle:
## Summary

Users were encounter issues in the studio where saving a scene, then immediately refreshing it, could result in the cached old version of the scene (or assets) being retrieved because the cache invalidation had not been resolved yet.

Made a new function to add a unique hash onto the query params of asset fetches
if in the editor. This will bust the cache every time so that this stale loading
does not occur. Currently the hash is epoch time in seconds; in the future it
should be the hash of the file if that information is easily available to the
client (retrieving it right now would be cumbersome and slow).

## References
closes #_insert number here_


packages/engine/src/assets/functions/resourceLoaderHooks.ts
```diff
@@ -57,7 +59,7 @@ function useLoader<T extends ResourceAssetType>(
   }, [])
 
   useImmediateEffect(() => {
-    const _url = url
+    const _url = getState(EngineState).isEditing ? bustURLCache(url) : url
```
MichaelEstes, line 62: If this issue is just for server cache busting the LoadingManager that this code runs through has a url modifier callback that might be a better place for this logic so our internal caches don't get busted

packages/ui/src/components/editor/panels/Files/icon/index.tsx
```diff
@@ -75,7 +78,7 @@ export const FileIcon = ({
   color?: string
 }) => {
   const FallbackIcon = FileIconType[type ?? '']
-
+  thumbnailURL = getState(EngineState).isEditing ? bustURLCache(thumbnailURL) : thumbnailURL
```
HexaField, line 81: everywhere this is done in the body of a react component, it will bust the cache every time it re-renders if it's been more than a second since the last re-render, which could re-request very often

HexaField (changes requested):

packages/ui/src/components/editor/panels/Files/icon/index.tsx
barankyle: What do you think would be a better option here? I'd originally considered making bustURLCache use every 10 seconds as the default hash, but thought that might still result in the old-version cache getting loaded in some cases. I suppose with File Icons loading an old version might be not particularly problematic, should we just remove the cache busting here?

HexaField: i think the ideal solution is to use our CMS generated hashes everywhere, following the cms branch. pending that, i don't see there being a realistic way of scaling this solution well

barankyle: I agree, which is why bustURLCache can also take in a hash instead of using the timestamp default hash; I assumed we'd pass the CMS-generated hashes once that's in. I wasn't sure which PR would get merged first - if this gets in before the CMS branch, then once that's in someone would just need to pass the CMS hash along to bustURLCache in a few places. If CMS gets in first I can modify this PR to do that CMS hash passing.

The default timestamp hash was to have *something* as a default, and the timestamp seemed unique enough.

HexaField:
closing in favour of https://github.com/EtherealEngine/etherealengine/pull/10411

PR #10381, opened 2024-06-13
chore: Remove discardQuery('project') in static-resource.hooks

MoizAdnan:


hanzlamateen:
This quick fix ensures only resource of queried project are removed. Currently, its removing all static-resource.

hanzlamateen (approved):

HexaField:
What is projectId used for?

hanzlamateen:
> What is projectId used for?

image

So projectId, is something that is created/populated by `resolveProjectId` hook in above screenshot. This hook retrieve projectId based on `project` field in query which is project's name.

This projectId is then used in `verifyProjectPermission` hook. Other than this there is no use of it.

@jose-galvan right?

PR #10384, opened 2024-06-13
Ir 2543 create prefab

JT00y:
## References
closes #_insert number here_


packages/engine/src/scene/components/ModelComponent.tsx
```diff
@@ -80,7 +80,8 @@ export const ModelComponent = defineComponent({
       convertToVRM: false,
       scene: null as Group | null,
       asset: null as VRM | GLTF | null,
-      dereference: false
+      dereference: false,
+      save: false
```
dinomut1: we shouldn't add this "save" attribute, this should be handled through just calling "exportModelGLTF" explicitly

packages/editor/src/components/properties/ModelNodeEditor.tsx
```diff
@@ -120,6 +140,13 @@ export const ModelNodeEditor: EditorComponentType = (props) => {
     )
   }, [modelComponent.asset])
 
+  useEffect(() => {
+    if (modelComponent.save.value) {
+      onExportPrefabModel()
```
dinomut1: let's just call this function rather than using a reactor to set it off

dinomut1 (changes requested):

packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+      EditorControlFunctions.addOrRemoveComponent([entity], ModelComponent, true)
+      setComponent(entity, ModelComponent)
+    }
+    const editorState = getState(EditorState)
+    const fileName = defaultPrefabFolder + '/' + prefabName + '.gltf'
+    const srcProject = editorState.projectName!
+    try {
+      exportRelativeGLTF(entity, srcProject, fileName)
+      //pass static resource
+      const resources = await Engine.instance.api.service(staticResourcePath).find({
```
dinomut1, line 74: this will always fail because the file doesn't exist yet

dinomut1: we should remove this call, and construct a static resource object to pass for upload

packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+        query: { key: 'projects/' + srcProject + fileName }
+      })
+      if (resources.data.length === 0) {
+        throw new Error('User not found')
+      }
+      const resource = resources.data[0]
+      await Engine.instance.api.service(staticResourcePath).patch(resource.id, { tags: prefabTag })
```
dinomut1: This needs to be a "create" call instead of "patch"

packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
```diff
@@ -0,0 +1,124 @@
...
+import { EditorControlFunctions } from '../../functions/EditorControlFunctions'
+import { exportRelativeGLTF } from '../../functions/exportGLTF'
+import { EditorState } from '../../services/EditorServices'
+import { HeirarchyTreeNodeType } from '../hierarchy/HeirarchyTreeWalker'
+
+export default function CreatePrefabPanel({ node }: { node?: HeirarchyTreeNodeType }) {
+  const entity = node?.entity as Entity
+  const [defaultPrefabFolder, setDefaultPrefabFolder] = useState<string>('/assets/custom-prefabs')
+  const [prefabName, setPrefabName] = useState<string>('prefab')
+  const [prefabTag, setPrefabTag] = useState<string[]>([])
```
dinomut1: Change these to Hookstate objects instead of normal react state

dinomut1 (changes requested):

dinomut1 (changes requested):
Getting close. We need to fix the export of .bin + texture files in .gltf export mode. I tried making a prefab out of the portal frame in the default project and ran into this issue:
[Screencast from 06-17-2024 03:06:49 PM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/d5a9d831-be4b-4b85-9200-eda684f99e56)
The 'model-resources' folder is getting saved into the wrong location. When saving the .gtlf into the `assets/custom-prefabs` folder, the model resources are getting saved into `assets/assets/custom-prefabs/model-resources/...`:
[Screencast from 06-17-2024 03:09:12 PM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/018113e3-d628-4fdc-8623-4db0101a4918)

dinomut1 (approved):
Looks good

PR #10388, opened 2024-06-14
IR-2381 adding custom scrollbar with transparent background to our tailwind config for use with the xrui scrolling

SamMazerIR:
## Summary
adding custom scrollbar with transparent background to our tailwind config for use with the xrui scrolling

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -74,9 +74,44 @@ module.exports = {
       },
       colors: {
         'blue-primary': '#375DAF'
+      },
+      scrollbar: {
+        'transparent-track': {
+          'thumb': 'rgba(255, 255, 255, 0.5)',
+          'track': 'transparent',
+          'width': '6px',
+          'border-radius': '4px'
+        },
       }
     }
   },
+  plugins: [
+    function ({ addUtilities }) {
+      addUtilities({
+        '.scrollbar-thin': {
+          'scrollbar-width': 'thin',
+        },
+        '.scrollbar-thumb': {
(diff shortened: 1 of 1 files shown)


tailwind.config.js
```diff
@@ -74,9 +74,44 @@ module.exports = {
...
+          'track': 'transparent',
+          'width': '6px',
+          'border-radius': '4px'
+        },
       }
     }
   },
+  plugins: [
+    function ({ addUtilities }) {
+      addUtilities({
```
HexaField, line 90: what's the reasoning for doing it this here? by hardcoding things into our tailwind config we make it effectively impossible to customize dynamically. we need to do this with CSS injection or some other runtime method.

HexaField (changes requested):

tailwind.config.js
```diff
@@ -74,9 +74,44 @@ module.exports = {
       },
       colors: {
         'blue-primary': '#375DAF'
+      },
+      scrollbar: {
+        'transparent-track': {
```
HexaField, line 79: we need to use CSS variables here to ensure things are dynamically themeable

HexaField (changes requested):

tailwind.config.js
SamMazerIR: happy to change back to my first iteration, i had a CSS file in the ecommerce project I was importing only into the component I wanted. Is that preferable? I honestly don't know much about our tailwind/css practices, only the very basics of CSS. This was what I got when I asked chatGPT "how to apply this style using tailwind" lol

PR #10395, opened 2024-06-15
App core tech debt componentization

HexaField:
## Summary

- webapp injection into component
- auth login hook
- notification to components
- simplify engine loader lazy load
- refactor tailwind theme provider into hook

## References
closes #_insert number here_


packages/client-core/src/common/services/NotificationService.tsx
```diff
@@ -23,13 +23,15 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
+import React, { CSSProperties, Fragment, useEffect, useRef } from 'react'
 
 import multiLogger from '@etherealengine/common/src/logger'
 import { AudioEffectPlayer } from '@etherealengine/engine/src/audio/systems/MediaSystem'
-import { defineState, getState } from '@etherealengine/hyperflux'
+import { defineState, getState, useMutableState } from '@etherealengine/hyperflux'
 
-import { defaultAction } from '../components/NotificationActions'
+import Icon from '@etherealengine/ui/src/primitives/mui/Icon'
+import IconButton from '@etherealengine/ui/src/primitives/mui/IconButton'
```
aditya-mitra, line 34: we can use `HiXMark` or similar react-icon instead of using mui icon

packages/client-core/src/components/LoadWebappInjection.tsx
```diff
@@ -23,19 +23,32 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
-  const { closeSnackbar } = useSnackbar()
+  const projectComponents = useHookstate(null as null | any[])
+
+  useEffect(() => {
+    loadWebappInjection().then((result) => {
+      projectComponents.set(result)
+    })
+  }, [])
+
+  if (!projectComponents.value) return <LoadingCircle message={t('common:loader.authenticating')} />
```
aditya-mitra: we should make this `<LoadingView />`
since `LoadingCircle` is mui

packages/client/src/engine.tsx
```diff
@@ -53,12 +54,22 @@ initializeBrowser()
...
+    const redirectUrl = urlSearchParams.get('redirectUrl')
+    if (redirectUrl) {
+      history.push(redirectUrl)
+    }
+  }, [])
+
+  return (
+    <>
+      <BrowserRouter history={history}>
+        <Suspense fallback={<LoadingCircle message={t('common:loader.loadingClient')} />}>{children}</Suspense>
```
aditya-mitra: we should also add `<LoadingView />` here

aditya-mitra (changes requested):
Just a few changes and then we are good to go!
Once we remove material-ui, a lot more can be removed. I will start removing it in the next few weeks with the studio first.

packages/client-core/src/common/services/NotificationService.tsx
HexaField: i've just copied the implementation we have, i'd consider this out of scope for now

aditya-mitra (approved):
Looks good ✅

barankyle (approved):

PR #10397, opened 2024-06-17
file properties on context menu

aditya-mitra:
## References
closes part of https://tsu.atlassian.net/browse/IR-2624

[Figma Design](https://www.figma.com/design/DXwFHGa5g0peqqSjyjpxQz/Studio-Design?node-id=4452-11971&t=HMvscZdomTo77dwk-0)

## QA Steps

image


CITIZENDOT (changes requested):
- triangle count, texture count and Author properties are missing (compared to design). is this intentional?
image
- fontFamily is `Figtree` in design
- Just a minor UX issue (not much important): 
image
When clicked `Enter` after typing a tag, it would be nice if that gets added. (just like clicking `+` button)

aditya-mitra:
> triangle count, texture count and Author properties are missing (compared to design). is this intentional?

The texture and triangle count need to be present in the model (static-resource). Currently we do not get this property from the API. 
Added the author property.
image

> fontFamily is Figtree in design

added

> Just a minor UX issue (not much important):

added

HexaField:
FYI, we do have a field for triangle and texture count, the `stats` field, but we have not populated these yet. 

See https://github.com/etherealengine/etherealengine/blob/dev/packages/server-core/src/media/static-resource/static-resource-helper.ts#L191

CITIZENDOT (approved):

PR #10401, opened 2024-06-17
IR-1791: Fix Hierarchy Panel Updating on GLTF load

dinomut1:
Adds Source Component query dependency to hierarchy tree generation. This forces the hierarchy to update every time a new sourced entity is added to the scene (ie on GLTF load)

Closes https://tsu.atlassian.net/browse/IR-1791


JT00y (approved):
yes, this fix the problem.

PR #10402, opened 2024-06-17
IR-2715: Fix Blender-created model GLTF data duplication on export

dinomut1:
Fixes issue where ECS data specified in a blender file via Custom Properties gets duplicated when imported and then exported from studio. This is done by clearing the userData from meshes after they have their ecs data parsed during the import process

Closes https://tsu.atlassian.net/browse/IR-2715


JT00y (commented):
how to navigate to this bug? i use default scene platform to test export seems no error.

dinomut1:
> how to navigate to this bug? i use default scene platform to test export seems no error.

that's enough to test it. Just load any gltf file that has a collider embedded in it from blender, use model export in the studio, and then load the newly exported file back into the studio. On dev this will throw an error, on this branch it shouldn't.

JT00y (approved):

HexaField:
Screenshot from 2024-06-21 09-16-27
i'm guessing this PR broke the sky station

PR #10403, opened 2024-06-17
add commitProperty to colorInput onChange

JT00y:
## References
closes #_insert number here_


dinomut1:
I'm worried about adding snapshot generation to onChange. If someone is dragging the color picker around it could be generating a new snapshot every frame which we don't want

dinomut1 (approved):

PR #10406, opened 2024-06-18
IR-2726 Material Component Breakup

AidanCaruso:
## Summary
Breaks up the material component array into explicitly defined components

## References
closes https://tsu.atlassian.net/browse/IR-2726

## QA Steps
Switch to the material-component-breakup branch on the advanced materials repo, see also
https://github.com/theinfinitereality/eepro-advanced-materials/pull/35


dinomut1 (changes requested):
[Screencast from 06-24-2024 10:24:31 AM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/6bf2df9a-021c-412a-9ba6-a9a21622c2e6)
This video shows two bugs I'm seeing.
1. Material prototype selection doesn't differentiate between materials which are extensions of a certain material class. In this video, I'm trying to switch between the Lava material and the Standard material. The Lava material is an extension of the Standard material, and so the UI doesn't recognize that they're different, and the prototype selection is a no-op.
- This can be fixed by using the userData.type field "type":
material-types

2. Changing a material prototype only modifies one instance of the material
- looks like materials are not being shared across entities. When the prototype is changed from Lava to Basic, only one of the lava materials in the scene actually changes -- the other remains as a lava material.

dinomut1 (approved):
I'm gonna go ahead and approve this since the bugs mentioned above don't specifically pertain to the splitting up of the Material Component. However let's get those resolved in a subsequent PR ASAP

HexaField (approved):

PR #10408, opened 2024-06-18
Lit and Unlit modes implementation

SYBIOTE:
## Summary

fixes the drop shadows in the application. The changes ensure that the drop shadows are correctly applied to mesh and model and primitive geom entities.

## References

https://tsu.atlassian.net/browse/IR-2264


packages/engine/src/scene/components/ModelComponent.tsx
```diff
@@ -255,7 +256,9 @@ export const useMeshOrModel = (entity: Entity) => {
   const isModel = !!useOptionalComponent(entity, ModelComponent)
   const isChildOfModel = !!useAncestorWithComponent(entity, ModelComponent)
   const hasMesh = !!useOptionalComponent(entity, MeshComponent)
-  return isModel && !isChildOfModel && hasMesh
+  const isPrimitiveGeom = !!useOptionalComponent(entity, PrimitiveGeometryComponent)
```
HexaField: hmm, i think the original logic here is almost correct, and was changed recently here
https://github.com/EtherealEngine/etherealengine/pull/10262

I think the correct logic here should return when the entity is either:
- mesh that is not a part of a model
- a model

```ts

export const useMeshOrModel = (entity: Entity) => {
  const isModel = !!useOptionalComponent(entity, ModelComponent)
  const isChildOfModel = !!useAncestorWithComponent(entity, ModelComponent)
  const hasMesh = !!useOptionalComponent(entity, MeshComponent)
  return isModel || (!isChildOfModel && hasMesh)
}
```

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -429,9 +432,8 @@ const updateDropShadowTransforms = () => {
     const sizeBias = 0.3
     const finalRadius = sizeBias * dropShadow.radius + dropShadow.radius * centerCorrectedDist * 0.5
 
-    const shadowMaterial = (getComponent(dropShadow.entity, GroupComponent)[0] as any).material as Material
-    shadowMaterial.opacity = Math.min(1 / (1 + centerCorrectedDist), 1) * 0.6
-
+    const shadowMaterial = (getComponent(dropShadow.entity, GroupComponent)[0] as Mesh).material as Material
+    shadowMaterial.opacity = Math.min(1 / (1 + centerCorrectedDist), 1) * 2
```
HexaField: i wonder why the size here needs to change, my guess is something with our bounding boxes or similar is affecting this. this value was tweaked to look good previously.

HexaField (changes requested):

packages/engine/src/scene/components/ModelComponent.tsx
SYBIOTE: primitive geometry is also a type of model, do we not need to check for that?

packages/engine/src/scene/systems/ShadowSystem.tsx
SYBIOTE: yes, by the old value the final opacity made the drop shadows almost invisible, this gives a darker shadow

SYBIOTE:
for the lit - unlit implementation, I hide all light in unlit mode and turn on the lights when in any other mode 
in unlit mode we add a temp ambient light to the scene and remove it when we are in any other mode

packages/engine/src/scene/components/ModelComponent.tsx
HexaField: That should be covered by the condition of `!model && mesh`, which is the point of this complex hook

packages/spatial/src/renderer/functions/changeRenderMode.ts
```diff
@@ -60,24 +64,20 @@ export function changeRenderMode() {
...
-      break
     case RenderModes.LIT:
-      renderPass.overrideMaterial = null!
-      break
     case RenderModes.SHADOW:
+      iterateEntityNode(getState(EditorState).rootEntity, (entity, index) => {
+        const groups = getComponent(entity, GroupComponent)
+        for (const group of groups) {
+          if ((group as any).isLight && !(group as any).isAmbientLight) {
+            setVisibleComponent(group.entity, renderMode !== RenderModes.UNLIT)
```
HexaField: rather than doing this here, we have two other options:
- have each light component individually react to the renderer state
- create a system just before the render step that overrides the visible component to disable the lights based on renderer state

I'd preferrably opt for the second option.

HexaField (changes requested):

packages/spatial/src/renderer/functions/changeRenderMode.ts
SYBIOTE: here, we have put the logic in the webgl render system reactor, is there an inherent benefit to making it a separate system?

packages/spatial/src/renderer/WebGLRendererSystem.tsx
```diff
@@ -314,6 +316,15 @@ const execute = () => {
     _scene.background =
       sessionMode === 'immersive-ar' ? null : renderMode === RenderModes.WIREFRAME ? new Color(0xffffff) : background
 
+    if (renderMode === RenderModes.UNLIT) {
```
HexaField: let's move this to it's own system inserted `{ before: WebGLRendererSystem }`, and use ECS queries to disable each light.

HexaField (changes requested):

packages/engine/src/scene/components/ModelComponent.tsx
HexaField: @SYBIOTE can we try make this work without checking for the primitive geom component here

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -271,9 +272,12 @@ const EntityChildCSMReactor = (props: { rendererEntity: Entity }) => {
 function _CSMReactor() {
   const rendererEntity = useEntityContext()
   const renderSettingsEntity = useChildWithComponent(rendererEntity, RenderSettingsComponent)
+  const isEditor = useHookstate(getMutableState(EngineState).isEditor).value
+  const renderMode = useHookstate(getMutableState(RendererState).renderMode).value
 
   if (!rendererEntity) return null
   if (!renderSettingsEntity) return null
+  if (isEditor && ['Unlit', 'Lit'].includes(renderMode)) return null
```
HexaField: Let's use const/enum values for the strings here

HexaField (changes requested):

packages/engine/src/scene/components/ModelComponent.tsx
SYBIOTE: We certainly can But do we not want drop shadows for the primitive geom component?

SYBIOTE: > That should be covered by the condition of `!model && mesh`, which is the point of this complex hook

I don't think I remember it covering primitive geom, ill revert it and test it

packages/spatial/src/renderer/ViewportLightingSystem.tsx
```diff
@@ -0,0 +1,70 @@
...
+        child.visible = !child.isLight
+      })
+    }
+  }
+}
+
+const reactor = () => {
+  const renderer = useMutableState(RendererState)
+  useEffect(() => {
+    const root = getState(EditorState).rootEntity
```
HexaField: we cannot import the editor into the spatial package

packages/editor/src/components/properties/AmbientLightNodeEditor.tsx
```diff
@@ -28,7 +28,7 @@ import React from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { useComponent } from '@etherealengine/ecs/src/ComponentFunctions'
-import { AmbientLightComponent } from '@etherealengine/spatial/src/renderer/components/AmbientLightComponent'
+import { AmbientLightComponent } from '@etherealengine/spatial/src/renderer/components/lights/AmbientLightComponent'
```
HexaField, line 31: Let's keep the components where they were

HexaField (changes requested):

packages/editor/src/components/properties/AmbientLightNodeEditor.tsx
SYBIOTE: They are in the same place but it's easier to find the lights now, is there a reason why we don't separate out the lights from the rest of the renderer?

packages/spatial/src/renderer/ViewportLightingSystem.tsx
SYBIOTE: Ok, we can replace with origin entity

HexaField: Yeah that's fine for now

packages/editor/src/components/properties/AmbientLightNodeEditor.tsx
HexaField: I suppose it's fine. Just wanted to keep the folder structure simple.

packages/spatial/src/renderer/ViewportLightingSystem.tsx
SYBIOTE: done

packages/engine/src/scene/components/PrimitiveGeometryComponent.ts
```diff
@@ -73,33 +71,18 @@ export const PrimitiveGeometryComponent = defineComponent({
...
-  [GeometryTypeEnum.CircleGeometry]: Shapes.Mesh,
-  [GeometryTypeEnum.RingGeometry]: Shapes.Mesh,
-  [GeometryTypeEnum.TorusGeometry]: Shapes.Mesh,
-  [GeometryTypeEnum.DodecahedronGeometry]: Shapes.Mesh,
-  [GeometryTypeEnum.IcosahedronGeometry]: Shapes.Mesh,
-  [GeometryTypeEnum.OctahedronGeometry]: Shapes.Mesh,
-  [GeometryTypeEnum.TetrahedronGeometry]: Shapes.Mesh,
-  [GeometryTypeEnum.TorusKnotGeometry]: Shapes.Mesh
-} as Record<GeometryTypeEnum, Shape>
+export const usePrimitiveGeom = (entity: Entity) => {
```
MichaelEstes: I don't think this is necessary, a PrimitiveGeometryComponent will always have a MeshComponent on the same entity

packages/engine/src/scene/components/ModelComponent.tsx
```diff
@@ -255,7 +256,9 @@ export const useMeshOrModel = (entity: Entity) => {
...
+
+  return ((isModel && !isChildOfModel) || isPrimitiveGeom) && hasMesh
```
MichaelEstes: I don't think this logic is right. it'll return false for any entity that doesn't have both a model component and a mesh component on it. The primitive geometry check isn't necessary here either for the same reason as the other comment.

MichaelEstes (dismissed):

packages/engine/src/scene/components/PrimitiveGeometryComponent.ts
SYBIOTE: yes , but we need the check, for the drop shadows to work on primitve geom components

SYBIOTE: since the model check only works for model and mesh, while this one only works for primitive geometry component and mesh

packages/engine/src/scene/components/ModelComponent.tsx
SYBIOTE: i removed the primitive geom check

SYBIOTE: the check is specific for model and mesh

MichaelEstes: I'm not sure that a check for a mesh and model component on the same entity makes a lot of sense, every entity I've tested this with has either a mesh component or a model component and the mesh components are on children entities of the model component.

HexaField (approved):
working fantastic now

PR #10409, opened 2024-06-18
Assets must populate from all projects you have perms for

DanielBelmes:
## Summary
So we have a lot of permissions considerations. Currently in this state we have removed the need for static resource read. Then we are building queries with all projects you have perms for if you don't supply an individual project in the query.

- [x] asset panel populates only with `file` & `asset` resource types
- [x] moved thumbnails from `/thumbnails` to `/public/thumbnails` since they need to be public read access in s3/cloudfront or they'll be unreachable in deployments
- [x] any files in the `/public/thumbnails` folder or that end with '.thumbnail.jpg' (scene thumbnails) are tagged as 'thumbnail'
- [x] resources tagged with 'thumbnail' will not populate into resources.json
- [x] auto-generated thumbnails now have the 'thumbnail' type
- [x] in dev mode, all users with `static_resource:write` scope will automatically be given owner project permission for all projects

## References
https://tsu.atlassian.net/browse/IR-2616
IR-2616
## QA Steps

[IR-2616]: https://tsu.atlassian.net/browse/IR-2616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HexaField (approved):

HexaField (changes requested):
this won't work for users without `static-resource:read` perms

packages/server-core/src/hooks/resolveProjectsByPermission.ts
```diff
@@ -0,0 +1,63 @@
...
+      })) as Paginated<ProjectPermissionType>
+
+      for (const projP of data) {
+        if (!context.params.query?.$or) {
+          context.params.query.$or = []
+        }
+        const project = await context.app.service(projectPath).get(projP.projectId)
+        if (project !== undefined) context.params.query?.$or?.push({ project: project.name })
+      }
+      context.params.query?.$or?.push({ project: 'default-project' })
```
DanielBelmes: Just noting an alternative to this is rejecting the request if they have no perms. Because otherwise the user without static resource read would get all assets.

packages/ui/src/components/editor/panels/Assets/container/index.tsx
```diff
@@ -370,8 +368,6 @@ const AssetPanel = () => {
         key: {
           $like: `%${searchText.value}%`
         },
-        type: 'asset',
```
HexaField, line 376: @DanielBelmes @dinomut1 why are we removing the 'asset' type here, we decided that we want a dedicated 'asset' type (that currently auto-populates if the tags field has more than one tag, meaning it was intentially given a tag by the content team beyond the autogenerated file type tag). This allows us to filter out thumbnails, scenes, avatars, recordings etc. Otherwise all of these will populate into the asset panel. We need something that explicitly defines a file is an asset.

HexaField (changes requested):

packages/ui/src/components/editor/panels/Assets/container/index.tsx
DanielBelmes: just to keep a record here. At least for the time being I think it's best to remove this part of the query until we get all the assets tagged as asset.

dinomut1: @HexaField "Assets" don't exist right now. We want both files and assets in the assets panel. We only want to hide files when an asset exists that represents said file as a LODded version. As we don't have that implemented right now, every file should go into the assets panel

packages/server-core/src/hooks/verify-project-permission.ts
```diff
@@ -58,6 +58,8 @@ export default (types: string[]) => {
 
     if (!project) throw new NotFound('Project not found')
 
+    if (project.name === 'default-project') return context
```
DanielBelmes: also ensuring non admins will be able to access default proj so the panel works

MichaelEstes (approved):

packages/server-core/src/media/static-resource/static-resource.resolvers.ts
```diff
@@ -82,17 +82,10 @@ export const staticResourceResolver = resolve<StaticResourceType, HookContext>(
...
-        query: {
-          key: staticResource.thumbnailKey
-        }
-      })
-      if (!thumbnailStaticResource.data.length) return
       return (
         storageProvider.getCachedURL(staticResource.thumbnailKey, context.params.isInternal) +
         '?hash=' +
-        thumbnailStaticResource.data[0].hash.slice(0, 6)
+        staticResource.hash.slice(0, 6)
```
HexaField: why are we adding the hash for the static resource to the static resource's thumbnail? this will invalidate cache incorrectly for the thumbnail

packages/server-core/src/media/static-resource/static-resource.hooks.ts
```diff
@@ -165,25 +183,24 @@ export default {
         iffElse(
           checkScope('static_resource', 'read'),
```
HexaField: we need to add the admin action back such that we run the else condition here for all users, not just users who aren't admins. this will cause any find operations to give back different data for admins (specificlaly those with static_resource:read scopes)

HexaField (changes requested):

packages/server-core/src/media/static-resource/static-resource.resolvers.ts
DanielBelmes: I'm removing this change. I don't think it's related to this PR. Basically I believe the CMS branch broke this by not migrating any of the older thumbnails into static resources. I don't have a leg in the putting thumbnails into static resource conversation as I am very much new to this conversation.

packages/server-core/src/media/static-resource/static-resource.hooks.ts
DanielBelmes: So we don't want admins to be able to view everything in assets? the action is built into the query. If we ever want an admin to actually see the assets within the studio we will have to query for admin and add into the action within the editor. I was just under the assumption we want admins to see everything within the studio.

HexaField: The action is for the admin panel. The studio should act the same for everyone.

packages/server-core/src/media/static-resource/static-resource.resolvers.ts
HexaField: The resources.json is now the source of truth and needs to be updated accordingly.

packages/server-core/src/media/static-resource/static-resource.hooks.ts
DanielBelmes: Sounds good to me

dinomut1 (changes requested):
thumbnails are not static resources, this will not work

HexaField:
> thumbnails are not static resources, this will not work

Thumbnails are static resources, there is a `'thumbnail'` type for them.

packages/server-core/src/media/static-resource/static-resource.resolvers.ts
dinomut1: thumbnails are not static resources and shouldn't be in resources.json

dinomut1:
> > thumbnails are not static resources, this will not work
> 
> Thumbnails are static resources, there is a `'thumbnail'` type for them.

They shouldn't be, they're not first class files and shouldn't be visible or accessible to the user

HexaField:
> > > thumbnails are not static resources, this will not work
> > 
> > 
> > Thumbnails are static resources, there is a `'thumbnail'` type for them.
> 
> They shouldn't be, they're not first class files and shouldn't be visible or accessible to the user

They need to be. We can change the queries to return 'file' or 'asset' types for now.

dinomut1 (approved):

HexaField:
Will wait for @hanzlamateen to sign off on this

packages/server-core/src/media/static-resource/static-resource.hooks.ts
```diff
@@ -163,27 +182,26 @@ export default {
...
+            iffElse(
+              hasProjectField,
+              [
+                verifyScope('editor', 'write'),
+                resolveProjectId(),
+                verifyProjectPermission(['owner', 'editor', 'reviewer'])
+              ],
+              [verifyScope('editor', 'write'), resolveProjectsByPermission()]
+            ) as any
+          ]
```
hanzlamateen: Since `verifyScope('editor', 'write')` is checked in both cases, therefore it can be moved before iffElse.

```suggestion
          [
            verifyScope('editor', 'write'),
            iffElse(
              hasProjectField,
              [
                resolveProjectId(),
                verifyProjectPermission(['owner', 'editor', 'reviewer'])
              ],
              [resolveProjectsByPermission()]
            ) as any
          ]
```

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
```diff
@@ -197,6 +198,31 @@ async function addScopes(context: HookContext<IdentityProviderService>) {
   }
 }
 
+const addDevProjectPermissions = async (context: HookContext<IdentityProviderService>) => {
+  if (!isDev || (context.actualData as IdentityProviderType).type !== 'admin') return
```
hanzlamateen: We dont have identity provider type as 'admin', rather type provider represents the SSO provider.

https://github.com/EtherealEngine/etherealengine/blob/a1ff227aa9e602201dd76290e5708780036f507d/packages/common/src/schemas/user/identity-provider.schema.ts#L38

I dont see it being passed anywhere in our code base either:
image

packages/server-core/src/hooks/resolveProjectsByPermission.ts
```diff
@@ -0,0 +1,68 @@
...
+        throw new Forbidden(`Project permissions not found`)
+      }
+
+      for (const projP of data) {
+        if (!context.params.query?.$or) {
+          context.params.query.$or = []
+        }
+        const project = await context.app.service(projectPath).get(projP.projectId)
+        if (project !== undefined) context.params.query?.$or?.push({ project: project.name })
+      }
```
hanzlamateen: I think we should be using `query.project.$in: []` here instead of $or. https://feathersjs.com/api/databases/querying#in-nin

This is because current implementation has a loop hole if the requested query from client is:

```
query: {
  $or: [
    { key: 'abc' },
    { key: 'xyz' }
  ]
}
```

Then this hook is going to translated query to:
```
query: {
  $or: [
    { key: 'abc' },
    { key: 'xyz' },
    { project: 'proj1' },
    { project: 'proj2' }
  ]
}
```

This is not desired. Because it can return data if key `abc` exists in `proj3`. Rather, it should be:
```
query: {
  $or: [
    { key: 'abc' },
    { key: 'xyz' }
  ],
  { project: { $in: ['proj1', 'proj2'] }  }
}
```

hanzlamateen (dismissed):

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
DanielBelmes: yea we don't have admin but I wanted to surface that we do set type to 'guest'

```
await Engine.instance.api.service(identityProviderPath).create({
    type: 'guest',
```

packages/server-core/src/hooks/resolve-projects-by-permission.ts
```diff
@@ -0,0 +1,67 @@
...
+ * @returns
+ */
+export default () => {
+  return async (context: HookContext<Application>) => {
+    if (!context.params.query?.project) {
+      const loggedInUser = context.params.user as UserType
+      const { data } = (await context.app.service(projectPermissionPath).find({
+        query: {
+          userId: loggedInUser.id,
+          $limit: 1000 //idk what's a good number
```
barankyle: You can get all results by removing $limit and adding `paginate: false` next to query, i.e.

```
query: {
      userId: loggedInUser.id
},
paginate: false
```

If you look for `paginate: false` and see it inside `query` in some places, that's because we do that when requesting an un-paginated response from the client side, and there's a hook `enable-client-pagination` that moves it from `params.query` to `params`.

barankyle (approved):

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
HexaField: Weird I just copied it from another hook in user hooks

HexaField (approved):

PR #10410, opened 2024-06-18
Improving DUX more

DanielBelmes:
## Summary
From my research we were actually just missing an event on our branch-build that prevented merge queue from working.

## References
closes #_insert number here_


DanielBelmes:
Still thinking about restricting by file type. I keep considering edge cases that start to include .css files and files I am trying to avoid tests from running. I think the issue is that we don't have a separate action for checking if the application builds. Vs our unit tests. So I might extract tests out. I am not sure.

DanielBelmes:
I have a plan to better cache the modules to work across workflows hopefully. Then I think I will have a specific unit tests workflow. Separate from ensuring build client works which will always run.

DanielBelmes:
Scratching plan and just fixing merge queue

HexaField (approved):

PR #10411, opened 2024-06-19
Append hash to bust cache on all static resources

HexaField:
## References
closes #_insert number here_


barankyle (changes requested):
There's some bugs around saving scenes. If you change a scene and then save it, the hash is getting saved onto the scene key in the DB, which we don't want. If you change the scene and save it again, things get weirder, with the scene now having a key of e.g. `projects/default-project/public/scenes/defaulthash91b94e.gltf?hash=5129ca` and a new GLTF file, of type 'file', with a key like `projects/default-project/public/scenes/defaulthash91b94e.gltf`. Hopefully fixing the former prevents the latter from happening.

barankyle (approved):

MichaelEstes (approved):

PR #10423, opened 2024-06-20
IR-1579-Resource-tracker

MichaelEstes:
## Summary
- Cleans up cloned geometries from gizmo
## References
closes #_insert number here_


SYBIOTE (approved):
gizmo changes look good and we return the scene id with load location

packages/client-core/src/components/World/LoadLocationScene.tsx
```diff
@@ -88,10 +88,15 @@ export const useLoadLocation = (props: { locationName: string }) => {
 export const useLoadScene = (props: { projectName: string; sceneName: string }) => {
   const sceneKey = `projects/${props.projectName}/${props.sceneName}`
   const assetID = useFind(staticResourcePath, { query: { key: sceneKey, type: 'scene' } })
+  const sceneID = useHookstate<string | undefined>(undefined)
```
HexaField: what's the reason for adding this?

MichaelEstes: The CMS changes broke the scene loading in the dev test suite, it could hook into the LocationState and get the id that way, but it feels very unintuitive to have loading a scene be a side effect

HexaField: hmm we should talk about this. i think i fixed it in a recent PR

PR #10425, opened 2024-06-20
Remove react-is hackfix install

barankyle:
## Summary

Recent updates to typescript and eslint versions made the hackfix install of react-is in building the client error out, and seems to no longer be necessary to get it to install properly.

Moved @emotion/react and @emotion/styled to main dependencies from devDependencies since the client build was throwing errors about them being missing.

## References
closes #_insert number here_

### DIFF
--- a/dockerfiles/api/Dockerfile-api-client
+++ b/dockerfiles/api/Dockerfile-api-client
@@ -40,8 +40,6 @@ COPY patches/ ./patches/
 
 ARG NODE_ENV
 RUN npm install --loglevel notice --legacy-peer-deps --production
-# Hackfix for issue with broken installation of react-is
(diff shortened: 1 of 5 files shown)


HexaField (approved):

HexaField (changes requested):
`@emotion` packages dont seem to be used in the core engine, so we can probably remove them

barankyle:
> `@emotion` packages dont seem to be used in the core engine, so we can probably remove them

They seem to be necessary to build the client files - build-client was erroring without them, as something in one of our dependencies (don't remember which) was erroring about one of those packages being missing.

HexaField (approved):
weird, ok

PR #10427, opened 2024-06-20
only check scene name on asset type scene

DanielBelmes:
## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/media/file-browser/file-browser.class.ts
+++ b/packages/server-core/src/media/file-browser/file-browser.class.ts
@@ -264,7 +264,7 @@ export class FileBrowserService
       }
     }
 
-    if (data.type !== 'scene') validateSceneName(data.path)
+    if (data.type === 'scene') validateSceneName(data.path)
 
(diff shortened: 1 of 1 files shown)


packages/server-core/src/media/file-browser/file-browser.class.ts
```diff
@@ -264,7 +264,7 @@ export class FileBrowserService
       }
     }
 
-    validateSceneName(data.path)
+    if (data.type === 'scene') validateSceneName(data.path)
```
HexaField, line 267: we need to validate all files. which files are being found as illegal?

HexaField (changes requested):

packages/server-core/src/media/file-browser/file-browser.class.ts
DanielBelmes: Thumbnails being uploaded. Underscores are allowed in files but not in our scenes names.

HexaField (approved):
on further inspection this is a good quick fix and we shold address this properly another time

PR #10436, opened 2024-06-21
studio: tailwind color palette

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-2728

* add iR color palette to tailwind config with different shades
* update current dark mode colors to use tailwind values

## References
closes #_insert number here_


tailwind.config.js
```diff
@@ -73,7 +73,92 @@ module.exports = {
         }
       },
       colors: {
-        'blue-primary': '#375DAF'
+        // BRAND COLORS
+        'magenta': {
```
HexaField, line 77: why are we adding specific colours? these should all be categorical as part of the theme, and parsed as css variables

HexaField (changes requested):

tailwind.config.js
achen5671: these colors are the palette specified in our figma designs. We are not using all of them. I added it here in case for the future when we use them. I can provide the figma file if you're interested. We also can add this in the future when we do need them. Let me know what you think 🙂

HexaField: it's fine to add new colors, but they need to be configurable via our theme abstraction via CSS variables as the other ones are, and ideally as few as are necessary

DanielBelmes: I believe what we want as an alternative is to not use tailwind color styling but to instead use our original color variables with tailwind like so `bg-[--color-primary]` see packages/client/src/pages/styles.scss for the vars

HexaField: No those are being deprecated, as they're only used for the location route. We are starting fresh with tailwind.

DanielBelmes: sorry I missed this. Tailwind doesn't support theming in the way we do it. We do have to continue using css color vars. We can give tailwind config our vars and modify the vars realtime I believe. That way we can use things like bg-blue.

HexaField:
We can try this again on a fresh branch as this one has become too stale

PR #10444, opened 2024-06-24
IR-2719: Unify & unit-test all regex patterns at one place

CITIZENDOT:
## Summary
- Move all regex patterns at one place
- Reduce usage of `new RegExp` as much as possible, since they create an instance.
  - Use simple string operations wherever possible instead of Regex
  - Prefer static regex literal instead of new `RegExp` instance
  - Remove unused `RegExp`s
- Unit tests for the patterns

## References
closes #_insert number here_


HexaField (dismissed):
The issue doesn't specify to move all regexes to the same file, and I don't think it is a good architectural decision to do so either. We defnitely want them tested, but the regexes should exist in the files or modules that use them, which goes along with our principles of modularity, composability and keeping concepts decentralized & hierarchically composed.

CITIZENDOT:
@HexaField, regex patterns are across multiple packages and some are repeated. (`EMAIL_REGEX` is declared twice with different definitions).

So, if we store all the regexes at one place, we don't have to worry about repetitions and we can test all of them in the same package.

For eg: If I am searching for `EMAILREGEX`, and couldn't find it (because `EMAIL_REGEX` is defined, but `EMAILREGEX` isn't); I might define it somewhere and use it. But if there's a single source of all regexes, I'll just go through that file and use the appropriate one.

hanzlamateen:
> The issue doesn't specify to move all regexes to the same file, and I don't think it is a good architectural decision to do so either. We defnitely want them tested, but the regexes should exist in the files or modules that use them, which goes along with our principles of modularity, composability and keeping concepts decentralized & hierarchically composed.

I think having them in a single place is good atleast for the ones that are too common. Like email, project name validation, url validation, file validation, etc. This makes it easy to use and find them.

Though the ones that are not common can be part of their respective files.

Like project name validation is required in core engine as well as MT so its good to have such regex in common place.

packages/common/src/regex/index.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Matches CSS imports & URLS.
+ * For eg: `@import "styles.css"`, `url(image.png)`. Captures the resource in group 2 or group 3.
+ */
+export const CSS_URL_REGEX = /(@import\s+["']([^"']+)["']|url\((?!['"]?(?:data):)['"]?([^'"\)]+)['"]?\))/gi
+
+/**
+ * Matches absolute URLs. For eg: `http://example.com`, `https://example.com`, `ftp://example.com`, `//example.com`, etc.
+ * This Does NOT match relative URLs like `example.com`
+ */
+export const ABSOLUTE_URL_REGEX = /^(?:[a-z][a-z0-9+.-]*:|\/\/)/
```
aditya-mitra: this seems to match only the protocol part of the URL (`http://abc` also matches). if this happens to be so, we can rename this variable

packages/common/src/regex/index.ts
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * Matches `https://`
+ */
+export const HTTPS_REGEX = /https:\/\//
```
aditya-mitra: This can be a simple string as well.

aditya-mitra (dismissed):
Awesome refactor. `regex/index.ts` is very well written

packages/common/src/regex/index.ts
CITIZENDOT: but then we can't use `HTTPS_REGEX.test()` function, like here: https://github.com/EtherealEngine/etherealengine/blob/5bde350dd6e733d6da0d5faca51128975268f0d5/packages/server-core/src/setting/client-setting/client-setting.hooks.ts#L76-L81

however, we can remove this pattern altogether and use `startsWith("https://")` there. Shall I do that instead?

packages/common/src/regex/index.ts
CITIZENDOT: Done!

packages/common/src/regex/index.ts
aditya-mitra: Yes that would be better. You can even use `includes` if you want to search in the string globally.

CITIZENDOT: Done

hanzlamateen (approved):

hanzlamateen (changes requested):
github actions are failing

CITIZENDOT:
Tests for **test_docker (server-core) (pull_request)** are failing due to:
```
Error: listen EADDRINUSE: address already in use 0.0.0.0:8642
```
But It seems unrelated?

image

CITIZENDOT:
> github actions are failing

I think I found the cause. `GITHUB_URL_REGEX` captures the group 1 instead of 2. I'll modify tests for patterns which are capturing groups.

Edit: Done

hanzlamateen (approved):

PR #10447, opened 2024-06-25
hot-fix upload assets button disabled

aditya-mitra:
## References
closes #_insert number here_


aditya-mitra:
closing as it is fixed in current dev

PR #10448, opened 2024-06-25
fix drag and drop on file browser

aditya-mitra:
## Summary

the starting `/` was producing error on the [file-browser patch if condition](https://github.com/EtherealEngine/etherealengine/blob/2d3027a2d1af0eaa3c90583d4528766e1f8af1ff/packages/server-core/src/media/file-browser/file-browser.class.ts/#L253)

## References
closes #_insert number here_


SYBIOTE (approved):
this was all that caused the issue? 
I'm assuming when dragging and dropping we got an error path so drag and drop was failing this solves it?

aditya-mitra:
yes

PR #10451, opened 2024-06-25
IR 2768 Set z offset to zero when exiting first person mode

AidanCaruso:
## Summary
Fixes camera offset never being set back to zero after exiting first person mode

## References
closes #_insert number here_


HexaField:
Probably can be closed or needs to be tested again after conflict is resolved with new follow camera refactor

PR #10452, opened 2024-06-25
IR-2727: Zendesk authentication

jose-galvan:
## Summary

This PR introduces Zendesk authentication, enabling a more personalized and seamless support experience. Key features include:

- **User Identification for Support**: Allows customer support to identify and authenticate end users reaching out for support.
- **Cross-Device Chat Access**: Enables end users to access their chat history and ongoing conversations across multiple devices.
- **Improved User Experience**: Provides a consistent and secure authentication process for all users.

See: [Authenticating end users for messaging](https://support.zendesk.com/hc/en-us/articles/4411666638746-Authenticating-end-users-for-messaging)

### Admin panel setting 
image

## References
closes #_insert number here_

## QA Steps

Pre-requisite: run the migrations and then in the admin panel set the Zendesk key.

1. Open the app in a new tab
    - If the user is logged out, the chat widget should not be displayed.
2. Log in
    - Authenticate using valid credentials.
3. Start a chat with customer support
    - Initiate a conversation using the chat widget.
4. Log out
    - Ensure the chat session is cleaned up.
    - Verify that the chat widget is hidden.
5. Log in again
    - Authenticate using the same user credentials.
6. Check the chat
    - Confirm that previous conversations with support are visible.


packages/server-core/src/integrations/zendesk/zendesk.class.ts
```diff
@@ -0,0 +1,53 @@
...
+ * A class for Signing zendesk tokens service
+ */
+export class ZendeskAuthenticationService implements ServiceInterface<string, ZendeskAuthenticationType> {
+  app: Application
+
+  constructor(app: Application) {
+    this.app = app
+  }
+
+  async create(data: ZendeskAuthenticationType): Promise<string> {
```
hanzlamateen: You can do this in a hook.

`packages/projects/projects/eepro-multitenancy/services/entri/entri-auth` is a similar example of who its uses hooks.

The .class file should inherit from `BaseService` like in `entri-auth.class.ts`

hanzlamateen (changes requested):

.env.local.default
```diff
@@ -32,6 +32,7 @@ VITE_APP_HOST=localhost
 VITE_APP_PORT=3000
 VITE_ZENDESK_ENABLED=false
 VITE_ZENDESK_KEY=
+VITE_ZENDESK_AUTHENTICATION_ENABLED=false
```
hanzlamateen, line 35: When will a user need to use `VITE_ZENDESK_AUTHENTICATION_ENABLED`? 

What if, `VITE_ZENDESK_ENABLED` is `true` and `VITE_ZENDESK_AUTHENTICATION_ENABLED` is `false`. How will it work?

hanzlamateen (changes requested):

.env.local.default
jose-galvan: So authentication is not required to use Zendesk, if it's not enabled users can still contact support but they have to provide their name and email, with Zendesk authentication enabled end users  can skip that step since the information is already stored in the token used to authenticate them. 

An added benefit of Zendesk is that they can view their conversation on multiple devices or even after cleaning the cache in their devices.

packages/server-core/src/integrations/zendesk/zendesk.class.ts
jose-galvan: Done! Moved logic from service to hooks

HexaField:
Which routes need zendesk? We probably want to use the `useZendesk` hook on as as-needs basis, rather than on every route. This is especially important for reducing the TTL for auth routes.

packages/common/src/schemas/integrations/zendesk/zendesk.schema.ts
```diff
@@ -0,0 +1,48 @@
...
+
+import { Static, Type, querySyntax } from '@feathersjs/typebox'
+
+export const zendeskAuthenticationPath = 'zendesk-authentication'
+
+export const zendeskAuthenticationMethods = ['create'] as const
+
+export const zendeskAuthenticationSchema = Type.Object({
+  scope: Type.String(),
+  external_id: Type.Optional(Type.String()),
```
hanzlamateen: We use camelCase Notation. Update to `externalId`

hanzlamateen (changes requested):
Save in admin panel does not seems to be working

https://github.com/EtherealEngine/etherealengine/assets/10975502/69279011-ca07-473d-b628-040c2e2805ac

packages/common/src/schemas/integrations/zendesk/zendesk.schema.ts
jose-galvan: No, zendesk uses snake case, if I use camelCase they won't find the user id in the token.
image

jose-galvan:
> Which routes need zendesk? We probably want to use the `useZendesk` hook on as as-needs basis, rather than on every route. This is especially important for reducing the TTL for auth routes.

@HexaField the requirement was to always display the widget as long as the user is authenticated. I'm talking with UI/UX to see if we change the way users interact with the widget. i.e for the click on a link in profile settings. 

The Zendesk script is installed just once after the user has logged in, and the Zendesk authentication process only starts when a chat is initiated.

packages/common/src/schemas/integrations/zendesk/zendesk.schema.ts
hanzlamateen: @jose-galvan so, I made changes in schema. Please have a look of following commit

https://github.com/EtherealEngine/etherealengine/pull/10452/commits/56a71f47c4e44ea86d322774f9bcc48b19bbffd4

For your reference:

The Type schema i.e. IdentityProviderType or ZendeskAuthenticationType is used to represent the return object type of service. Now since for this service as the return type from sign() method was a string. It seems like we dont need ZendeskAuthenticationType all together. There for I have remove it and uploaded zendesk.class with string as this is what this service returns.

hanzlamateen:
@jose-galvan also we follow this convention to have service name and its file similar to path variable

Made changes in following commit to make them sync with file names and service name.

https://github.com/EtherealEngine/etherealengine/pull/10452/commits/afc33828057990dd6cd7c365709c87a287f72b9a

hanzlamateen (approved):

jose-galvan:
@HexaField Can we merge this PR?

HexaField (approved):

HexaField:
yep looks good

PR #10453, opened 2024-06-26
fix(IR-2799): fix feature flags state

kattsushi:
## References
closes #_insert number here_


packages/common/src/schemas/setting/feature-flag-setting.schema.ts
```diff
@@ -44,7 +44,10 @@ export const featureFlagSettingSchema = Type.Object(
       'ir.client.menu.social',
       'ir.client.menu.emote',
       'ir.client.menu.avaturn',
-      'ir.client.menu.readyPlayerMe'
+      'ir.client.menu.readyPlayerMe',
+      'ir.mt.menu.myDomains',
```
HexaField: We should make this just an untyped string. We don't want MT concepts in the core engine.

packages/engine/src/FeatureFlagsState.tsx
```diff
@@ -37,7 +37,7 @@ export const FeatureFlagsState = defineState({
   },
   useEnabled(flagName: FeatureFlag) {
     const state = useHookstate(getMutableState(FeatureFlagsState)[flagName]).value
-    return typeof state === 'boolean' ? state : true
+    return typeof state === 'boolean' ? state : typeof state === 'number' && state === 0 ? false : true
```
HexaField: When would this equal a number?

HexaField (changes requested):

packages/engine/src/FeatureFlagsState.tsx
kattsushi: Although at the schema level it is of type boolean, in the database (MariaDB) it is represented as 0 and 1 instead of false and true respectively, which is what the `state` value number is returning. I don't see any unit test for this, and that's what I found when I was debugging.

packages/common/src/schemas/setting/feature-flag-setting.schema.ts
kattsushi: not sure about this one,  also the FeatureFlagsState lives in the core engine so how the projects can use that feature knowing the feature flag name ? thoughts  @hanzlamateen @MoizAdnan  ? 

other way would be for example `ir-client.menu.myDomains` but i was assuming that we will have a Admin dashboard for features flags for each project and the way to identified is by the structure of the feature flag name like `ir.${project}.flag`

packages/engine/src/FeatureFlagsState.tsx
kattsushi: Screenshot 2024-06-26 at 6 52 52 a m
i tested out modifying directly in the database and is working considering that, because in other hand always return true because the type of `state` isn't boolean

packages/common/src/schemas/setting/feature-flag-setting.schema.ts
kattsushi: i already untyped the flagName to decouple flag names from the core engine, just leaving a generic in the actions of the state to remain type safety in the project context

packages/engine/src/FeatureFlagsState.tsx
hanzlamateen: @kattsushi I think you need a resolver for flagValue like https://github.com/EtherealEngine/etherealengine/blob/4ea6942fc0f39af7dbeacd195942a0b9ca1deaa8/packages/server-core/src/social/location/location.resolvers.ts#L80

hanzlamateen (changes requested):

packages/engine/src/FeatureFlagsState.tsx
kattsushi: thanks @hanzlamateen fixed

hanzlamateen (approved):

packages/engine/src/FeatureFlagsState.tsx
```diff
@@ -23,19 +23,19 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
...
 import { defineState, getMutableState, useHookstate } from '@etherealengine/hyperflux/functions/StateFunctions'
 import { useFind } from '@etherealengine/spatial/src/common/functions/FeathersHooks'
 import { useEffect } from 'react'
 
 export const FeatureFlagsState = defineState({
   name: 'ee.engine.FeatureFlagsState',
-  initial: {} as Record<FeatureFlag, boolean>,
-  enabled(flagName: FeatureFlag) {
+  initial: {} as Record<string, boolean>,
+  enabled<FlagName extends string>(flagName: FlagName) {
```
HexaField: Let's keep the types simple here, string is more than enough. This doesn't really add any extra typechecking or safety.

HexaField (changes requested):

hanzlamateen:
image


Added ability to toggle feature flags from settings in admin panel.

hanzlamateen (approved):

packages/spatial/src/common/functions/string.ts
```diff
@@ -48,3 +48,24 @@ export function insertAfterString(source: string, searchTerm: string, addition:
...
   )
 }
+
+/**
+ * Method used to get all leaf node strings from an object.
+ * https://stackoverflow.com/a/63100031/2077741
+ * @param arg
+ * @returns
+ */
+export function getAllStrings(arg: any) {
```
HexaField: let's move this to the common package

HexaField (changes requested):

packages/spatial/src/common/functions/string.ts
hanzlamateen: Move that whole `string.ts` file to common package? or only the method that I added?

hanzlamateen: Done

PR #10455, opened 2024-06-26
studio: adding a second URL path breaks the video component

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-2707

## References
closes #_insert number here_


packages/editor/src/components/properties/Util.ts
```diff
@@ -64,9 +64,13 @@ export const updateProperties = <C extends Component>(
     : editorState.lockPropertiesPanel.value
     ? [UUIDComponent.getEntityByUUID(editorState.lockPropertiesPanel.value)]
     : SelectionState.getSelectedEntities()
-  for (let i = 0; i < affectedNodes.length; i++) {
-    const node = affectedNodes[i]
-    updateComponent(node, component, properties)
+
+  for (const node of affectedNodes) {
+    try {
+      updateComponent(node, component, properties)
```
HexaField, line 70: what is the underlying issue? we should to solve it in the video component and allow it to fail gracefully if possible, rather than just try catching

HexaField (changes requested):

packages/editor/src/components/properties/Util.ts
achen5671: From what I observed, we dont have any error handling within the `updateComponent` function. In the context of this PR, when I add a URL path that is invalid i.e empty or some random text, `updateComponent` will fail which causes the video to stop.

(For some reason I just saw this comment, I apologize for the delay response)

achen5671: We could throw an error and display it on the screen but I think that's more of a design question. Im open for feedback.

HexaField: error handling should happen in the component itself, and we should surface those to the UI of the properties editor. Many components leverage the `ErrorComponent` via `errorFunctions` for this

PR #10460, opened 2024-06-26
change grid defaults

DanielBelmes:
## Summary
grid was really bright and the distance was too big to get aliased and look nice. I also modified the area of no draw around the axis lines. 

refactored slightly for readability for the distance fade

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/renderer/components/InfiniteGridHelper.ts
+++ b/packages/spatial/src/renderer/components/InfiniteGridHelper.ts
@@ -94,13 +94,13 @@ float getGrid(float size) {
 }
 
 float getXAxisLine() {
-  float lineWidth = 0.02; // Adjust line width if needed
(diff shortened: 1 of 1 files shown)


SYBIOTE (approved):
Looks good, 
Why is the distance 200 though 
Do we want to grid to be centered around the origin and not spread across infinitely?

DanielBelmes:
> Looks good, Why is the distance 200 though Do we want to grid to be centered around the origin and not spread across infinitely?

It just looks better. it could be bigger but then we get a lot of aliasing in the distance

HexaField:
> > Looks good, Why is the distance 200 though Do we want to grid to be centered around the origin and not spread across infinitely?
> 
> It just looks better. it could be bigger but then we get a lot of aliasing in the distance

We might want to scale it with something like proximity to camera or have further out lines be smaller, but this is good for now.

PR #10461, opened 2024-06-26
Allow the compression or multiple selected images/models and the modification of properties of multiple files

Rezmason:
## Summary
Currently, if the user selects one file in the file browser, they can compress it (if it is an image or model) or modify its properties.

The work in this PR allows the same UI to be used to compress or modify properties of _multiple_ selected files in the file browser at once.

## References

https://tsu.atlassian.net/browse/IR-2340

### DIFF
--- a/packages/ui/src/components/editor/panels/Files/browserGrid/index.tsx
+++ b/packages/ui/src/components/editor/panels/Files/browserGrid/index.tsx
@@ -187,7 +187,7 @@ export const FileGridItem: React.FC<FileGridItemProps> = (props) => {
       </div>
 
       <Tooltip title={t(props.item.fullName)}>
-        <div className="text-secondary text-wrap line-clamp-1 w-full break-all text-sm">{props.item.fullName}</div>
+        <div className="text-secondary line-clamp-1 w-full text-wrap break-all text-sm">{props.item.fullName}</div>
       </Tooltip>
     </div>
   )
@@ -357,7 +357,7 @@ export function FileBrowserItem({
       )}
 
       <ContextMenu anchorEvent={anchorEvent} onClose={() => setAnchorEvent(undefined)}>
-        <div className="min-w-44 flex w-fit flex-col gap-1 truncate rounded-lg bg-neutral-900 shadow-lg">
+        <div className="flex w-fit min-w-44 flex-col gap-1 truncate rounded-lg bg-neutral-900 shadow-lg">
           <Button variant="outline" size="small" fullWidth onClick={addFolder}>
             {t('editor:layout.filebrowser.addNewFolder')}
           </Button>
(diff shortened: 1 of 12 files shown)


packages/ui/src/components/editor/panels/Files/browserGrid/index.tsx
```diff
@@ -416,21 +411,31 @@ export function FileBrowserItem({
...
+          <Button
+            variant="outline"
+            size="small"
+            fullWidth
+            // disabled={!fileConsistsOfContentType(item, 'image')} // TODO: move context menu to its own component, with a State<Filetype[]> -JS
+            onClick={(e) => setOpenImageCompress(true)}
+          >
+            {t('editor:layout.filebrowser.compress')}
+          </Button>
+        )}
```
aditya-mitra: we can make both of these shorter by combining them into one button and using a ternary operator for the onClick
```tsx
onClick={()=> {
          if (fileConsistsOfContentType(item, 'model')) {
              PopoverState.showPopupover(
                <ModelCompressionPanel selectedFile={item as FileType} refreshDirectory={refreshDirectory} />
              )
            } else if (fileConsistsOfContentType(item, 'image')) {
              PopoverState.showPopupover(
                <ImageCompressionPanel selectedFile={item as FileType} refreshDirectory={refreshDirectory} />
              )
            }
```

also we should make the button disabled instead of conditional rendering

aditya-mitra (changes requested):
This looks good currently

Once #10397 and #10498 get merged, we will need to refactor FileProperties and ImageCompressionPanel

While opening the model compression menu, the 3 default LOD levels which were previously present have become absent
image

when trying to add an LOD, we are getting thrown with an error
image

packages/ui/src/components/editor/panels/Files/browserGrid/index.tsx
Rezmason: Agreed; my change here is an intentional stopgap that allows the context menus to surface the proper compression panel (image vs model) without requiring knowledge of every selected file.

Once the context menu is moved to the container tsx, we'll be able to modify its state to be an array of files instead of just one item, and the previous code can triumphantly return 🌟

dinomut1 (changes requested):
[Screencast from 07-03-2024 12:55:54 PM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/f63d67b0-db8f-425b-b0e6-a68a4f3c64a0)
Running into an issue where when multiple files are selected, the properties panel doesn't show up until i click on an individual file afterward.

packages/ui/src/components/editor/panels/Files/browserGrid/index.tsx
aditya-mitra: new compression and files properties modals now in file browser container!

Rezmason:
Fixed several bugs including the ones mentioned here.

dinomut1 (changes requested):
[Screencast from 07-10-2024 01:58:05 PM.webm](https://github.com/EtherealEngine/etherealengine/assets/94419856/a8af476d-ee27-47b8-99ac-69b4838ea102)
Seeing issue with adding tags

dinomut1 (approved):
Terrific. We'll optimize the one feathers call per selected file later

aditya-mitra (approved):

PR #10462, opened 2024-06-26
IR-2824-Reinstate-original-publish-workflow-in-studio

HexaField:
## References
closes #_insert number here_


aditya-mitra (approved):
Works good!
(at later point, we can replace the AddEditLocationModal with the before planned modal)

PR #10463, opened 2024-06-26
Fix fog color and improve colorInput

DanielBelmes:
## Summary
Okay just a small fix for fog. But I also wanted to bring up the difficult of hovering over the color swatch and then going to edit the color and it unhovers. So I modified to to work only on focus and to have a pointer on hover. I think the UX is much better this way.

focus-within property is needed so that when you click on the swatches or inputs in the color component that it stays open.

## References
closes #_insert number here_


SYBIOTE (approved):
Looks good

PR #10465, opened 2024-06-27
Optimize-routing

HexaField:
## References
closes #_insert number here_


HexaField:
Worth deploying this to QAT and ensuring auth login still works

packages/client-core/src/admin/components/settings/tabs/client.tsx
```diff
@@ -130,24 +104,34 @@ const ClientTab = forwardRef(({ open }: { open: boolean }, ref: React.MutableRef
...
         state.set({ loading: false, errorMessage: e.message })
       })
   }
 
   const handleCancel = () => {
-    settings.set(clientSetting)
+    settingsState.set(clientSettings)
   }
 
+  console.log(settingsState.value)
```
aditya-mitra: we can remove this log statement

aditya-mitra (approved):
Looks good!

packages/client-core/src/admin/components/settings/tabs/client.tsx
HexaField: done

PR #10467, opened 2024-06-27
Optimize engine import

HexaField:
## Summary

This leans out the initial load time, especially for auth routes, which now do not load any API requests or the full 3D engine and works something like 20 times faster locally.

The only regression here is UI sounds will be gone, but they have only been used in a few places, and need to be redesigned.

## References
closes #_insert number here_


aditya-mitra (approved):
Looks good! ✅

SYBIOTE (approved):
looks good

PR #10471, opened 2024-06-27
Fix xrui pointer input & improved follow camera 

SamMazerIR:
## Summary
XRUI pointer input changes, as well as orbit/follow camera work

## References
closes #_insert number here_


packages/engine/tsconfig.json
SamMazerIR, line 1: delete this - (so sayeth Daniel)

the-reality-engineer (approved):

DanielBelmes (approved):
Looking good. There are some regressions in camera zoom because of some assumptions made with zoomLevel. But we can fix them after

PR #10476, opened 2024-06-28
fix loading screens

aditya-mitra:
## References
closes #_insert number here_

#### Before

https://github.com/EtherealEngine/etherealengine/assets/55396651/c945670e-bc24-4d16-bf19-babef043db52

#### After

https://github.com/EtherealEngine/etherealengine/assets/55396651/a1b72706-317d-450f-942d-d33aa723a2c3


packages/client/src/main.tsx
```diff
@@ -92,7 +92,6 @@ const App = () => {
             element={
               <Suspense>
                 <AppPage>
-                  <Tailwind />
```
HexaField, line 95: The reason were we're lazily importing tailwind here is because the location route breaks with it there. We need to only import it for other routes.

HexaField (changes requested):

packages/client/src/main.tsx
aditya-mitra: fixed share menu qr

image

HexaField (approved):

PR #10478, opened 2024-06-28
studio: new settings menu ui

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-2867
## References
closes #_insert number here_


aditya-mitra (changes requested):
the client settings menu is in the location route which uses the old material ui currently and not tailwind. which happens to break the ui

do we have the design already for the user menu?

image

aditya-mitra:
merged with current dev which has tailwind planted in all routes. the select menus and tabs are non-functional and the modals appear broken

image

achen5671:
> merged with current dev which has tailwind planted in all routes. the select menus and tabs are non-functional and the modals appear broken
> 
> image

The last time I check the select menu has always been broken. I created a separate ticket to address the issue which I will take a look at. This PR just updates the UI.

achen5671:
> > merged with current dev which has tailwind planted in all routes. the select menus and tabs are non-functional and the modals appear broken
> > image
> 
> The last time I check the select menu has always been broken. I created a separate ticket to address the issue which I will take a look at. This PR just updates the UI.

fixed ui

Screenshot 2024-07-01 at 1 03 55 PM

achen5671:
@aditya-mitra  I updated it to use the Modal component. I also broke out the new settings to it's own component called settings2 like we talked about. Let me know if there are any other changes you will like :)

packages/client-core/src/user/UserUISystem.tsx
```diff
@@ -82,6 +84,7 @@ const reactor = () => {
     popupMenuState.menus.merge({
       [UserMenus.Profile]: ProfileMenu,
       [UserMenus.Settings]: SettingMenu,
```
HexaField, line 86: If the old settings menu is not being used anymore, we should remove it and replace it with this new one.

HexaField (dismissed):

packages/client-core/src/user/UserUISystem.tsx
HexaField: Oh sorry, I see that this new one is to be used in tailwind routes instead, is that correct?

achen5671:
@aditya-mitra per up on our talk last week. Im using the MUI modal here for now to be compatible with the location route. 

Our current modal isn't compatible but in the near future we should update our current modal so that it is.

pr is ready for rereview

aditya-mitra (approved):
awesome 🚀

PR #10480, opened 2024-06-28
Material IO Fixes

dinomut1:
Several fixes to material system that arose when integrating vertex color lerping material.

- Change EEMaterialImporter extension to be first in the extension list so it beats the default MeshPhysicalMaterial parser extension to define materials
- Force material update on editing properties in material editor to automatically propagate all changes in material
- Fix issue in material library system where imported materials are getting instantiated twice
- Add a couple more sections where material.userData.type is preferred over material.type


AidanCaruso:
This looks good, I'll be QA testing shortly.

AidanCaruso (approved):

jonathan-casarrubias (approved):

PR #10483, opened 2024-06-29
IR-2806 Material instancing fixes

AidanCaruso:
## Summary

1. Fixes material states being removed immediately on instance creation
2. Materials are no longer automatically deduped based on a hash map of material sources to entities
3. Moves some functions to the spatial package
4. Ensures that materials created of the same UUID are now replaced rather than duplicated
5. Improved handling of material state updates propagating to all instances to prevent material object references being lost 

## References
closes [IR-2806](https://tsu.atlassian.net/browse/IR-2806)

## QA Steps

[IR-2806]: https://tsu.atlassian.net/browse/IR-2806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HexaField:
Screenshot from 2024-07-08 09-14-12
some errors upon opening the studio with no scene loaded

AidanCaruso:
> Screenshot from 2024-07-08 09-14-12 some errors upon opening the studio with no scene loaded

fixed

packages/engine/src/scene/materials/systems/MaterialLibrarySystem.tsx
```diff
@@ -89,7 +89,7 @@ const MaterialEntityReactor = () => {
   const entity = useEntityContext()
   const materialComponent = useComponent(entity, MaterialStateComponent)
   useEffect(() => {
-    if (!materialComponent.instances.value!) return
+    if (!materialComponent.instances.value! || !SourceComponent.entitiesBySource) return
```
HexaField: entitiesBySource is an object and will always return as truthy, I think the correct fix here is to check for the record read on line 94 prior to the nested for loop

HexaField (changes requested):

packages/engine/src/scene/materials/systems/MaterialLibrarySystem.tsx
HexaField: oh, github was showing me out of date info lol...

packages/engine/src/scene/materials/systems/MaterialLibrarySystem.tsx
```diff
@@ -89,7 +89,7 @@ const MaterialEntityReactor = () => {
   const entity = useEntityContext()
   const materialComponent = useComponent(entity, MaterialStateComponent)
   useEffect(() => {
-    if (!materialComponent.instances.value! || !SourceComponent.entitiesBySource) return
+    if (!materialComponent.instances.value! || !Object.keys(SourceComponent.entitiesBySource)) return
```
HexaField: by checking for object keys here, we escaping all conditionals rather than checking if sourceEntity has a source component and if entitiesBySource has data for that entity? that seems to be a more appropriate check

HexaField (changes requested):

packages/engine/src/scene/materials/systems/MaterialLibrarySystem.tsx
AidanCaruso: Yes you're right - my bad for multitasking while debugging this . Fixed.

dinomut1 (changes requested):
Running into issues with custom material initialization when loading a scene that has a .gltf in it /w custom materials.

To reproduce:
1. Download this file: [double-mat-test.zip](https://github.com/user-attachments/files/16135405/double-mat-test.zip)
This has two cubes in it that have the LavaMaterial from our advanced materials repo. 
2. In any scene, add this file to the scene. The file should load correctly, with two lava cubes (with depth test turned off) loading into the scene.
3. Save the scene.
4. Refresh the scene. On loading, the lava material is instead just a MeshStandardMaterial

So the material is correctly instantiated when added to an already open scene, but doesn't correctly instantiate when loaded as part of a scene

dinomut1:
Found the issue -- custom materials are getting exported with null plugins:

Screenshot from 2024-07-08 21-36-35

Upon reimport, these null plugins are causing the customProgramCacheKey to be overwritten with an empty string, which means all custom materials get cached as the same shader in the WebGL context. I partially fixed this by making our customProgramCacheKey handling consider the shader code itself in addition to the plugins. However still not sure how the null plugins are getting there in the first place. 

If this change to customProgramCacheKey causes problems feel free to revert these commits but I think this is closer to what we want.

AidanCaruso:
> Found the issue -- custom materials are getting exported with null plugins:
> 
> Screenshot from 2024-07-08 21-36-35
> 
> Upon reimport, these null plugins are causing the customProgramCacheKey to be overwritten with an empty string, which means all custom materials get cached as the same shader in the WebGL context. I partially fixed this by making our customProgramCacheKey handling consider the shader code itself in addition to the plugins. However still not sure how the null plugins are getting there in the first place.
> 
> If this change to customProgramCacheKey causes problems feel free to revert these commits but I think this is closer to what we want.

Fixed the null plugins being added - I believe this was a minor oversight in how plugins are mapped to values in the material exporter. Reexporting the platform.glb with the lava material now yields an empty array of plugins rather than null values.

dinomut1 (approved):

HexaField (approved):

PR #10486, opened 2024-07-01
ProjectService: Replace API calls with feathers hooks

CITIZENDOT:
## Summary
Replaced API calls with feathers hooks in project service.

## QA Steps

[IR-2670]: https://tsu.atlassian.net/browse/IR-2670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/client-core/src/admin/common/Project/ProjectDrawer.tsx
```diff
@@ -70,14 +70,14 @@ const ProjectDrawer = ({ open, inputProject, existingProject = false, onClose, c
 
   const projectUpdateStatus = useHookstate(getMutableState(ProjectUpdateState)[project.name]).value
 
-  const handleSubmit = async () => {
+  const handleSubmit = () => {
     try {
       if (existingProject && changeDestination) {
-        if (inputProject) await ProjectService.setRepositoryPath(inputProject.id, projectUpdateStatus.destinationURL)
```
HexaField, line 76: we need to still be awaiting all these calls, otherwise the UI will never reflect requests in progress

packages/client-core/src/common/services/ProjectService.ts
```diff
@@ -77,33 +77,37 @@ export const ProjectState = defineState({
...
     }
   },
 
   // restricted to admin scope
-  createProject: async (name: string, params?: ProjectUpdateParams) => {
-    const result = await Engine.instance.api.service(projectPath).create({ name }, params)
-    logger.info({ result }, 'Create project result')
-    await ProjectService.fetchProjects()
+  createProject: (name: string, params?: ProjectUpdateParams) => {
+    const result = useService(projectPath, 'create', {
```
HexaField, line 100: these need to all stay with the old API, react hooks can only be used in a react context

HexaField (changes requested):

packages/client-core/src/admin/common/Project/ProjectDrawer.tsx
CITIZENDOT: The functions (in this case `setRepositoryPath`) has no await call in it, Is it still valid if it's defined as async function?

HexaField: The point is that these are API calls, they are necessarily going to be async.

PR #10487, opened 2024-07-01
upload from assets panel

Adam-Zimmer:
## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/components/editor/panels/Assets/container/index.tsx
+++ b/packages/ui/src/components/editor/panels/Assets/container/index.tsx
@@ -27,16 +27,21 @@ import { clone, debounce, isEmpty, last } from 'lodash'
 import React, { createContext, useContext, useEffect, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 
+import { NotificationService } from '@etherealengine/client-core/src/common/services/NotificationService'
 import {
-  staticResourcePath,
   StaticResourceQuery,
-  StaticResourceType
+  StaticResourceType,
+  fileBrowserPath,
+  staticResourcePath
 } from '@etherealengine/common/src/schema.type.module'
 import { Engine } from '@etherealengine/ecs/src/Engine'
 import { AssetsPanelCategories } from '@etherealengine/editor/src/components/assets/AssetsPanelCategories'
 import { AssetSelectionChangePropsType } from '@etherealengine/editor/src/components/assets/AssetsPreviewPanel'
+import { inputFileWithAddToScene } from '@etherealengine/editor/src/functions/assetFunctions'
+import { EditorState } from '@etherealengine/editor/src/services/EditorServices'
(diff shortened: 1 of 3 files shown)


packages/ui/src/components/editor/panels/Assets/container/index.tsx
```diff
@@ -315,6 +321,14 @@ const AssetPanel = () => {
   const searchedStaticResources = useHookstate<StaticResourceType[]>([])
   const searchText = useHookstate('')
   const breadcrumbPath = useHookstate('')
+  const originalPath = useMutableState(EditorState).projectName.value
+
+  console.log(`ORIGPATH: ${originalPath}`)
```
DanielBelmes: debug log

packages/ui/src/components/editor/panels/Assets/container/index.tsx
```diff
@@ -473,8 +486,8 @@ const AssetPanel = () => {
     // TODO: add settings functionality
   }
 
-  const handleUpdateAsset = () => {
-    // TODO: add upload asset functionality
+  const refreshDirectory = async () => {
+    assetsQuery.refetch()
   }
```
jonathan-casarrubias: refreshDirectory is defined as an async arrow function, but there's no awaiting within the block context, maybe assetsQuery.refresh() is missing an await?

packages/ui/src/components/editor/panels/Assets/container/index.tsx
```diff
@@ -473,8 +486,8 @@ const AssetPanel = () => {
     // TODO: add settings functionality
   }
 
-  const handleUpdateAsset = () => {
-    // TODO: add upload asset functionality
+  const refreshDirectory = async () => {
```
jonathan-casarrubias: My apologies for the confusion, my previous wasn't a change request but more like a question... Based on the fact that I'm still unfamiliar with the system I wanted to bring up that maybe assetsQuery.refetch() was missing an await just because the wrapper function was defined as an async function.

But I just went through the code in order to confirm if refetch needed to be awaited or not, I found that refetch is a reference of a function named fetch() which isn't defined as a promise, therefore neither assetsQuery.refetch() requires to be awaited, nor refreshDirectory requires to be defined as a promise.

```ts
  const refreshDirectory = () {
    assetsQuery.refetch()
  }
```
I still have a question though, why do we need to wrap assetsQuery.refetch() within a new function when nothing else is being done within that function (refreshDirectory), can't we simply directly use assetsQuery.refetch() instead?

Example

```ts
onClick={() =>
            inputFileWithAddToScene({
              projectName: originalPath as string,
              directoryPath: `projects/${originalPath}/assets/`
            })
              .then(assetsQuery.refetch)
              .catch((err) => {
                NotificationService.dispatchNotify(err.message, { variant: 'error' })
              })
          }
```

DanielBelmes (approved):

jonathan-casarrubias (approved):

PR #10441, opened 2024-06-22
Coordinate Space Refactor + Separate Physics World Per Scene

HexaField:
## Summary

- [x] Pass physicsWorld into all physics functions
- [x] Rename SceneComponent “children” field to “scenes”
- [x] Separate physics world for each auto generated scene component
- [x] Change XR spaces
- [x] Scale scenes rather than camera

## References
closes #_insert number here_

PR #10492, opened 2024-07-01
IR-2865-Loading-locations-offscreen-throw-a-variety-of-breaking-errors

MichaelEstes:
## References
closes #_insert number here_


HexaField:
Since there are only physics world changes here, this might become redundant after https://github.com/EtherealEngine/etherealengine/pull/10441 

we should probably wait for that to be in before testing this again

dinomut1:
Is this still necessary after we've merged in the coordinate space refactor?

SamMazerIR (approved):

MichaelEstes:
> Since there are only physics world changes here, this might become redundant after #10441
> 
> we should probably wait for that to be in before testing this again

I can't reproduce the issues on dev with #10441 merged in, I'm going to close this

PR #10501, opened 2024-07-02
fix variable and template naming

SYBIOTE:
fixes lag in visual script and changes how variables and templates are named


packages/editor/src/components/visualScript/hooks/useTemplateHandler.ts
```diff
@@ -50,7 +48,7 @@ export const useTemplateHandler = ({
 
   const createGraphTemplate = (nodes: Node[], edges: Edge[]): GraphTemplate => ({
     id: uuidv4(),
-    name: uniqueId('New template '),
+    name: 'New template ' + Math.random().toString(36).slice(-6),
```
aditya-mitra, line 51: i think we should `uuidv4` instead

packages/editor/src/components/visualScript/hooks/useVariableHandler.ts
```diff
@@ -38,7 +37,7 @@ export const useVariableHandler = ({
 }: Pick<visualScriptFlow, 'variables' | 'setVariables'>) => {
   const createVariable = (): VariableJSON => ({
     id: uuidv4(),
-    name: uniqueId('variable '),
+    name: 'variable ' + Math.random().toString(36).slice(-6),
```
aditya-mitra, line 40: same here for uuid

packages/ui/src/components/editor/panels/VisualScript/node/picker/index.tsx
```diff
@@ -69,7 +70,12 @@ const PickerItem = ({ label, node, onPickNode, position, instance, color }) => {
 
   const handleNodeClick = () => {
     if (isLeafNode) {
-      onPickNode(node.type, instance.screenToFlowPosition(position))
+      console.log('DEBUG final pos', instance.screenToFlowPosition(position))
```
aditya-mitra: this debug log can be removed

aditya-mitra (changes requested):

packages/editor/src/components/visualScript/hooks/useTemplateHandler.ts
SYBIOTE: In both cases there is a chance we will encounter the same id again, it won't break anything but the probability is low.

I'm not sure how uuidv4 works so I choose random instead

aditya-mitra: Well, we are using uuids everywhere (even in the database as primary keys)
With uuid it can be just a shorter line `'New template ' + uuidv4()`

an oversimplified explanation of uuid would be 128 bits are generated and then converted to hexadecimal format with dashes.

SYBIOTE: we will get a giant uuid following the template, here we are keeping it short to 6 alphanumericals only

aditya-mitra (approved):

PR #10502, opened 2024-07-02
Replace API calls with feathers hooks in InviteService

CITIZENDOT:
## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/social/services/InviteService.ts
+++ b/packages/client-core/src/social/services/InviteService.ts
@@ -38,6 +38,7 @@ import {
 import { Engine } from '@etherealengine/ecs/src/Engine'
 import { defineState, getMutableState, getState } from '@etherealengine/hyperflux'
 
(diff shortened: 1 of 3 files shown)


packages/client-core/src/social/services/InviteService.ts
```diff
@@ -268,9 +269,9 @@ export const InviteService = {
       NotificationService.dispatchNotify(err.message, { variant: 'error' })
     }
   },
-  declineInvite: async (invite: InviteType) => {
+  declineInvite: async (invite: InviteType, inviteMutation: UseMutationReturnType<typeof invitePath>) => {
     try {
-      await Engine.instance.api.service(invitePath).remove(invite.id)
```
HexaField, line 273: any time there is a side effect, we must be awaiting the API request

HexaField (changes requested):

PR #10504, opened 2024-07-02
Fix scene preview camera

SYBIOTE:
fix scene preview camera


packages/ui/src/components/editor/properties/scene/previewCamera/index.tsx
```diff
@@ -64,14 +63,18 @@ export const ScenePreviewCameraNodeEditor: EditorComponentType = (props) => {
     EditorControlFunctions.commitTransformSave([props.entity])
   }
 
+  const scene = new Scene()
```
HexaField: we should put the scene in module scope to avoid creating a new one each render

HexaField (dismissed):

packages/ui/src/components/editor/properties/scene/previewCamera/index.tsx
SYBIOTE: done

MichaelEstes (approved):

HexaField (approved):

PR #10506, opened 2024-07-02
Fix lightmap configuration in studio

dinomut1:
add uv channel export in material exporter
fix texture thumbnails + inputs not updating correctly on add / remove texture


AidanCaruso:
Looks good, used this to finally fix my broken lightmaps in the BPCEM test scene - though the BPCEM in this scene is broken now, totally unrelated.
image

AidanCaruso (approved):

PR #10507, opened 2024-07-03
IR-2900-Studio-Deleting-Directional-Light-Broken

MichaelEstes:
## References
closes #_insert number here_


packages/spatial/src/renderer/components/DirectionalLightComponent.ts
```diff
@@ -36,13 +38,19 @@ import {
 import { useEntityContext } from '@etherealengine/ecs/src/EntityFunctions'
 import { matches, useMutableState } from '@etherealengine/hyperflux'
 
+import { EntityUUID, UUIDComponent, defineQuery } from '@etherealengine/ecs'
+import { RenderSettingsComponent } from '@etherealengine/engine/src/scene/components/RenderSettingsComponent'
```
HexaField: spatial package cannot access RenderSettingsComponent. cleaning up the CSM should not be the responsibility of the DirectionalLightComponent, we should implement this in the shadow system

HexaField (changes requested):

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -451,6 +434,41 @@ const execute = () => {
...
+    // CSM directional lights should be their own component
+    for (const renderSettingsEntity of renderSettingsQuery()) {
+      const renderSettings = getComponent(renderSettingsEntity, RenderSettingsComponent)
+      if (renderSettings.primaryLight === uuid) {
+        getMutableComponent(renderSettingsEntity, RenderSettingsComponent).merge({
+          csm: false,
+          primaryLight: '' as EntityUUID
+        })
+        if (getState(EngineState).isEditing) {
+          const source = getComponent(renderSettingsEntity, SourceComponent)
```
HexaField: Let's move this logic to the studio in a reactor, since it really is an authoring layer concept.

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -451,6 +434,41 @@ const execute = () => {
     return
   }
 
+  for (const entity of directionalLightQuery.enter()) {
```
HexaField: Using enter and exit queries is a practice we're trying to avoid for the exact reason the comment here specifies. I think this can be simplified in the studio by having a reactor run that listens for the entity's existence specified by the `primaryLight` field on the RenderSettingsComponent, setting it to `""` if the entity is removed, and if the value is `""` and then if the user adds a `DirectionalLightComponent`, it auto populates that field.

HexaField (changes requested):

packages/ui/src/components/editor/properties/render/index.tsx
```diff
@@ -123,6 +127,23 @@ export const RenderSettingsEditor: EditorComponentType = (props) => {
     })
   )
 
+  useEffect(() => {
```
HexaField, line 130: Unfortunately having this here means we will need to open the render settings propreties UI in order for the light to be updated, which kind of defeats the purpose. We can just move this to a reactor in a system somewhere to ensure it's always running

packages/engine/src/scene/components/RenderSettingsComponent.ts
```diff
@@ -93,6 +95,24 @@ export const RenderSettingsComponent = defineComponent({
...
+      let primaryLight = UndefinedEntity
+      for (const directionalLightEntity of directionalLightEntities) {
+        if (component.primaryLight.value === getComponent(directionalLightEntity, UUIDComponent)) {
+          primaryLight = directionalLightEntity
+          break
+        }
+      }
+
+      if (!primaryLight && component.primaryLight.value) {
+        component.merge({
```
HexaField: do we actually want to mutate this here? If the data is invalid it should just do nothing

HexaField (changes requested):

packages/ui/src/components/editor/properties/render/index.tsx
MichaelEstes: If the value points to a light that doesn't exist it shouldn't matter until that UI is opened anyways.

HexaField: that's a fair point!

HexaField (approved):

PR #10509, opened 2024-07-03
[IR-2491] studio: fix delete object from hierarchy

dtlehrer:
## Summary
[IR-2491] Error After Deleting Object from Hierarchy in Studio
- fixes bug: now able to update gltf state without throwing an error and crashing app when deleting/removing nodes from the hierarchy panel

[IR-2491]: https://tsu.atlassian.net/browse/IR-2491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HexaField (commented):
Looks good, can you add a regression failure case test to catch what caused the problem?

HexaField (approved):

PR #10510, opened 2024-07-03
IR-2748: Publish events on CRUD operations of static-resources

CITIZENDOT:
## References
https://tsu.atlassian.net/browse/IR-2748


packages/server-core/src/media/static-resource/static-resource.ts
```diff
@@ -57,4 +60,29 @@ export default (app: Application): void => {
 
   const service = app.service(staticResourcePath)
   service.hooks(hooks)
+
+  const onCRUD = (data: StaticResourceType | Paginated<StaticResourceType> | StaticResourceType[]) => {
+    const targetIds: string[] = []
```
HexaField: `userId` on static resources is not used currently. instead, we should do a query for all users who have project permissions for the project this static resource is a part of (which can be nullish, in which case we don't want to notify anyone)

HexaField (changes requested):

packages/server-core/src/media/static-resource/static-resource.ts
```diff
@@ -57,4 +60,29 @@ export default (app: Application): void => {
...
+    if (Array.isArray(data)) {
+      data.forEach((item) => {
+        targetIds.push(item.userId)
```
hanzlamateen: I dont think userId does much here.

This will trigger refresh in client only for user who uploaded that static-resource.

Maybe we can update onCRUD method to match our needs at the moment. Which is it should only triggered when static-resource's type column was `scene`.

Further more the targetIds which be evaluated based on the project that scene was in and then, project permissions should be used to find userId like its done in project.ts

hanzlamateen (changes requested):

packages/server-core/src/media/static-resource/static-resource.ts
```diff
@@ -61,28 +68,70 @@ export default (app: Application): void => {
...
+      for (const projectName of projectNames) {
+        const projectId = await app.service(projectPath).find({
+          query: {
+            name: projectName
+          }
+        })
+        if (projectId.total === 1) {
+          projectIds.push(projectId.data[0].id)
+        }
+      }
```
hanzlamateen: This can be optimized to single api call using 

```
query: {
 name: { $in: [ARRAY_OF_NAMES] },
 $select: ['id']
}
```

https://feathersjs.com/api/databases/querying.html#select
https://feathersjs.com/api/databases/querying.html#in-nin

packages/server-core/src/media/static-resource/static-resource.ts
```diff
@@ -61,28 +68,70 @@ export default (app: Application): void => {
...
+        data.data.forEach((item) => {
+          if (item.project && item.type === 'scene') {
+            projectNames.push(item.project)
+          }
+        })
+      } else {
+        if (data.project && data.type === 'scene') {
+          projectNames.push(data.project)
+        }
+      }
```
hanzlamateen:
```suggestion
      } else if (data.project && data.type === 'scene') {
          projectNames.push(data.project)
      }
```

packages/server-core/src/media/static-resource/static-resource.ts
```diff
@@ -61,28 +68,70 @@ export default (app: Application): void => {
...
+        const projectReadScopes = await app.service(scopePath).find({
+          query: {
+            type: 'projects:read' as ScopeType
+          },
+          paginate: false
+        })
+
+        projectReadScopes.forEach((scope) => {
+          targetIds.push(scope.userId)
+        })
```
hanzlamateen: We dont need to do this for each project of for loop, as `project:read` scope is global

hanzlamateen (changes requested):
If possible also make suggested enhancements in project.ts file

packages/server-core/src/media/static-resource/static-resource.ts
CITIZENDOT: Updated!

packages/server-core/src/media/static-resource/static-resource.ts
CITIZENDOT: Updated!

packages/server-core/src/media/static-resource/static-resource.ts
CITIZENDOT: This is great! thanks. Updated!

packages/server-core/src/media/static-resource/static-resource.ts
CITIZENDOT: Fixed!

packages/server-core/src/media/static-resource/static-resource.ts
CITIZENDOT: Fixed!

packages/server-core/src/media/static-resource/static-resource.ts
```diff
@@ -57,4 +67,65 @@ export default (app: Application): void => {
...
+
+      const projects = await app.service(projectPath).find({
+        query: {
+          name: { $in: projectNames },
+          $select: ['id']
+        },
+        paginate: false
+      })
+
+      const targetIds: string[] = []
```
HexaField: is this going to be sending all static resources to all users who have permission? or sending only each of the static resources to those who have permission for each project?

CITIZENDOT: So, this is what it's doing:
- Fetch which projects this static-resource belongs to (can be multiple because, static-resource can be array)
- Fetch users who have access to these projects
- Publish event to all these users.

HexaField: right, so if the event has an array, and that array contains resources from multiple projects, are we sending all the resources to all users who have permission for any of the projects, or sending only each resource in the array to the users who have permission?

CITIZENDOT: Interesting! Right now we're sending all data to all captured users. But:
> sending only each resource in the array to the users who have permission

this makes more sense.

hanzlamateen (approved):

PR #10514, opened 2024-07-03
File Browser: Fix Thumbnails Not Displaying for Files Not In Current Project

dinomut1:
Parses the project name of a given file from its path to correctly query static resources for its thumbnail


HexaField (changes requested):
This won't work in multitenancy as project names are of the format `orgname/projectname`

HexaField:
The file browser should already have the current directory's project name that can be passed down into the file grid component

dinomut1:
> This won't work in multitenancy as project names are of the format `orgname/projectname`

Yes it will. Regexes don't strictly match from the beginning of a string unless you put a "^" at the start of it.

See https://regexr.com/82gik

HexaField:
Except this is not the structure of MT

Screenshot from 2024-07-05 08-29-10

This is
Screenshot from 2024-07-05 08-28-35

And we want to capture `orgname/projectname` as the whole project name.

HexaField (approved):

dinomut1:
Gotcha also better to not use regexes anyways

aditya-mitra (approved):

PR #10517, opened 2024-07-05
IR: 2920: Re-evaluate feathers validator usage for user services

MoizAdnan:


packages/common/src/schemas/user/identity-provider.schema.ts
```diff
@@ -45,7 +45,9 @@ export const identityProviderTypes = [
   'facebook',
   'twitter',
   'linkedin',
-  'auth0'
+  'auth0',
+  'guest',
+  'user'
```
hanzlamateen: When is `user` type used? In don't see it being used anywhere

hanzlamateen (changes requested):

packages/common/src/schemas/user/identity-provider.schema.ts
MoizAdnan: it is being used in `identity-provider.test` on line 164

hanzlamateen: I think that test needs to be updated then.

PR #10520, opened 2024-07-05
refetch scenes after new scene on toolbar

aditya-mitra:
## Summary

the `newScene` and `deleteScene` functions use the file-browser api to create new scenes. Once the API responds, the scenes would refresh by listening to it.

## References

https://tsu.atlassian.net/browse/IR-2947

### DIFF
--- a/packages/ui/src/components/editor/panels/Scenes/container/index.tsx
+++ b/packages/ui/src/components/editor/panels/Scenes/container/index.tsx
@@ -24,12 +24,12 @@ Ethereal Engine. All Rights Reserved.
 */
 
 import { PopoverState } from '@etherealengine/client-core/src/common/services/PopoverState'
-import { StaticResourceType, staticResourcePath } from '@etherealengine/common/src/schema.type.module'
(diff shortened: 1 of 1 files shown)


SYBIOTE (approved):
basically, refetch has been abstracted away

packages/ui/src/components/editor/panels/Scenes/container/index.tsx
```diff
@@ -56,18 +56,18 @@ export default function ScenesPanel() {
     getMutableState(EditorState).scenePath.set(scene.key)
   }
 
+  useRealtime(fileBrowserPath, scenesQuery.refetch)
```
the-reality-engineer, line 59: We are already calling useRealtime in FeathersChildReactor for each feathers request. Why do we need to do it again here?

the-reality-engineer (dismissed):

packages/ui/src/components/editor/panels/Scenes/container/index.tsx
HexaField: It's causing a referch for a different query

HexaField (approved):

PR #10521, opened 2024-07-05
Update useRealtime hook to pass more data to refetch

CITIZENDOT:
## Summary

`useRealtime` now passes the message received from server's channel and the `eventType` which caused it.
`eventType` can be of the type: `'created' | 'updated' | 'patched' | 'removed'`.

Related to:
- https://tsu.atlassian.net/browse/IR-2748
- https://github.com/theinfinitereality/eepro-multitenancy/pull/307

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/common/functions/FeathersHooks.tsx
+++ b/packages/spatial/src/common/functions/FeathersHooks.tsx
@@ -357,14 +357,17 @@ export function hashObject(obj) {
  * An internal hook that will listen to realtime updates to a service
  * and update the cache as changes happen.
  */
-export function useRealtime(serviceName: keyof ServiceTypes, refetch: () => void) {
+export function useRealtime(
+  serviceName: keyof ServiceTypes,
+  refetch: (data: any, eventType: 'created' | 'updated' | 'patched' | 'removed') => void
+) {
   useLayoutEffect(() => {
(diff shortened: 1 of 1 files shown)


packages/spatial/src/common/functions/FeathersHooks.tsx
```diff
@@ -357,14 +357,17 @@ export function hashObject(obj) {
...
+  refetch: (data: any, eventType: 'created' | 'updated' | 'patched' | 'removed') => void
+) {
   useLayoutEffect(() => {
     const service = Engine.instance.api.service(serviceName)
 
-    service.on('created', refetch)
-    service.on('updated', refetch)
-    service.on('patched', refetch)
-    service.on('removed', refetch)
+    service.on('created', (data: any) => refetch(data, 'created'))
```
HexaField: this will not properly clean up the event handlers. they need to be defined as consts and those passed into the off handlers upon unmount of the effect.

HexaField (changes requested):

packages/spatial/src/common/functions/FeathersHooks.tsx
CITIZENDOT: Fixed!

HexaField (approved):

PR #10522, opened 2024-07-05
make figtree as main font 

SYBIOTE:
## Summary
makes figtree the main font in the code base
## References
closes #_insert number here_

### DIFF
--- a/packages/client/src/engine.tsx
+++ b/packages/client/src/engine.tsx
@@ -35,6 +35,7 @@ import { getMutableState } from '@etherealengine/hyperflux'
 import { EngineState } from '@etherealengine/spatial/src/EngineState'
 import { startTimer } from '@etherealengine/spatial/src/startTimer'
 
+import MetaTags from '@etherealengine/client-core/src/common/components/MetaTags'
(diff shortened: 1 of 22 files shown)


CITIZENDOT (approved):
LGTM!

`eepro-multitenancy` has several instances of `fontFamily=['Figtree']`. Could you remove those too?

SYBIOTE:
i think we need those for tailwind to apply the font itself, and even if its redundant i don't see why that's harmful, if we need to remove inter I'm up for it

> LGTM!
> 
> `eepro-multitenancy` has several instances of `fontFamily=['Figtree']`. Could you remove those too?

CITIZENDOT:
Currently, using the `fontFamily=['Figtree']` class repeatedly isn't causing any issues since it's our default font. However, for future flexibility, if we decide to change the default font, it would be much easier to update it in a single place (excluding exceptions).

It also helps to declutter the components when we pass only the necessary attributes.

Edit: Just merged this in: https://github.com/theinfinitereality/eepro-multitenancy/pull/316, Everything is `Figtree` now!

PR #10537, opened 2024-07-08
 add modal to new scene button

JT00y:
## References
closes #_insert number here_


packages/editor/src/components/EditorContainerState.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const wizardSceneButton = (): JSX.Element => {
+  return (
+    <Button size="small" variant="outline" className="text-center">
+      wizard Scene
+    </Button>
+  )
+}
+
+export const EditorContainerState = defineState({
```
dinomut1: let's use EditorState for this instead, as used in https://github.com/EtherealEngine/etherealengine/pull/10538

packages/editor/src/components/EditorContainerState.tsx
```diff
@@ -0,0 +1,61 @@
...
+      onClick={() => {
+        onNewScene()
+      }}
+    >
+      new empty Scene
+    </Button>
+  )
+}
+
+const wizardSceneButton = (): JSX.Element => {
```
dinomut1: We'll inject this into the New Scene modal from the wizard project

dinomut1 (changes requested):

packages/editor/src/components/Editor2Container.tsx
```diff
@@ -119,6 +119,19 @@ const defaultLayout: LayoutData = {
 
 const EditorContainer = () => {
   const { sceneAssetID, sceneName, projectName, scenePath, uiEnabled, uiAddons } = useMutableState(EditorState)
+  uiAddons.newScene['emptyScene'].set(
```
dinomut1: Let's hard-code this into the create scene panel dialog, rather than here. It's not technically an addon so this could confuse 3rd party devs

dinomut1 (approved):

PR #10550, opened 2024-07-09
IR-2883-Studio-Unable-to-reparent-child-to-different-parent-directly

MichaelEstes:
## Summary
Builds the scene hierarchy panel based on the data in the authoring layer instead of the data in the ECS. By default this hides children created from model components as those entities can not be edited or re-parented, that can be changed with the feature flag 'ir.editor.ui.hierarchy.hideModelChildren'.

## References
closes #_insert number here_


dinomut1 (approved):
looks good

PR #10551, opened 2024-07-09
Optimize Thumbnail Fetching

dinomut1:
Fetch thumbnails all at once in file browser 'Find' operation


HexaField:
Added another small optimization by doing the thumbnail query after project perm filtering

HexaField (approved):

PR #10553, opened 2024-07-09
chore: Update project permission schema and query schema

MoizAdnan:


DanielBelmes (approved):
QA'd locally and it works again

PR #10556, opened 2024-07-09
Michael/ir 2479 trigger

MbfloydIR:
## Summary
Added logic to check/wait for the collider component's collider existence before trying to make it a trigger

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/physics/components/ColliderComponent.tsx
+++ b/packages/spatial/src/physics/components/ColliderComponent.tsx
@@ -27,7 +27,7 @@ import { useEffect, useLayoutEffect } from 'react'
 import { Vector3 } from 'three'
 
 import { defineComponent, useComponent, useEntityContext, useOptionalComponent } from '@etherealengine/ecs'
-import { getState } from '@etherealengine/hyperflux'
+import { getState, useState } from '@etherealengine/hyperflux'
 
 import { useAncestorWithComponent } from '../../transform/components/EntityTree'
 import { TransformComponent } from '../../transform/components/TransformComponent'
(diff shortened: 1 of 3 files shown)


HexaField:
@cla-bot check

packages/spatial/src/physics/components/TriggerComponent.ts
```diff
@@ -91,14 +91,14 @@ export const TriggerComponent = defineComponent({
     const colliderComponent = useComponent(entity, ColliderComponent)
 
     useEffect(() => {
-      if (!colliderComponent?.value) return
+      if (!colliderComponent?.value || !colliderComponent?.hasCollider.value) return
```
HexaField: we should just move this useEffect to the collider component

HexaField (changes requested):

packages/spatial/src/physics/components/TriggerComponent.ts
MbfloydIR: implemented

packages/spatial/src/physics/components/ColliderComponent.tsx
```diff
@@ -49,7 +50,8 @@ export const ColliderComponent = defineComponent({
       friction: 0.5,
       restitution: 0.5,
       collisionLayer: CollisionGroups.Default,
-      collisionMask: DefaultCollisionMask
+      collisionMask: DefaultCollisionMask,
+      hasCollider: false
```
HexaField: let's scope hasCollider to the reactor as to not expose it on the component properties

HexaField (changes requested):

packages/spatial/src/physics/components/ColliderComponent.tsx
MbfloydIR: implemented

HexaField (approved):

HexaField:
Just need to fix tests. Likely need to ensure reactors have run, or change the order of setComponents to have the trigger before the collider

PR #10560, opened 2024-07-10
Publish dialog link unpublish

HexaField:
## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/primitives/tailwind/Modal/index.tsx
+++ b/packages/ui/src/primitives/tailwind/Modal/index.tsx
@@ -41,11 +41,14 @@ export interface ModalProps {
   submitLoading?: boolean
   showCloseButton?: boolean
   closeButtonDisabled?: boolean
+  customButtonDisabled?: boolean
   submitButtonDisabled?: boolean
   closeButtonText?: string
+  customButtonText?: string
   submitButtonText?: string
   onClose?: (isHeader: boolean) => void
   onSubmit?: () => void
+  onCustom?: () => void
 }
 
 export const ModalHeader = ({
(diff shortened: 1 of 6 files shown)


packages/ui/src/primitives/tailwind/Modal/index.tsx
```diff
@@ -114,33 +128,39 @@ export const ModalFooter = ({
...
   closeButtonDisabled,
+  customButtonDisabled,
   submitButtonDisabled,
   showCloseButton = true
 }: ModalProps) => {
   const twClassName = twMerge('relative z-50 max-h-[80vh] w-full bg-theme-surface-main', className)
   return (
     <div className={twClassName}>
       <div className="relative rounded-lg shadow">
-        {onClose && <ModalHeader title={title} onClose={onClose} />}
```
aditya-mitra, line 133: this conditional rendering is used by some modals to hide the modal header

packages/ui/src/primitives/tailwind/Modal/index.tsx
```diff
@@ -114,33 +128,39 @@ export const ModalFooter = ({
 const Modal = ({
   title,
   onClose,
+  onCustom,
```
aditya-mitra, line 131: we should not put 3 separate props just for this modal

in these cases, it is best to make another modal - also it looks like publish modal will need a modal of its own at some point

i can help create this modal quickly if you want

aditya-mitra (changes requested):

packages/ui/src/primitives/tailwind/Modal/index.tsx
HexaField: Yes please do

PR #10564, opened 2024-07-10
location url and unpublish functionalities

aditya-mitra:
## Summary

- Adds an unpublish button to unpublish/delete a published location
- Shows the url of the published location

## References
closes #_insert number here_


packages/editor/src/components/dialogs/PublishModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+      }
+    }
+  }
+
+  return (
+    <div className="relative z-50 max-h-[80vh] w-[50vw] bg-theme-surface-main">
+      <div className="relative rounded-lg shadow">
+        <ModalHeader
+          onClose={PopoverState.hidePopupover}
+          title={location?.id ? t('editor:toolbar.publishLocation.create') : t('editor:toolbar.publishLocation.update')}
```
hanzlamateen: image

After I unpusblished, its showing update text. And wrong `Publish Location` title is shown when a location is published.

packages/editor/src/components/dialogs/PublishModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+            />
+            <Input
+              type="number"
+              label={t('admin:components.location.lbl-maxuser')}
+              value={maxUsers.value}
+              onChange={(event) => maxUsers.set(Math.max(parseInt(event.target.value, 0), 0))}
+              error={errors.maxUsers.value}
+              disabled={submitLoading.value}
+            />
+            <Select
```
hanzlamateen: The values shown in this dropdown are wrong.

Also why do we need to show this dropdown at all? Because we are essentially in a scene publishing it. Maybe publish button appears only when we have a scene selected?

packages/editor/src/components/dialogs/PublishModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+  const scene = useHookstate((location ? location.sceneId : sceneID) ?? '')
+  const videoEnabled = useHookstate<boolean>(location?.locationSetting.videoEnabled || true)
+  const audioEnabled = useHookstate<boolean>(location?.locationSetting.audioEnabled || true)
+  const screenSharingEnabled = useHookstate<boolean>(location?.locationSetting.screenSharingEnabled || true)
+  const locationType = useHookstate(location?.locationSetting.locationType || 'public')
+
+  const scenes = useFind(staticResourcePath, {
+    query: {
+      paginate: false
```
hanzlamateen: This needs to have `type: 'scene'`

Do this fix for admin panel dialog aswell.

packages/editor/src/components/dialogs/PublishModal.tsx
```diff
@@ -0,0 +1,271 @@
...
+  serverError: ''
+})
+
+const locationTypeOptions = [
+  { label: 'Private', value: 'private' },
+  { label: 'Public', value: 'public' },
+  { label: 'Showroom', value: 'showroom' }
+]
+
+export default function PublishSceneModal({ location, sceneID }: { location?: LocationType; sceneID: string | null }) {
```
hanzlamateen: @HexaField do you think we need a separate control for this or `AddEditLocationModal` should be used?

hanzlamateen (changes requested):

packages/editor/src/components/dialogs/PublishModal.tsx
HexaField: If it's similar enough we should reuse AddEditLocationModal

packages/editor/src/components/dialogs/PublishModal.tsx
aditya-mitra: fixed!

packages/editor/src/components/dialogs/PublishModal.tsx
aditya-mitra: disabled it for studio publish button.
yes, the publish button is only visible when a scene is selected

packages/editor/src/components/dialogs/PublishModal.tsx
aditya-mitra: done

packages/editor/src/components/dialogs/PublishModal.tsx
aditya-mitra: done

packages/client-core/src/admin/components/locations/AddEditLocationModal.tsx
```diff
@@ -130,80 +143,145 @@ export default function AddEditLocationModal({
...
 
-  return (
-    <Modal
-      title={
-        location?.id ? t('admin:components.location.updateLocation') : t('admin:components.location.createLocation')
+  const unpublishLocation = async () => {
+    if (location?.id) {
+      submitLoading.set(true)
+      try {
+        await locationMutation.remove(location.id)
```
hanzlamateen:
```suggestion
        await locationMutation.remove(location.id, { query: {projectId: // need to send location id here} })
```

Getting following error when I am not not admin in MT. This is due to verifyProjectPermission hook failing (https://github.com/EtherealEngine/etherealengine/blob/e018fc9ed1c923ee1ce71c85fbd527b3d8f1e8d0/packages/server-core/src/social/location/location.hooks.ts#L250)

Same could be for patch and create

image

hanzlamateen (dismissed):

HexaField (changes requested):
Can we make it so when we click the publish/unpublish/update button, the modal doesn't close and refreshes with the new data?

HexaField (changes requested):
i'm seeing errors in the console too 
Screenshot from 2024-07-13 11-49-28

HexaField (approved):

PR #10568, opened 2024-07-10
[IR-3020] studio: hide ScenePreviewCameraComponent

dtlehrer:
## Summary
[IR-3020]

## References
closes #_insert number here_

## QA Steps

[IR-3020]: https://tsu.atlassian.net/browse/IR-3020?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/engine/src/FeatureFlagsState.tsx
+++ b/packages/engine/src/FeatureFlagsState.tsx
@@ -33,7 +33,7 @@ export const FeatureFlagsState = defineState({
   initial: {} as Record<string, boolean>,
   enabled(flagName: string) {
     const state = getMutableState(FeatureFlagsState)[flagName].value
-    return typeof state === 'boolean' ? state : true
+    return typeof state === 'boolean' ? state : false // returns false if state has not been set
   },
   useEnabled(flagName: string) {
     const state = useHookstate(getMutableState(FeatureFlagsState)[flagName]).value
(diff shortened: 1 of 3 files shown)


HexaField (changes requested):
we need to use feature flags for this

packages/engine/src/FeatureFlagsState.tsx
```diff
@@ -33,7 +33,7 @@ export const FeatureFlagsState = defineState({
   initial: {} as Record<string, boolean>,
   enabled(flagName: string) {
     const state = getMutableState(FeatureFlagsState)[flagName].value
-    return typeof state === 'boolean' ? state : true
+    return typeof state === 'boolean' ? state : false // returns false if state has not been set
```
HexaField: we want to preserve the default functionality of everything being enabled for local dev. we either need to seed everything into the database, or return this to default to true.

HexaField (dismissed):

packages/engine/src/FeatureFlagsState.tsx
dtlehrer: makes sense for local dev at the moment (and would be nice to have optionality around what is on/off + the ability to quickly toggle or set flag values in the future)

is there somewhere we are turning flags off for features that we don't want to be enabled in other environments?

dtlehrer: or should we just go ahead and seed/store key-value pairs for each flag in the db per environment?

noticing that we're reaching a point where there are several parts of the studio that others are asking to hide/disable beyond the local env

HexaField: reverting is good for now, we can address seeding later

MichaelEstes (approved):

PR #10572, opened 2024-07-10
Fixed issues with archiving projects

barankyle:
## Summary

Cloudfront function was not passing /temp paths as-is, and was mistakenly redirecting them to client/index.html

Some setups' projects have slashes in the project name (<org_name>/<project_name>), and slashes cannot be included in k8s job/service/pod/etc. names. Now replacing all disallowed characters in job names with slashes.

Resolves IR-1351

## References
closes #_insert number here_


HexaField (approved):
LGTM

PR #10576, opened 2024-07-10
added delete button for materials

MbfloydIR:
## Summary
Added a delete button to delete materials.

## References
closes #_insert number here_


packages/spatial/src/renderer/materials/materialFunctions.ts
```diff
@@ -197,6 +197,13 @@ export const assignMaterial = (entity: Entity, materialEntity: Entity) => {
   materialStateComponent.instances.set([...materialStateComponent.instances.value, entity])
 }
 
+export const removeMaterialEntity = (materialEntity: Entity) => {
```
AidanCaruso: This doesn't need to be its own function, let's just use removeEntity(materialEntity). Material entities are handled through queries for the material state component so they're gaurunteed to have this component.

packages/ui/src/components/editor/panels/Properties/material/index.tsx
```diff
@@ -302,6 +306,18 @@ export function MaterialEditor(props: { materialUUID: EntityUUID }) {
           </Button>
         </div>
       )}
+      <Button
+        className="w-full text-xs"
+        onClick={() => {
+          getMutableState(MaterialSelectionState).selectedMaterial.set(null)
+          removeMaterialEntity(entity)
+
+          //const newMaterial = new MeshBasicMaterial({ name: 'New Material' })
```
AidanCaruso: Remove commented code

AidanCaruso (changes requested):
Looks lke this PR deletes material entities but we aren't cleaning up the associated material instance component references or threejs materials. Looking at adding some cleanup to that end in the material library system reactors for IR-2943, will merge this PR as soon as that's in.

packages/ui/src/components/editor/panels/Properties/material/index.tsx
MbfloydIR: implemented

packages/spatial/src/renderer/materials/materialFunctions.ts
MbfloydIR: implemented

AidanCaruso:
image
@MbfloydIR Could we move the delete button to be in line with the Save/New buttons (defined in `packages/ui/src/components/editor/panels/Materials/container/index.tsx`? It makes more sense to place it with the other material state buttons rather than at the bottom of the material properties tab.

MbfloydIR:
@AidanCaruso I moved the delete button to the area next to the save and new buttons. 
I also coded it to only show up once a material has been selected, since you can't use it if no material is selected

AidanCaruso:
> @AidanCaruso I moved the delete button to the area next to the save and new buttons. I also coded it to only show up once a material has been selected, since you can't use it if no material is selected

Awesome, thank you!

AidanCaruso (approved):

packages/ui/src/components/editor/panels/Materials/container/index.tsx
```diff
@@ -144,6 +150,19 @@ export default function MaterialLibraryPanel() {
...
+            {hasSelectedMaterial.value && (
+              <Button
+                className="w-full text-xs"
+                onClick={() => {
+                  const entity = UUIDComponent.getEntityByUUID(selectedMaterial.value as EntityUUID)
+                  selectedMaterial.set(null)
+                  removeEntity(entity)
+                }}
+              >
+                Delete
```
SYBIOTE, line 163: can we put these in editor json, or use existing declarations

packages/ui/src/components/editor/panels/Materials/container/index.tsx
```diff
@@ -144,6 +150,19 @@ export default function MaterialLibraryPanel() {
             >
               New
```
SYBIOTE, line 151: can we put these in editor json, or use existing declarations

PR #10577, opened 2024-07-10
disallow project query without pagination

HexaField:
## Summary

We no longer are using the project service from the client without pagination, so we should disallow this to avoid causing considerable load on the server if someone was to request this.

## References
closes #_insert number here_

### DIFF
--- a/packages/common/src/schemas/projects/project.schema.ts
+++ b/packages/common/src/schemas/projects/project.schema.ts
@@ -117,7 +117,6 @@ export const projectQuerySchema = Type.Intersect(
     Type.Object(
       {
         assetsOnly: Type.Optional(Type.Boolean()),
-        paginate: Type.Optional(Type.Boolean()),
         action: Type.Optional(Type.String()),
         sourceURL: Type.Optional(Type.String()),
         destinationURL: Type.Optional(Type.String()),
--- a/packages/ui/src/components/editor/panels/Files/container/index.tsx
+++ b/packages/ui/src/components/editor/panels/Files/container/index.tsx
@@ -191,7 +191,7 @@ function extractDirectoryWithoutOrgName(directory: string, orgName: string) {
  * @todo will be optimized away once orgname is fully supported
  */
 export const useValidProjectForFileBrowser = (path: string) => {
-  const [orgName, projectName] = path.split('/').slice(2, 3)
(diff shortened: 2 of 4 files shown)


packages/common/src/schemas/projects/project.schema.ts
```diff
@@ -117,7 +117,6 @@ export const projectQuerySchema = Type.Intersect(
     Type.Object(
       {
         assetsOnly: Type.Optional(Type.Boolean()),
-        paginate: Type.Optional(Type.Boolean()),
```
hanzlamateen, line 120: I see it being still used in:

1. https://github.com/EtherealEngine/etherealengine/blob/dd999046f9f6c30089f8e4df12d045d47fe588d8/packages/editor/src/components/projects/ProjectsPage.tsx#L200
2. https://github.com/EtherealEngine/etherealengine/blob/dd999046f9f6c30089f8e4df12d045d47fe588d8/packages/ui/src/components/editor/panels/Files/container/index.tsx#L195
3.

hanzlamateen (changes requested):

packages/common/src/schemas/projects/project.schema.ts
HexaField: @hanzlamateen I have addressed both of these. The file browser has been optimized to only look for the two possible values, and the projects page is only used in single tenant and requires client pagination UI

packages/ui/src/components/editor/panels/Files/container/index.tsx
```diff
@@ -188,16 +188,28 @@ function extractDirectoryWithoutOrgName(directory: string, orgName: string) {
 
 /**
  * Gets the project name that may or may not have a single slash it in from a list of valid project names
+ * @todo will be optimized away once orgname is fully supported
  */
-export const useValidProjectForFileBrowser = (projectName: string) => {
+export const useValidProjectForFileBrowser = (path: string) => {
+  const [orgName, projectName] = path.split('/').slice(2, 3)
```
hanzlamateen: @HexaField I don't think we get `projectName` for MT/orgname scenarios:

image

hanzlamateen (changes requested):

hanzlamateen (approved):

PR #10578, opened 2024-07-11
Made some project fields conditionally submitted

barankyle:
## Summary
   
New validators will reject if you supply null for a string value.
You have to either supply a string or not supply the field at all.
Fixed this on project.update when creating or patching the project
record.

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/projects/project/project-helper.ts
+++ b/packages/server-core/src/projects/project/project-helper.ts
@@ -1478,43 +1478,39 @@ export const updateProject = async (
 
   const { commitSHA, commitDate } = await getCommitSHADate(projectName)
 
-  const returned = !existingProject
-    ? // Add to DB
-      await app.service(projectPath).create(
-        {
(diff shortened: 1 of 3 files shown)


HexaField (approved):

packages/server-core/src/projects/project/project-helper.ts
```diff
@@ -1486,43 +1486,39 @@ export const updateProject = async (
...
-          sourceRepo: data.sourceURL,
-          sourceBranch: data.sourceBranch,
-          updateType: data.updateType,
-          updateSchedule: data.updateSchedule,
-          updateUserId: userId || null
-        },
-        params
-      )
-
+  let returned
```
HexaField: We should preserve the type expected for `returned` here

HexaField (changes requested):

HexaField (approved):

PR #10579, opened 2024-07-11
server loads empty texture without crash

HexaField:
## References
closes #_insert number here_


anselm:
Seems like exactly the right thing to do - doesn't break anything for me.

anselm (approved):

PR #10583, opened 2024-07-11
[IR-3062] fmt: Flatten the `ClientInputSystem.execute` arrow pattern

heysokam:
## Summary
Simple refactor of the ClientInputSystem.execute function, to flatten its Arrow Anti-pattern shape, by using:
- Clause Guards
- Composed Functions pattern

**Process:**
Divide the code into functions that perform only a single, identifiable task. 

Keep all of the operations of each function at the same level of abstraction.The Composed Functions pattern will move nested operations into a separate function that is called from the parent function. 

The goal of this process it to create small functions that are more easily debug-able, and can be tested by themselves in isolation, without being affected by the processes applied by all other blocks of code (which are also converted into separate functions themselves).

## Breaking Changes
There must be none. QA process should ensure this is the case.

## References
Please reference https://tsu.atlassian.net/browse/IR-3062.

## QA Steps
- [x] Review the names of the new functions
- [x] Review that code organization/structure does not affect its goal
- [x] Ensure that changes do not affect the logic/behavior at all

### DIFF
--- a/packages/spatial/src/input/systems/ClientInputSystem.tsx
+++ b/packages/spatial/src/input/systems/ClientInputSystem.tsx
@@ -250,7 +250,7 @@ const execute = () => {
   for (const eid of xrSpaces()) {
     const space = getComponent(eid, XRSpaceComponent)
     const pose = xrFrame?.getPose(space.space, space.baseSpace)
-    if (!pose) continue // @NOTE: Clause Guard. This was nested as   if (pose) { ... }
+    if (!pose) continue // @note Clause Guard. This was nested as   if (pose) { ... }
     TransformComponent.position.x[eid] = pose.transform.position.x
     TransformComponent.position.y[eid] = pose.transform.position.y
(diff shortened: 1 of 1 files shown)


SamMazerIR (commented):
looks good, will need new function names to replace TODO names but overall good cleanup. I'm not sure I agree with un-nesting lines 253-261, as it makes pose related code clearly isolated (in case more code were added within the for each loop later)

heysokam:
> @SamMazerIR
> I'm not sure I agree with un-nesting lines 253-261, as it makes pose related code clearly isolated (in case more code were added within the for each loop later)

Could there be a reason why code for XRSpace would be added, for cases where `pose` would be falsy?
Also, would there be a reason why such code couldn't be its own block, in order to separate the concerns of that block of code from the others?

Asking because it feels like this mindset is what created the arrow to begin with :thinking:

SamMazerIR (approved):

PR #10584, opened 2024-07-11
IR-3065: hide assets panel categories that are empty

dinomut1:
Closes https://tsu.atlassian.net/browse/IR-3065


the-reality-engineer:
Maybe we should put tags into a separate table so that we can count them easily, instead of fetching a huge number of resources?

DanielBelmes:
copy pasting my slack comment "Can we not just remove the tabs we know are currently void of assets? I worry about adding a potentially huge query to hide some ui elements if we're not using the query results for anything else. Can we just identify some important categories that encompass all the assets?"

Rezmason (approved):

DanielBelmes (changes requested):
Sadly looks like when you click on the nested category it removes the brother/sister categories.

DanielBelmes (approved):

HexaField:
We really need to either generate the categories at runtime or create an API in projects to specify them.

PR #10597, opened 2024-07-12
fix(ff): validate undefined value [IR-2888]

mikeplascdev:
## Summary

- change the value of the mutation when the result is a success
- Return false if the get FF's is in progress

## References
closes #_insert number here_

## Evidence
ff

## Jira
[IR-2888](https://tsu.atlassian.net/browse/IR-2888)

[IR-2888]: https://tsu.atlassian.net/browse/IR-2888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/engine/src/FeatureFlagsState.tsx
+++ b/packages/engine/src/FeatureFlagsState.tsx
@@ -26,6 +26,7 @@ Ethereal Engine. All Rights Reserved.
 import { featureFlagSettingPath } from '@etherealengine/common/src/schema.type.module'
 import { defineState, getMutableState, useHookstate } from '@etherealengine/hyperflux/functions/StateFunctions'
 import { useFind } from '@etherealengine/spatial/src/common/functions/FeathersHooks'
+import { isUndefined } from 'lodash'
 import { useEffect } from 'react'
 
 export const FeatureFlagsState = defineState({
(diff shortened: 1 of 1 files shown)


HexaField:
@cla-bot check

packages/engine/src/FeatureFlagsState.tsx
```diff
@@ -37,7 +38,7 @@ export const FeatureFlagsState = defineState({
   },
   useEnabled(flagName: string) {
     const state = useHookstate(getMutableState(FeatureFlagsState)[flagName]).value
-    return typeof state === 'boolean' ? state : true
+    return typeof state === 'boolean' ? state : !isUndefined(state)
```
hanzlamateen: Instead of this you can simply make it `false`.

Though @HexaField with this change features will be `off` by default in contrast to how they currently are `on` by default.

hanzlamateen (changes requested):

packages/engine/src/FeatureFlagsState.tsx
HexaField: we need to automatically seed all flags if we are going to turn the off by default. local dev should have all flags on by default, otherwise it'll be a massive hassle any time someone reinits.

hanzlamateen: @mikeplascdev the point of concern in the issue was, maybe hide the feature as data regarding feature flag is being fetched from backend. And then if an item is not found in feature flags data then it means its enabled unless its explicitly turned off.

HexaField: Oh I see that there could be race conditions or things appearing momentarily until the data is fetched. In that case yes we should go ahead and assume they are disabled by default, (else condition false) and add seeding for local dev

HexaField: We also do want to replace FeatureFlagState with in-place useFind with the flagname in the query, since feature flags will paginate and we are getting close to having 10 settings, so we should do that soon.

hanzlamateen: @mikeplascdev marked this PR as draft as above needs to be addressed

mikeplascdev:
@hanzlamateen 
cc: @HexaField @jonathan-casarrubias 
Closed because we created a new [Pull Request](https://github.com/theinfinitereality/eepro-multitenancy/pull/353) in the [eepro-multitenancy](https://github.com/theinfinitereality/eepro-multitenancy) project with the new feature flag hook

PR #10599, opened 2024-07-12
Hardened instanceserver initialization

barankyle:
## Summary

Instanceserver initialization was making new instance records if it couldn't find one that matched the IP address and location/channel ID provided by the connecting user. This could be abused by a malicious actor. Now, the instanceserver just checks if there is a record for itself, and does nothing if it can't find one.

Moved the initial authorizeUserToJoinServer call earlier so that nothing else happens if the user shouldn't be on the server in the first place. loadEngine is not called if the server isn't initialized for any reason. User doesn't join the instanceserver if not authorized to do so. Cleaned up some duplicate logic.

Resolves IR-2671

## References
closes #_insert number here_

### DIFF
--- a/packages/instanceserver/src/InstanceServerState.ts
+++ b/packages/instanceserver/src/InstanceServerState.ts
@@ -41,6 +41,7 @@ export const InstanceServerState = defineState({
     ready: false,
     instance: null! as InstanceType,
     isMediaInstance: false,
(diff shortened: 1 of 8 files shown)


barankyle:
I see the newly-failing test, I'm updating it so that it'll succeed.

HexaField (approved):

PR #10603, opened 2024-07-14
chore(ir-engine): implemented multilogger optional modifier hook

jonathan-casarrubias:
## Summary
Introduced a decoupled mechanism that enables to optionally modify logs parameters payloads sent to the API, this is useful when additional parameters other than the default msg parameter are required, by centralizing logic that can possibly be required within different components, this approach enhances the development experience by introducing an optional modifier hook that can be wired up to multilogger instances that require these additional parameters.

This implementation is backwards compatible since it won't affect multilogger instances previously created and optimized to avoid running the modifier on every multilogger instance not requiring these parameters at all.

## Subtasks Checklist
- Implemented multilogger optional modifier hook (modify params object)
- Implemented clientContextParams functions that collects contextual parameters from store
- Wired up multilogger optional hook with clientContextParams when logging app crashes in ErrorBoundary's error handler callback.

## Breaking Changes
None

### DIFF
--- a/packages/client-core/src/util/contextParams.ts
+++ b/packages/client-core/src/util/contextParams.ts
@@ -23,28 +23,27 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
-import React from 'react'
+import { LogParamsObject } from '@etherealengine/common/src/logger'
+import { getState } from '@etherealengine/hyperflux'
+import { LocationState } from '../social/services/LocationService'
 
-import multiLogger from '@etherealengine/common/src/logger'
-import { createErrorBoundary } from '@etherealengine/common/src/utils/createErrorBoundary'
-import { clientContextParams } from '../../util/contextParams'
+/**
+ * @function clientContextParams
+ * @description This function will collect contextual parameters
+ * from url's query params
+ */
+export function clientContextParams(params: LogParamsObject) {
+  const locationState = getState(LocationState)
+  /*
+  console.log('IR> location state', locationState.currentLocation.location)
 
-const logger = multiLogger.child({ component: 'client-core:system-crash', modifier: clientContextParams })
+  console.log('IR> location_id', locationState.currentLocation.location.id)
 
-const ErrorBoundary = createErrorBoundary(
-  function error(props, error?: Error) {
-    if (error) {
-      return (
-        <div className="error-screen">
-          <h2 style={{ fontSize: '100%', fontWeight: 'normal' }}>An error has occured</h2>
-          <h4 style={{ fontSize: '100%', fontWeight: 'normal' }}>{error.message}</h4>
-        </div>
-      )
-    } else {
-      return <React.Fragment>{props.children}</React.Fragment>
-    }
-  },
-  (error) => logger.error(error)
-)
-
-export default ErrorBoundary
+  console.log('IR> project_id', locationState.currentLocation.location.projectId)
+*/
+  return {
+    ...params,
+    location_id: locationState.currentLocation.location.id,
+    project_id: locationState.currentLocation.location.projectId
+  }
+}
--- a/packages/client-core/src/world/Location.tsx
+++ b/packages/client-core/src/world/Location.tsx
@@ -23,7 +23,7 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
-import React from 'react'
+import React, { useEffect } from 'react'
 import { useParams } from 'react-router-dom'
 
 import { LocationIcons } from '@etherealengine/client-core/src/components/LocationIcons'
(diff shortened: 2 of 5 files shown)


packages/client-core/src/util/contextParams.ts
```diff
@@ -0,0 +1,68 @@
...
+export function clientContextParams(params: LogParamsObject) {
+  const authState = getState(AuthState)
+  const locationState = getState(LocationState)
+  const editorState = getState(EditorState)
+
+  return {
+    ...params,
+    userId: authState.user?.id,
+    url: window.location.href,
+    locationId: locationState.currentLocation.location.id,
```
HexaField: i feel like getting all these states is a bit unnecessary, since all the data (aside from userId) is going to be derivable from the URL anyway

HexaField (changes requested):

HexaField:
@cla-bot check

packages/client-core/src/util/contextParams.ts
hanzlamateen: In backend to which these logs are sent is automatically extracting userId based on the user who is making the request from params.

So it can be skipped here.

https://github.com/EtherealEngine/etherealengine/blob/2b0535b1de34d7a624f323ae74e609d8a22aa9c0/packages/server-core/src/cluster/logs-api/logs-api.class.ts#L49

HexaField: @jonathan-casarrubias yeah so i think URL is sufficient

hanzlamateen (approved):
Looks good to me.

@HexaField can you give it a second look

packages/client-core/src/util/contextParams.ts
```diff
@@ -0,0 +1,40 @@
...
+
+import { LogParamsObject } from '@etherealengine/common/src/logger'
+
+/**
+ * @function clientContextParams
+ * @description This function will collect contextual parameters
+ * from url's query params
+ */
+export function clientContextParams(params: LogParamsObject) {
+  const urlParams = new URLSearchParams(window.location.search)
```
HexaField: why not the entire URL href?

HexaField (changes requested):

packages/client-core/src/util/contextParams.ts
hanzlamateen: Yes, I agree we will need to have entire URL. Because subdomain will tell which account the location was running for.

jonathan-casarrubias: Well, I'm just trying to avoid the necessity to parse or to decode params down the road, if I don't do it I would be passing the responsibility ball to someone else, this time I'm passing the entire thing but as params which would be much more helpful for Balasz than the entire URL as a string.

HexaField: we should put these on a named object then, rather than polluting the namespace of the object returned

HexaField: and we should include the domain name and URL pathname too

packages/client-core/src/util/contextParams.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  return {
+    ...params,
+    ...Object.fromEntries(urlParams),
```
HexaField: we should put tihs under a `search` property, as if a param is named `path` or `host` it will be overrided by the following values

HexaField (changes requested):

packages/client-core/src/util/contextParams.ts
jonathan-casarrubias: I see your point, though putting into a search property wouldn't address Balazs' requirements, so I guess I'll simply use the needed params, at the end he did mention that if a param is not available is better for him sending the param empty rather than not sending it at all, kind of { project_id: "" }. Will resubmit with these changes

HexaField (approved):

packages/client-core/src/world/Location.tsx
```diff
@@ -38,11 +38,15 @@ import './LocationModule'
...
 
+import multiLogger from '@etherealengine/common/src/logger'
 import { StyledEngineProvider } from '@mui/material/styles'
 import { LoadingCircle } from '../components/LoadingCircle'
 import { useLoadEngineWithScene, useNetwork } from '../components/World/EngineHooks'
 import { LocationService } from '../social/services/LocationService'
 import { LoadingUISystemState } from '../systems/LoadingUISystem'
+import { clientContextParams } from '../util/contextParams'
+
+const logger = multiLogger.child({ component: 'system:location ', modifier: clientContextParams })
```
DanielBelmes, line 49: Just because I noticed while managing weird conflict errors when updating MT-INT. The component string has a space at the end here.

hanzlamateen: I fixed this, and another one in https://github.com/EtherealEngine/etherealengine/pull/10759/files

PR #10605, opened 2024-07-15
Fix material cleanup logic

AidanCaruso:
## References
closes #_insert number here_


SYBIOTE (commented):
looks good, we check for a material instance this time, and only then assign fallback material

dinomut1 (approved):

PR #10606, opened 2024-07-15
fix fog setting update

JT00y:
## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/renderer/components/FogSettingsComponent.ts
+++ b/packages/spatial/src/renderer/components/FogSettingsComponent.ts
@@ -132,12 +132,12 @@ export const FogSettingsComponent = defineComponent({
 
     useEffect(() => {
       const fogComponent = getOptionalComponent(entity, FogComponent)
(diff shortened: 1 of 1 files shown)


packages/spatial/src/renderer/components/FogSettingsComponent.ts
```diff
@@ -132,12 +132,12 @@ export const FogSettingsComponent = defineComponent({
 
     useEffect(() => {
       const fogComponent = getOptionalComponent(entity, FogComponent)
-      if (fogComponent && fog.type.value !== FogType.Linear) (fogComponent as Fog).near = fog.near.value
+      if (fogComponent && fog.type.value == FogType.Linear) (fogComponent as Fog).near = fog.near.value
```
dinomut1: why not just remove the second part of this conditional entirely?

dinomut1: So it's just `if (fogComponent)`

dinomut1 (changes requested):

dinomut1 (approved):

PR #10610, opened 2024-07-15
Disabling screenshare if it's not supported.

barankyle:
## Summary

Mobile browsers do not support navigator.mediaDevices.getDisplayMedia, which is how one retrieves a display for screensharing over WebRTC. getDisplayMedia is undefined on all mobile browsers (or should be - older versions of most browsers it was exposed, but returned a NotAllowedError). Now hiding the screenshare button if isMobile is true or getDisplayMedia is undefined.

Resolves IR-3136

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/components/MediaIconsBox/index.tsx
+++ b/packages/client-core/src/components/MediaIconsBox/index.tsx
@@ -55,6 +55,7 @@ import IconButtonWithTooltip from '@etherealengine/ui/src/primitives/mui/IconBut
 
 import { FeatureFlags } from '@etherealengine/common/src/constants/FeatureFlags'
 import { FeatureFlagsState } from '@etherealengine/engine/src/FeatureFlagsState'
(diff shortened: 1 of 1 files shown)


HexaField (approved):

packages/client-core/src/components/MediaIconsBox/index.tsx
```diff
@@ -204,7 +205,11 @@ export const MediaIconsBox = () => {
           )}
         </>
       ) : null}
-      {screenshareEnabled && mediaNetworkReady && mediaNetworkState?.ready.value ? (
+      {!isMobile &&
+      navigator.mediaDevices.getDisplayMedia != null &&
```
HexaField: can we do `!navigator.mediaDevices.getDisplayMedia` or `typeof navigator.mediaDevices.getDisplayMedia === 'undefined'` instead?

HexaField (changes requested):

HexaField (approved):

PR #10612, opened 2024-07-15
IR-2985-Point-Click-doesnt-work-with-Assets-Panel

MichaelEstes:
## References
closes #_insert number here_


packages/ui/src/components/editor/panels/Assets/container/index.tsx
```diff
@@ -70,7 +79,13 @@ type Category = {
   depth: number
 }
 
-const AssetsPreviewContext = createContext({ onAssetSelectionChanged: (props: AssetSelectionChangePropsType) => {} })
+const AssetsPreviewContext = hookstate({
```
HexaField: any reason we arent using hyperflux here? or making it a scoped useHookstate?

HexaField (changes requested):

packages/ui/src/components/editor/panels/Assets/container/index.tsx
MichaelEstes: Nope, switched it to a hookstate

HexaField (changes requested):
This doesn't seem to be working for me. I opened the asset panel, switch to Click & Place mode, but i was not able to select an asset and place, as if it was still in drag and drop mode. When i switched to the file view, i was able to click and place, and when i then switched to assets the old file selected from the file view was still click and placing.

HexaField:
> This doesn't seem to be working for me. I opened the asset panel, switch to Click & Place mode, but i was not able to select an asset and place, as if it was still in drag and drop mode. When i switched to the file view, i was able to click and place, and when i then switched to assets the old file selected from the file view was still click and placing.

Oh! Actually, it just looks like `.glb` files do not trigger the click and place, i was able to select .fbx and .gltf files and place those from the asset panel

HexaField:
> > This doesn't seem to be working for me. I opened the asset panel, switch to Click & Place mode, but i was not able to select an asset and place, as if it was still in drag and drop mode. When i switched to the file view, i was able to click and place, and when i then switched to assets the old file selected from the file view was still click and placing.
> 
> Oh! Actually, it just looks like `.glb` files do not trigger the click and place, i was able to select .fbx and .gltf files and place those from the asset panel

Actually, on further inspection, it just seems like certain files are not changing the active asset to click and place. Everything else I said seems downstream from that.

MichaelEstes:
> > > This doesn't seem to be working for me. I opened the asset panel, switch to Click & Place mode, but i was not able to select an asset and place, as if it was still in drag and drop mode. When i switched to the file view, i was able to click and place, and when i then switched to assets the old file selected from the file view was still click and placing.
> > 
> > 
> > Oh! Actually, it just looks like `.glb` files do not trigger the click and place, i was able to select .fbx and .gltf files and place those from the asset panel
> 
> Actually, on further inspection, it just seems like certain files are not changing the active asset to click and place. Everything else I said seems downstream from that.

Not all assets are valid for click placement, I've added some warning messaging around that. GLBs not working was a bug though, those should work now. I have no idea how you got an FBX file to work.

HexaField (approved):

PR #10616, opened 2024-07-16
IR-3133: Changes to fix import for injected feature flags

hanzlamateen:
## Summary
https://tsu.atlassian.net/browse/IR-3133

## References
closes #_insert number here_


hanzlamateen:
Verified using `nm run local` the issue was reproducible before, and this fixed it

PR #10625, opened 2024-07-16
Drop shadows

MichaelEstes:
## References
closes #_insert number here_


SYBIOTE (approved):

packages/engine/src/scene/systems/ShadowSystem.tsx
```diff
@@ -482,3 +469,16 @@ export const ShadowSystem = defineSystem({
   execute,
   reactor
 })
+
+export const DropShadowSystem = defineSystem({
+  uuid: 'ee.engine.DropShadowSystem',
+  insert: { after: TransformSystem },
```
HexaField, line 475: this being after the TransformSystem means matrices are out of date. if we need another entity's matrix to be updated, we should use the ComputedTransformComponent to handle that relationship if we can

PR #10626, opened 2024-07-16
IR-3118: Material thumbnail generation

dinomut1:
Closes https://tsu.atlassian.net/browse/IR-3118


AidanCaruso:
Works, but I am seeing an error in the material importer extension that can prevent thumbnail generation of subsequent material gltfs, to reproduce create a new material and save it before reloading the /mat directory. Should be a simple fix though. 
image

AidanCaruso (approved):
Added a check to the importer extension to stop the error which would occasionally prevent thumbnail generation. Looking good now.

PR #10627, opened 2024-07-16
IR-3186/3185/3184/3220 Fix material object reference loss, material library rerendering, fallback and UI improvements 

AidanCaruso:
- FIx material library rerendering every frame on material selection change, preventing a ton of lag
- Fix threejs material object references sometimes being lost by UUID lookup when setting a material to a preexisting UUID
- Return fallback material when a material state entity does not exist for a given UUID
- Prevent fallback material populating into material library
- Materials with empty names now fall back to the material's type as a name
- Material preview panel now has a max height to make it easier to see both the currently selected material and the material list

## References
closes #_insert number here_


packages/ui/src/components/editor/panels/Properties/material/index.tsx
```diff
@@ -153,7 +152,7 @@ export function MaterialEditor(props: { materialUUID: EntityUUID }) {
 
   useEffect(() => {
     clearThumbs().then(createThumbnails).then(checkThumbs)
-  }, [materialName, prototypeName])
+  }, [prototypeName])
```
dinomut1, line 155: needs to have currentSelectedMaterial as a dependency as well

packages/ui/src/components/editor/panels/Properties/material/index.tsx
```diff
@@ -187,7 +186,7 @@ export function MaterialEditor(props: { materialUUID: EntityUUID }) {
   useEffect(() => {
     pluginValues.set({})
     pluginParameters.set({})
-  }, [materialName, selectedPlugin])
+  }, [selectedPlugin])
```
dinomut1, line 189: needs currentSelectedMaterial dependency

dinomut1 (changes requested):

dinomut1:
Above changes are already in https://github.com/EtherealEngine/etherealengine/pull/10640 so we can also just merge this in now

AidanCaruso:
@dinomut1 Resolved conflicts with dev, if you mean this is ready to go into dev can you give an approval?

dinomut1 (approved):

PR #10630, opened 2024-07-17
IR-346 Add nonce attribute to apple SSO

muradkhateeb78:
Adding nonce attribute to Apple SSO will add an additional security layer.


muradkhateeb78:
This PR is not needed anymore.

PR #10633, opened 2024-07-17
Fix video freezing on new track

SYBIOTE:
This pull request fixes the issue of video freezing on a new track. The code changes ensure that the track is not changed if the current track is not ended. Additionally, the pull request handles the case where all resources are empty and skips over null paths to move to the next proper track. This resolves the freezing issue and improves the overall functionality of the video player.


HexaField:
@speigg @CITIZENDOT please review

the-reality-engineer (approved):

PR #10635, opened 2024-07-17
updated the highlight 

MbfloydIR:
## Summary
updated the highlight logic to also highlight children of the selected object
increased thickness of the highlight

## References
closes #_insert number here_


packages/spatial/src/renderer/components/HighlightComponent.ts
```diff
@@ -26,23 +26,32 @@ Ethereal Engine. All Rights Reserved.
...
 export const HighlightComponent = defineComponent({ name: 'HighlightComponent' })
 
-const highlightQuery = defineQuery([HighlightComponent, MeshComponent, GroupComponent, VisibleComponent])
+const highlightQuery = defineQuery([HighlightComponent, VisibleComponent])
 
 const execute = () => {
   /** @todo support multiple scenes */
   if (!hasComponent(Engine.instance.viewerEntity, RendererComponent)) return
 
+  let highlightObjects: Object3D[] = []
```
MichaelEstes: If you change this type to a Set<Object3D> it'll remove the need for the ts-ignore below

packages/spatial/src/renderer/components/PostProcessingComponent.tsx
```diff
@@ -135,6 +137,9 @@ const PostProcessingReactor = (props: { entity: Entity; rendererEntity: Entity }
     // //always have the outline effect for the highlight selection
     const outlineEffect = new OutlineEffect(scene as Scene, camera.value as ArrayCamera, getState(HighlightState))
     outlineEffect.selectionLayer = ObjectLayers.HighlightEffect
+    outlineEffect.blur = true
```
HexaField: we should use the values in `HighlightState` instead of hardcoding these, and fix those to be the settings we want

HexaField (changes requested):

packages/spatial/src/renderer/components/PostProcessingComponent.tsx
MbfloydIR: implemented

HexaField (approved):

HexaField (changes requested):
Screenshot from 2024-07-19 08-05-01
we need to check if the effect composer exists or not

packages/spatial/src/renderer/components/HighlightComponent.ts
```diff
@@ -26,23 +26,31 @@ Ethereal Engine. All Rights Reserved.
...
 
-  // @ts-ignore @todo why does typescript freak out here?
-  getComponent(Engine.instance.viewerEntity, RendererComponent).effectComposer?.OutlineEffect?.selection.set(
-    highlightQuery().map((entity) => getComponent(entity, MeshComponent))
-  )
+  let highlightObjects = new Set<Object3D>()
+  for (const entity of highlightQuery()) {
+    traverseEntityNode(entity, (child, index) => {
+      if (!hasComponent(child, MeshComponent)) return
+      if (!hasComponent(child, GroupComponent)) return
```
HexaField, line 48: we need to exclude non visible entities here too

packages/spatial/src/renderer/components/HighlightComponent.ts
```diff
@@ -26,23 +26,31 @@ Ethereal Engine. All Rights Reserved.
...
 
 const execute = () => {
   /** @todo support multiple scenes */
   if (!hasComponent(Engine.instance.viewerEntity, RendererComponent)) return
 
-  // @ts-ignore @todo why does typescript freak out here?
-  getComponent(Engine.instance.viewerEntity, RendererComponent).effectComposer?.OutlineEffect?.selection.set(
-    highlightQuery().map((entity) => getComponent(entity, MeshComponent))
-  )
+  let highlightObjects = new Set<Object3D>()
```
HexaField: this `let` should be a `const`

HexaField (changes requested):

packages/spatial/src/renderer/components/HighlightComponent.ts
MbfloydIR: implemented

packages/spatial/src/renderer/components/HighlightComponent.ts
MbfloydIR: Implemented

HexaField (approved):

PR #10636, opened 2024-07-17
studio: update active dock bg color

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-2965
## References
closes #_insert number here_


packages/editor/src/components/Editor2Container.css
```diff
@@ -31,7 +31,7 @@
 }
 
 .dock-tab.dock-tab-active {
-  background: var(--bg-surface-main) !important;
+  background: #212226 !important;
```
HexaField: if we need a new colour, we should add it to the theme, rather than hardcoding it

HexaField (dismissed):

packages/editor/src/components/Editor2Container.css
achen5671: I updated to use tailwind values

aditya-mitra (approved):
LGTM

PR #10641, opened 2024-07-18
Fixed issues with creating consumers

barankyle:
## Summary

There was a bug when joining an instance that already had producers from other users running - consumers were not created for those producers, and they would never be created since pausing/resuming a producer does not trigger a new creation of consumers if they do not exist.

Creation of consumers is triggered from the presence of producers in MediasoupMediaProducerConsumerState. onUpdatePeers removes producers that are not in MediasoupTransportState. onUpdatePeers was getting received by new clients before networks had been set up and were ready, so it was seeing no transports and removing all producers before their consumers could be created. The fix is to not run onUpdatePeers' logic if the network is not ready.

Added some hardening to resuming consumers to address some crashes seen in testing. For unclear reasons that stopped happening at all, one client being consumed by others that disconnected would trigger the useEffect that handles consumers getting paused or resumed, and the consumers of the now-not-running producers were trying to resume, causing Mediasoup to throw errors about the channel being resumed not existing. Hopefully checking whether the producer that that consumer is consuming exists and is not closed will avoid this.

In onDisconnection, updated local behavior to look for instanceId on instanceServerState. There were disconnection bugs where the instanceId was no longer on the connection object, and the user would not be removed from records.

Fixed visual bug with "user is speaking" green circle triggering for oneself even if one's audio is paused. The issue was that audioStreamPaused is not set to true for one's own audio; instead, that's audioProducerPaused.

Resolves IR-3207 and IR-3210

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/media/PeerMedia.tsx
+++ b/packages/client-core/src/media/PeerMedia.tsx
@@ -162,7 +162,7 @@ export const NetworkProducer = (props: { networkID: InstanceID; producerID: stri
   const networkState = useHookstate(getMutableState(NetworkState).networks[networkID])
 
   useEffect(() => {
-    if (!networkState.ready.value) return
+    if (!networkState.ready?.value) return
 
     const peerID = producerState.peerID.value
     // dont need to request our own consumers
@@ -182,7 +182,7 @@ export const NetworkProducer = (props: { networkID: InstanceID; producerID: stri
         $to: network.hostPeerID
       })
     )
-  }, [networkState.ready.value])
+  }, [networkState.ready?.value])
 
   return null
 }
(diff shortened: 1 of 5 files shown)


packages/client-core/src/media/PeerMedia.tsx
```diff
@@ -182,7 +182,7 @@ export const NetworkProducer = (props: { networkID: InstanceID; producerID: stri
         $to: network.hostPeerID
       })
     )
-  }, [networkState.ready.value])
+  }, [networkState.ready?.value])
```
HexaField, line 185: These ? shouldn't be necessary, the network should always exist if the reactor is mounted

barankyle: I was getting some errors about it not existing during some of the testing I was doing. Maybe it won't happen now that I've fixed the main issue?

HexaField: it can't hurt i suppose!

HexaField (approved):

PR #10642, opened 2024-07-18
IR-3172-Unable-to-change-Image-URL

MichaelEstes:
## Summary
The image url changing bug was not reproducible for me, I'm assuming from the error logs that it got fixed by one of Aidan's MaterialComponent PRs. 

The issue with the image's transform resetting after changing properties I don't think had anything to do with the Image component, but with the pointer leaving the canvas while the image was being dragged which caused the transform to not be committed. This PR adds committing the transform when the pointer is lost.

## References
closes #_insert number here_


HexaField (approved):
nice

PR #10650, opened 2024-07-18
fix popup clipping 

SYBIOTE:
This pull request fixes an overflow issue in the Popup component. The content style of the Popup component has been updated to set the overflow property to visible. This ensures that the content inside the Popup component is not clipped when it exceeds the container's dimensions.


aditya-mitra (approved):
awesome

PR #10651, opened 2024-07-18
IR-3227: Close popover on mousedown instead of a click

CITIZENDOT:
## References
closes #_insert number here_


aditya-mitra (approved):
LGTM

PR #10652, opened 2024-07-18
refactor(IR-3108): replace FeatureFlagState with useFind in react components

kattsushi:
refactor(IR-3108): replace FeatureFlagState with useFind in react components

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/components/MediaIconsBox/index.tsx
+++ b/packages/client-core/src/components/MediaIconsBox/index.tsx
@@ -54,7 +54,7 @@ import Icon from '@etherealengine/ui/src/primitives/mui/Icon'
 import IconButtonWithTooltip from '@etherealengine/ui/src/primitives/mui/IconButtonWithTooltip'
 
 import { FeatureFlags } from '@etherealengine/common/src/constants/FeatureFlags'
-import { FeatureFlagsState } from '@etherealengine/engine/src/FeatureFlagsState'
+import { useFeatureFlags } from '@etherealengine/engine/src/FeatureFlagsHook'
 import { isMobile } from '@etherealengine/spatial/src/common/functions/isMobile'
 import { VrIcon } from '../../common/components/Icons/VrIcon'
 import { RecordingUIState } from '../../systems/ui/RecordingsWidgetUI'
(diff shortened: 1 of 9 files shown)


packages/client-core/src/components/MediaIconsBox/index.tsx
```diff
@@ -99,8 +100,25 @@ export const MediaIconsBox = () => {
...
+        {
+          flagName: FeatureFlags.Client.Menu.MotionCapture
+        },
+        {
+          flagName: FeatureFlags.Client.Menu.AR
+        }
+      ],
+      paginate: false
+    }
+  })
```
aditya-mitra: can we convert all of these hook calls into a `useFeatureFlag` hook?
- this hook in turn will call useFind with the common query
- if there are 2 feature flags it would automatically append the `$or` operator
- it can return an array of results in the order of feature flags sent to the hook params
(this will also reduce the code lines and make feature flag more friendly to be used on the client)


somewhat like this:
```tsx
const [motionCaptureEnabled, arEnabled] = useFeatureFlags([FeatureFlags.Client.Menu.MotionCapture, FeatureFlags.Client.Menu.AR]);
```

aditya-mitra (changes requested):

packages/client-core/src/components/MediaIconsBox/index.tsx
HexaField: yes agreed, this would be good

kattsushi: yes good point applied! thanks @aditya-mitra

packages/engine/src/FeatureFlagsState.tsx
```diff
@@ -24,7 +24,7 @@ Ethereal Engine. All Rights Reserved.
 */
 
 import { featureFlagSettingPath } from '@etherealengine/common/src/schema.type.module'
-import { defineState, getMutableState, useHookstate } from '@etherealengine/hyperflux/functions/StateFunctions'
+import { defineState, getMutableState } from '@etherealengine/hyperflux/functions/StateFunctions'
 import { useFind } from '@etherealengine/spatial/src/common/functions/FeathersHooks'
 import { useEffect } from 'react'
```
hanzlamateen, line 30: I think we need to completely get rid of this file. @HexaField thoughts?

hanzlamateen (commented):
Very nice and clean implementation.

Just left 1 comment. Apart from that it looks good.

packages/engine/src/FeatureFlagsState.tsx
kattsushi: I agree, but some services don't use React components; instead, they use services that still rely on state as a service. Is there another way to make the request without using the `useFind` hook for those cases? If we get rid of that file, we will impact those services in the core of EE.

@HexaField, there's a way to achieve this? make the request `find` without using feathers.js hooks

HexaField: I pushed a commit to remove the scene preview camera flag, since the implementation to filter it from the categories state is creating a race conditions, and i dont think we need it anymore.

as for the hierarchy flag, i will need to talk to @MichaelEstes to see if there is a way we get it reactively

HexaField: I think this is good for now, we should add a deprecation notice to the state and call this PR good

PR #10658, opened 2024-07-18
Remove close button

SYBIOTE:
This pull request removes the close button from the ResourceFile component. The close button was previously included in the context menu, but it is no longer needed. This change simplifies the UI and improves the user experience.


aditya-mitra (approved):
good

PR #10659, opened 2024-07-18
Make pointer in gizmo tools highlightable

SYBIOTE:
This pull request adds the functionality to make the pointer in gizmo tools highlightable. Previously, the pointer was not visually distinguishable when selected. With this change, the pointer will now have a highlighted background color when selected.


aditya-mitra:
seems already implemented

PR #10661, opened 2024-07-18
IR-3023 Fix fly camera system inputs applying to all cameras

AidanCaruso:
## References
closes https://tsu.atlassian.net/browse/IR-3023


packages/spatial/src/input/systems/FlyControlSystem.ts
```diff
@@ -87,7 +87,8 @@ const execute = () => {
 
   /** Since we have nothing that specifies whether we should use orbit/fly controls or not, just tie it to the camera orbit component for the studio */
   for (const entity of cameraQuery()) {
-    const inputPointerEntity = InputPointerComponent.getPointersForCamera(entity)
+    const inputPointerEntity = InputPointerComponent.getPointersForCamera(entity)[0]
```
the-reality-engineer: If you just get the merged buttons from the input sources associated with each camera, it should ensure that only the cameras that are intersecting with the pointer receive button input. 

```ts
const inputPointerEntities = InputPointerComponent.getPointersForCamera(entity)
const buttons = InputComponent.getMergedButtonsForInputSources(inputPointerEntities)
```

the-reality-engineer (dismissed):

DanielBelmes (approved):

PR #10664, opened 2024-07-19
Changes to NotificationService

MoizAdnan:
Changes to NotificationService:
- Returns enqueued snackbar key
- Adds option for persisting notifications
- Adds option to close a notification with snackbar key
- Adds ReactNode as message type


packages/spatial/src/common/functions/FeathersHooks.tsx
```diff
@@ -369,6 +370,15 @@ export function useRealtime(
     const handlePatched = (data: any) => refetch(data, 'patched')
     const handleRemoved = (data: any) => refetch(data, 'removed')
 
+    if (customEvent) {
```
HexaField: what custom events do we have?

packages/client-core/src/common/services/NotificationService.tsx
```diff
@@ -59,16 +60,21 @@ export const NotificationActions = {
 }
 
 export const NotificationService = {
-  dispatchNotify(message: string, options: NotificationOptions) {
-    if (options?.variant === 'error') {
+  dispatchNotify(message: string | React.ReactNode, options: NotificationOptions) {
```
hanzlamateen:
```suggestion
  dispatchNotify(message: React.ReactNode, options: NotificationOptions) {
```

packages/client-core/src/common/services/NotificationService.tsx
```diff
@@ -59,16 +60,21 @@ export const NotificationActions = {
...
+    if (options?.variant === 'error' && typeof message === 'string') {
```
hanzlamateen: Why do we need `&& typeof message === 'string'`

hanzlamateen (changes requested):

hanzlamateen (approved):

PR #10669, opened 2024-07-19
Revert "Fixed a bug where peer video would not be hidden."

barankyle:
Reverts EtherealEngine/etherealengine#10660

Overlooked that this imports something from client-core in the network package, and client-core is not present on the backend.


DanielBelmes (approved):

HexaField:
Oh boy... we need to make sure we have a check to catch this, maybe a separate CICD sub-workflow that deletes the client packages and runs the instance server? @DanielBelmes @barankyle

barankyle:
Yeah, given how often it happens, it would be good to have something to catch it before this gets merged in.

HexaField:
https://github.com/EtherealEngine/etherealengine/pull/10678

PR #10673, opened 2024-07-19
IR-2940 Material library nodes show materials by source scene & selected/all material node scoping

AidanCaruso:
## References
closes #_insert number here_


packages/ecs/src/UUIDComponent.ts
```diff
@@ -42,6 +42,9 @@ export const UUIDComponent = defineComponent({
 
     if (component.value === uuid) return
 
+    // sanitize uui
+    uuid = uuid.replace(/[^0-9a-zA-Z-]/gi, '') as EntityUUID
```
HexaField, line 44: this logic needs to come before the check on line 43

HexaField (changes requested):

HexaField (changes requested):
I think we should discuss UUID sanitization a bit before we decide if this is exactly what we want

PR #10674, opened 2024-07-19
Polish model properties UI and move gltf transform prop to UI package

SYBIOTE:
This pull request includes the following changes:

- Polish the UI for model properties

- Move the gltf transform property to the UI package

These changes improve the user interface and organization of the code.


aditya-mitra:
we need some more precision to match design - https://www.figma.com/design/DXwFHGa5g0peqqSjyjpxQz/Studio-Design?node-id=4261-8275&t=w8LRHn3vB4HwrhrV-0

image

aditya-mitra (dismissed):
we need some more precision to match design - https://www.figma.com/design/DXwFHGa5g0peqqSjyjpxQz/Studio-Design?node-id=4261-8275&t=w8LRHn3vB4HwrhrV-0

image

SYBIOTE:
The existing primitives cannot accommodate the design, we would have to make changes to the primitives themselves, is that fine?
it will break the primitives we have built around the rest of the components

AidanCaruso (approved):

PR #10675, opened 2024-07-19
Fix special character search

SYBIOTE:
This pull request fixes an issue with the special character search functionality. Previously, when searching for special characters, the search would fail and throw an error. This PR adds a try-catch block to handle special characters in the search value and adjusts the regular expression pattern to escape special characters. Now, the search functionality works correctly even with special characters in the search query.


packages/ui/src/components/editor/panels/Hierarchy/container/index.tsx
```diff
@@ -145,11 +145,16 @@ function HierarchyPanelContents(props: { sceneURL: string; rootEntityUUID: Entit
 
   const searchedNodes: HierarchyTreeNodeType[] = []
   if (searchHierarchy.value.length > 0) {
-    const condition = new RegExp(searchHierarchy.value.toLowerCase())
-    entityHierarchy.value.forEach((node) => {
-      if (node.entity && condition.test(getComponent(node.entity, NameComponent)?.toLowerCase() ?? ''))
-        searchedNodes.push(node)
-    })
+    try {
+      const adjustedSearchValue = searchHierarchy.value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
```
CITIZENDOT: Can you add this regex to this module: `packages/common/src/regex/index.ts`?

And few test cases here: `packages/common/tests/regex.test.ts` (ChatGPT can generate them)

SYBIOTE: done

packages/common/tests/regex.test.ts
```diff
@@ -86,6 +86,52 @@ describe('regex.test', () => {
...
+        'path/to/file',
+        'back\\slash',
+        'pipe|symbol',
+        'question?mark',
+        'asterisk*char',
+        'control\0char',
+        'another\ncontrol'
+      ]
+      inValidHeirarchySearchs.forEach((filename) => {
+        assert.ok(INVALID_FILENAME_REGEX.test(filename), `Expected '${filename}' to be invalid`)
```
CITIZENDOT: I think you meant to use  `VALID_HEIRACHY_SEARCH_REGEX`?

packages/common/tests/regex.test.ts
```diff
@@ -86,6 +86,52 @@ describe('regex.test', () => {
...
+      })
+    })
+
+    const escapeSpecialChars = (input) => {
+      return input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
```
CITIZENDOT: Use `VALID_HEIRACHY_SEARCH_REGEX` here.

CITIZENDOT (changes requested):

packages/common/tests/regex.test.ts
SYBIOTE: The regex is for validating and fixing any hierarchy searches

packages/common/tests/regex.test.ts
SYBIOTE: done

packages/common/tests/regex.test.ts
SYBIOTE: my bad, fixed it, nice catch

CITIZENDOT (approved):

PR #10677, opened 2024-07-19
Fix video volume issue and progress bar stats

SYBIOTE:
This pull request fixes the issue with the video volume and updates the progress bar to show accurate statistics.


packages/ui/src/components/editor/properties/video/index.tsx
```diff
@@ -99,7 +105,11 @@ export const VideoNodeEditor: EditorComponentType = (props) => {
       description={t('editor:properties.video.description')}
       icon={<VideoNodeEditor.iconComponent />}
     >
-      <ProgressBar value={5} paused={false} totalTime={100} />
+      <ProgressBar
```
HexaField: progress bare doesnt update since this data is not reactive

HexaField (changes requested):

packages/ui/src/components/editor/properties/video/index.tsx
SYBIOTE: Yes, that's true, I still left it in as we can obtain the duration of the Clip from this. Ill remove it nonetheless

packages/ui/src/components/editor/input/Progress/index.tsx
```diff
@@ -34,10 +34,11 @@ export interface ProgressBarProps extends ProgressProps {
 }
 
 export default function ProgressBar({ value, paused, totalTime, ...rest }: ProgressBarProps) {
+  console.log('DEBUG value', value)
```
HexaField: remove console log

HexaField (dismissed):

packages/ui/src/components/editor/input/Progress/index.tsx
SYBIOTE: Done

AidanCaruso:
@SYBIOTE I'll give an approving review because this is out of scope, but we really need to fix the audio seemingly starting at 1 in a range of 1-100, where 1 is full volume and 100 is subwoofer-breaking. Shown in the video - warning, it's really loud.

https://github.com/user-attachments/assets/019cf33a-cff9-4356-b7dc-8fc7a0043698

AidanCaruso (approved):

SYBIOTE:
oh, cool ill fix it now

PR #10682, opened 2024-07-21
Fix Server-Core Project Service Tests

HexaField:
## Summary

Had intermittent failures in the test, largely because simple-git usage was not async, but also because we were persisting data between tests

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/projects/project/project.test.ts
+++ b/packages/server-core/src/projects/project/project.test.ts
@@ -63,12 +63,10 @@ describe('project.test', () => {
     }
   })
 
-  before(async () => {
+  beforeEach(async () => {
     app = createFeathersKoaApp()
     await app.setup()
-  })
 
-  before(async () => {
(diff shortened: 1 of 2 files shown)


packages/server-core/src/projects/project/project.test.ts
```diff
@@ -63,12 +63,10 @@ describe('project.test', () => {
     }
   })
 
-  before(async () => {
+  beforeEach(async () => {
     app = createFeathersKoaApp()
     await app.setup()
-  })
```
aditya-mitra, line 69: do we need to setup and destroy engine after each test case?

aditya-mitra (approved):
looks good

packages/server-core/src/projects/project/project.test.ts
HexaField: I think it's a good practice to ensure all state is wiped.

PR #10689, opened 2024-07-22
scene name validation and submission

aditya-mitra:
## Summary

allow scenes names with >= 4 characters and disable the submit button until the new scene name is different from old

## References

https://tsu.atlassian.net/browse/IR-2972


packages/ui/src/components/editor/panels/Scenes/modals/RenameScene.tsx
```diff
@@ -62,6 +62,7 @@ export default function RenameSceneModal({ sceneName, refetch, scene }: Props) {
       className="w-[50vw] max-w-2xl"
       onSubmit={handleSubmit}
       onClose={PopoverState.hidePopupover}
+      submitButtonDisabled={newSceneName.value === sceneName}
```
HexaField: let's also validate the name for disabling submit
Screenshot from 2024-07-23 12-32-00

HexaField (dismissed):

packages/ui/src/components/editor/panels/Scenes/modals/RenameScene.tsx
aditya-mitra: implemented
image

packages/common/src/utils/validateSceneName.ts
```diff
@@ -31,7 +31,7 @@ import {
 
 export default function isValidSceneName(sceneName: string) {
   return (
-    sceneName.length >= 3 &&
+    sceneName.length >= 4 &&
```
CITIZENDOT: I was thinking, we can enforce all four of this conditions in the `VALID_SCENE_NAME_REGEX` itself.

`/^[a-zA-Z0-9][a-zA-Z0-9-]{2,62}[a-zA-Z0-9_\-]$/` (Updated `{1,62}` to `{2,62}`)

This checks all the conditions (allowed length, allowes only valid characters).

CITIZENDOT (changes requested):

packages/common/src/utils/validateSceneName.ts
aditya-mitra: nice. implemented.

CITIZENDOT (approved):

PR #10699, opened 2024-07-22
chore(ir-engine): implement analytics events

jonathan-casarrubias:
## Summary

https://docs.google.com/spreadsheets/d/1rt3hebPrnL7WZXfvQ2NVPBA0ImNlyBEOxsJCKyCErqE/edit?gid=0#gid=0

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/components/Fullscreen/index.tsx
+++ b/packages/client-core/src/components/Fullscreen/index.tsx
@@ -30,8 +30,12 @@ import { AudioEffectPlayer } from '@etherealengine/engine/src/audio/systems/Medi
 import Icon from '@etherealengine/ui/src/primitives/mui/Icon'
 import IconButtonWithTooltip from '@etherealengine/ui/src/primitives/mui/IconButtonWithTooltip'
 
+import multiLogger from '@etherealengine/common/src/logger'
+import { clientContextParams } from '../../util/contextParams'
 import { useShelfStyles } from '../Shelves/useShelfStyles'
 import styles from './index.module.scss'
+
+const logger = multiLogger.child({ component: 'client-core:FullScreen', modifier: clientContextParams })
 
 export const Fullscreen = () => {
   const { t } = useTranslation()
@@ -42,6 +46,7 @@ export const Fullscreen = () => {
     setFullScreenActive(input)
     if (input) document.body.requestFullscreen()
     else document.exitFullscreen()
+    logger.info({ event_name: 'view_fullscreen', event_value: input })
   }
 
   return (
--- a/packages/client-core/src/components/MediaIconsBox/index.tsx
+++ b/packages/client-core/src/components/MediaIconsBox/index.tsx
@@ -34,7 +34,6 @@ import {
   toggleScreenshare,
   toggleWebcamPaused
 } from '@etherealengine/client-core/src/transports/SocketWebRTCClientFunctions'
-import logger from '@etherealengine/common/src/logger'
 import { Engine, defineQuery, getOptionalComponent } from '@etherealengine/ecs'
 import { AudioEffectPlayer } from '@etherealengine/engine/src/audio/systems/MediaSystem'
 import {
@@ -53,6 +52,7 @@ import Icon from '@etherealengine/ui/src/primitives/mui/Icon'
 import IconButtonWithTooltip from '@etherealengine/ui/src/primitives/mui/IconButtonWithTooltip'
 
 import { FeatureFlags } from '@etherealengine/common/src/constants/FeatureFlags'
+import multiLogger from '@etherealengine/common/src/logger'
 import { SceneSettingsComponent } from '@etherealengine/engine/src/scene/components/SceneSettingsComponent'
 import useFeatureFlags from '@etherealengine/engine/src/useFeatureFlags'
 import { isMobile } from '@etherealengine/spatial/src/common/functions/isMobile'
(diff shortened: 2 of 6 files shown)


kattsushi (approved):
LGTM

packages/ui/src/pages/Chat/chat.tsx
```diff
@@ -36,10 +36,14 @@ import { AuthService } from '@etherealengine/client-core/src/user/services/AuthS
 import '@etherealengine/engine/src/EngineModule'
 
 import { LocationService } from '@etherealengine/client-core/src/social/services/LocationService'
+import { clientContextParams } from '@etherealengine/client-core/src/util/contextParams'
+import multiLogger from '@etherealengine/common/src/logger'
 import { getMutableState } from '@etherealengine/hyperflux'
 import { NetworkState } from '@etherealengine/network'
 import { loadEngineInjection } from '@etherealengine/projects/loadEngineInjection'
 
+const logger = multiLogger.child({ component: 'engine:ecs:ProfileMenu', modifier: clientContextParams })
```
hanzlamateen: I dont think `engine:ecs:ProfileMenu` is the correct value for `component` prop here.
```suggestion
const logger = multiLogger.child({ component: 'ui:chat:chat', modifier: clientContextParams })
```

packages/client-core/src/components/MediaIconsBox/index.tsx
```diff
@@ -53,15 +53,22 @@ import Icon from '@etherealengine/ui/src/primitives/mui/Icon'
...
 import { VrIcon } from '../../common/components/Icons/VrIcon'
 import { SearchParamState } from '../../common/services/RouterService'
 import { RecordingUIState } from '../../systems/ui/RecordingsWidgetUI'
 import { MediaStreamService, MediaStreamState } from '../../transports/MediaStreams'
+import { clientContextParams } from '../../util/contextParams'
 import { useShelfStyles } from '../Shelves/useShelfStyles'
 import styles from './index.module.scss'
 
+const clogger = multiLogger.child({
+  component: 'client-core:SocketWebRTCClientFunctions',
```
hanzlamateen: This value for `component` prop is not correct
```suggestion
  component: 'client-core:MediaIconsBox',
```

packages/client-core/src/components/Fullscreen/index.tsx
```diff
@@ -30,9 +30,13 @@ import { AudioEffectPlayer } from '@etherealengine/engine/src/audio/systems/Medi
 import Icon from '@etherealengine/ui/src/primitives/mui/Icon'
 import IconButtonWithTooltip from '@etherealengine/ui/src/primitives/mui/IconButtonWithTooltip'
 
+import multiLogger from '@etherealengine/common/src/logger'
+import { clientContextParams } from '../../util/contextParams'
 import { useShelfStyles } from '../Shelves/useShelfStyles'
 import styles from './index.module.scss'
 
+const logger = multiLogger.child({ component: 'engine:FullScreen', modifier: clientContextParams })
```
hanzlamateen:
```suggestion
const logger = multiLogger.child({ component: 'client-core:FullScreen', modifier: clientContextParams })
```

hanzlamateen (changes requested):

packages/client-core/src/components/MediaIconsBox/index.tsx
```diff
@@ -53,15 +53,22 @@ import Icon from '@etherealengine/ui/src/primitives/mui/Icon'
...
 import { isMobile } from '@etherealengine/spatial/src/common/functions/isMobile'
 import { VrIcon } from '../../common/components/Icons/VrIcon'
 import { SearchParamState } from '../../common/services/RouterService'
 import { RecordingUIState } from '../../systems/ui/RecordingsWidgetUI'
 import { MediaStreamService, MediaStreamState } from '../../transports/MediaStreams'
+import { clientContextParams } from '../../util/contextParams'
 import { useShelfStyles } from '../Shelves/useShelfStyles'
 import styles from './index.module.scss'
 
+const clogger = multiLogger.child({
```
hanzlamateen: Instead of `clogger` you can name it `logger` and remove the import of other `logger` that is in this file i.e. `import logger from '@etherealengine/common/src/logger'`

hanzlamateen (changes requested):

packages/client-core/src/components/MediaIconsBox/index.tsx
jonathan-casarrubias: My apologies for the late response, somehow I missed these comments until I wanted to verify why the story isn't in done already.

Anyway, I made most of the change requests excepting this one, the reason I created a second logger instance is because the previously existing one is being used in a bunch of logs that are completely unrelated to the events we want to track for Analytics.

As I have mentioned, the reason I decided not to globally add the context params to every log in favor of the optional modifier hook is because it can be enabled only for the logs requiring to have these params for the sake of avoiding any negative impact into the previous implementations of the multilogger utility, perfect example of what I mean is the issue brought up by Josh mentioning that it became difficult to debug logs with properties having nothing to do with Analytics related logs.

I'm open for other naming suggestions rather than "clogger", but I'm not keen about reusing the previously existing "logger" instance because by doing that it defeats the purpose which is avoiding negative impact everywhere else the multilogger utility is being used.

Cheers,
Jon

hanzlamateen: The thing is `import logger from '@etherealengine/common/src/logger'` is wrong import and there would be no information regarding which `component:` the log is coming from.

jonathan-casarrubias: Yeah that makes sense, I just verified and is not creating an instance using multilogger.child I see what you mean, though that is existing code, it isn't introduced in this PR, since it's being used....
image

I guess the question is, should I remove the import line and where it's being used, or to simply remove the bad import and use the multilogger reference I did import to create the `logger = multilogger.child({ component: '' })` having no contextParams?

I'm usually not keen to introduce changes in PRs that are unrelated to the assignment, but this is a quick change so I have no problem making an exception.

hanzlamateen: I would suggest `remove the bad import and use the multilogger reference`.

Thanks.

hanzlamateen (approved):

PR #10710, opened 2024-07-24
Fix table sorting

CITIZENDOT:
## Summary
Setting `sortable` to `true` makes the column sortable. This feature is already present, but there's a bug that restricts sortable columns to only one. This PR fixes it.

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/src/admin/common/Table.tsx
+++ b/packages/client-core/src/admin/common/Table.tsx
@@ -25,7 +25,7 @@ Ethereal Engine. All Rights Reserved.
 
 import React, { ReactNode, useEffect } from 'react'
 import { useTranslation } from 'react-i18next'
-import { HiArrowSmallDown, HiArrowSmallUp } from 'react-icons/hi2'
+import { HiArrowSmallDown, HiArrowSmallUp, HiArrowsUpDown } from 'react-icons/hi2'
 
 import { ImmutableObject, NO_PROXY, useHookstate } from '@etherealengine/hyperflux'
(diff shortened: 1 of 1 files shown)


packages/client-core/src/admin/common/Table.tsx
```diff
@@ -49,20 +49,16 @@ export interface ITableHeadCell {
...
   const SortArrow = ({ columnId }: { columnId: string | number }) => {
-    if (columnId === orderBy) {
-      if (order === 1) {
-        return <HiArrowSmallUp onClick={() => onRequestSort(columnId, -1)} />
-      }
-      return <HiArrowSmallDown onClick={() => onRequestSort(columnId, 1)} />
+    if (columnId in order && order[columnId] === 1) {
+      return <HiArrowSmallUp onClick={() => onRequestSort(columnId, -1)} />
     }
-    return <HiArrowSmallUp className={'opacity-0 hover:opacity-50'} />
```
aditya-mitra, line 65: this statement was there to let know the user that the column was sortable
only 1 column can be sorted at a time

check the below image

image

in the above image, the user might get the wrong indication that the date column is sorted but only the name column is sorted

aditya-mitra (changes requested):

packages/client-core/src/admin/common/Table.tsx
CITIZENDOT: image
Added different icon for unsorted columns, but sortable.

aditya-mitra (approved):
nice work

PR #10711, opened 2024-07-24
IR-3088-MT-INT-Studio-Link-Component-does-not-redirect-to-the-link

MichaelEstes:
## Summary
Adds a warning message when trying to interact with a scene link in the studio
Adds cleanup functions for intractable component
Fixes race condition with hierarchy panel and click placement when the click placement entity is removed
## References
closes #_insert number here_


SamMazerIR (approved):

SamMazerIR (approved):

packages/editor/src/services/EditorServices.ts
```diff
@@ -86,5 +89,17 @@ export const EditorState = defineState({
     if (!rootEntity) return false
     return !!getState(GLTFModifiedState)[getComponent(rootEntity, SourceComponent)]
   },
-  extension: syncStateWithLocalStorage(['expandedNodes'])
+  extension: syncStateWithLocalStorage(['expandedNodes']),
+  reactor: () => {
+    const linkState = useMutableState(LinkState)
```
HexaField, line 94: ideally this should not be the responsibility if the EditorState, this should be handled in the link opening logic in client-core

packages/editor/src/systems/ClickPlacementSystem.tsx
```diff
@@ -143,9 +143,9 @@ const ClickPlacementReactor = (props: { parentEntity: Entity }) => {
         (uuid) => uuid !== getComponent(clickState.placementEntity.value, UUIDComponent)
       )
       EditorControlFunctions.removeObject([clickState.placementEntity.value])
-      removeEntity(clickState.placementEntity.value)
       clickState.placementEntity.set(UndefinedEntity)
       SelectionState.updateSelection(selectedEntities)
+      applyIncomingActions()
```
HexaField, line 148: Running applyIncomingActions here will likely have unintended consequences, what is the reason for doing this?

packages/editor/src/services/EditorServices.ts
MichaelEstes: The EngineHooks.tsx file in client-core that handles the scene link logic isn't loaded through the studio flow, I can make it a part of the engine loading flow for the studio though

packages/editor/src/systems/ClickPlacementSystem.tsx
MichaelEstes: Ok, I'll try to dig more into the root cause, at the base it seems like there's a race condition with removeObject where there's a frame in the GLTFSnapshotState where the entity no longer exists, but is still listed in the GLTF nodes

PR #10715, opened 2024-07-24
IR-3340: Fix Assets Panel Drag and Drop

dinomut1:
Fixes issues with drag and drop from the assets panel into properties panel. The data shape of the drag data from the assets panel now matches the file browser

Closes https://tsu.atlassian.net/browse/IR-3340


aditya-mitra (changes requested):
This does not seem to solve the issue

(in the below video, drag and drop of a model glb from the file browser works to change the model url but from assets panel does not)
https://github.com/user-attachments/assets/798259a8-4944-4800-beaf-c36254a41ba8

aditya-mitra (approved):
works fine

PR #10723, opened 2024-07-25
Changes to fix client build failing in docker

hanzlamateen:
## Summary

Quick fix to resolve build-client failing in deployments

## References
closes #_insert number here_

### DIFF
--- a/dockerfiles/client/Dockerfile-client-serve-static
+++ b/dockerfiles/client/Dockerfile-client-serve-static
@@ -22,6 +22,8 @@ RUN npm install --loglevel notice --legacy-peer-deps --production
 # copy then compile the code
 COPY . .
 
+RUN npm install vite@5.1.7
+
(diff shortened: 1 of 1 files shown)


dockerfiles/client/Dockerfile-client-serve-static
```diff
@@ -22,6 +22,8 @@ RUN npm install --loglevel notice --legacy-peer-deps --production
 # copy then compile the code
 COPY . .
 
+RUN npm install vite@5.1.7
```
DanielBelmes, line 25: This is really strange as the npm install above should be install vite

barankyle:
The problem seemed to be due to the addition of the dependency `vitest`. Unclear how a devDependency was messing up a production npm install, but [removing it](https://github.com/EtherealEngine/etherealengine/pull/10729) has resolved vite not being present for build-client.

PR #10695, opened 2024-07-22
Remove old editor files and UI elements

SYBIOTE:
This pull request removes old editor files, updates the editor name and route, and removes various panels, dialogs, layouts, and UI elements that are no longer needed. These changes help to slim down the editor packages and remove unused components.

PR #10727, opened 2024-07-25
3239 - Model compression added to FileBrowserContentPanel

internetcharles:
## Summary
Model compression was not enabled in FileBrowserContentPanel.

## References
closes #_insert number here_


dinomut1:
This is the old studio ui which is getting deleted as part of #10695
In the new studio ui this is already implemented

PR #10730, opened 2024-07-25
remove RenderOrderComponent

HexaField:
## Summary

`RenderOrderComponent` is redundant, was added for but does not actually improve performance, and actually removes functionality we can currently take advantage of

## References
closes #_insert number here_


the-reality-engineer (approved):
As I discussed w/ Josh, we aren't implementing this functionality in the ECS correctly right now; we can revisit implementing this in the ECS at a later date.

PR #10731, opened 2024-07-26
Make lodash esm compat

DanielBelmes:
## References
closes #_insert number here_


packages/engine/package.json
```diff
@@ -50,9 +50,9 @@
     "detect-gpu": "^5.0.38",
     "draco3dgltf": "^1.5.6",
     "fflate": "0.7.4",
-    "hls.js": "^1.3.5",
```
DanielBelmes, line 53: I also updated hls because they said it would lower bundle size but it didn't get that much lower :(

hanzlamateen (changes requested):
It seems like we also need to make changes in MT project to use `lodash-es`

PR #10732, opened 2024-07-26
Add TOS Checks & Disable Chat and Media if user has not accepted

HexaField:
- add terms of service and age checkboxes
- disable chat and media if user has not accept terms

## References
closes #_insert number here_


barankyle (approved):
It looks to be working well. Couple of thoughts:

Is it at all unintuitive that you have to open the user menu to accept the TOS and verify age? I loaded it locally  after only glancing at the code and was slightly surprised to not immediately see anything. But having some sort of popup that appears as soon as you load the page would be rather annoying and create friction. I don't know if a button up where the media controls would normally be that opened a small modal just for accepting TOS/verified age would be clunky or just as unintuitive.

I thought I heard talk of not letting guest users do media chat at all. Maybe that was ultimately ruled against. I suppose it would be relatively trivial to add later.

HexaField:
> It looks to be working well. Couple of thoughts:
> 
> Is it at all unintuitive that you have to open the user menu to accept the TOS and verify age? I loaded it locally after only glancing at the code and was slightly surprised to not immediately see anything. But having some sort of popup that appears as soon as you load the page would be rather annoying and create friction. I don't know if a button up where the media controls would normally be that opened a small modal just for accepting TOS/verified age would be clunky or just as unintuitive.
> 
> I thought I heard talk of not letting guest users do media chat at all. Maybe that was ultimately ruled against. I suppose it would be relatively trivial to add later.

We have some more comprehensive designs in figma for the location UI overhaul which includes some better flows for this. I thought also it might make sense to put a notification badge on the profile menu hot bar.

HexaField:
Some changes pending confirmation:

- require user to actually log in with Oauth, rather than just check the TOS
- should not hide buttons, but disable them prompting user to sign up

packages/client-core/src/components/InstanceChat/index.tsx
```diff
@@ -460,11 +460,32 @@ export const InstanceChatWrapper = () => {
   const { bottomShelfStyle } = useShelfStyles()
 
   const acceptedTOS = useMutableState(AuthState).user.acceptedTOS.value
+  const isGuest = useMutableState(AuthState).user.isGuest.value
 
   const networkWorldConfig = useHookstate(getMutableState(NetworkState).config.world)
   const targetChannelId = useHookstate(getMutableState(ChannelState).targetChannelId)
 
-  if (!acceptedTOS) return null
+  if (!isGuest)
```
barankyle: This should be `if (isGuest)`, right? Since we should only be telling people they need to log in if they're guests.

packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
```diff
@@ -463,7 +464,7 @@ const ProfileMenu = ({ hideLogin, onClose, isPopover }: Props): JSX.Element => {
               </Text>
             )}
 
-            {!originallyAcceptedTOS.value && (
+            {isGuest && (
```
barankyle, line 467: The TOS acceptance and 13orOver checkbox will never be saved for guest users since it's only going off of isGuest. Is that what we want? Or should a guest user have this information saved so they don't have to re-check the boxes every time they go to a location?

packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
```diff
@@ -500,10 +501,36 @@ const ProfileMenu = ({ hideLogin, onClose, isPopover }: Props): JSX.Element => {
...
+                      className={styles.termsLink}
+                    >
+                      {t('user:usermenu.profile.confirmAge13')}
+                    </div>
+                  }
+                />
+              </Grid>
+            )}
+
+            {!isGuest && !originallyAcceptedTOS.value && (
```
barankyle, line 524: It seems like this is a fallback for users who existed before this PR, right? Because I normally don't see this if I start as a new guest user, accept TOS and verify age, and then log in. But I am able to make it appear if I set the acceptedTOS column to false for a non-guest user. Fine if it is, just thinking out loud mostly.

packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
```diff
@@ -109,14 +109,15 @@ const ProfileMenu = ({ hideLogin, onClose, isPopover }: Props): JSX.Element => {
...
+  const checked13OrOver = useHookstate(!isGuest)
   const checked18OrOver = useHookstate(acceptedTOS)
-  const hasAcceptedTermsAndAge = checkedTOS.value && checked18OrOver.value
+  const hasAcceptedTermsAndAge = checkedTOS.value && checked13OrOver.value
 
   const originallyAcceptedTOS = useHookstate(acceptedTOS)
 
   useEffect(() => {
-    if (!originallyAcceptedTOS.value && hasAcceptedTermsAndAge) {
+    if (!originallyAcceptedTOS.value && checked18OrOver) {
```
barankyle: This needs to be checking checked18OrOver.value, as otherwise the reference to the hookstate proxy always reads as truthy, and acceptedTOS is getting patched to true as soon as the user profile modal opens. Should probably make the useEffect listener also listen on checked18OrOver.value too.

But, I also don't see where checking the TOS box itself is patching acceptedTOS to true anywhere.

barankyle (changes requested):
There seems to be a hole in patching acceptedTOS to true for guests, plus a few other minor issues.

packages/client-core/src/components/InstanceChat/index.tsx
HexaField: Whoops! You're right

packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
HexaField: The idea here is that log in is not allowed until they accept the TOS & verify 13+, the only information that is saved is 18+, which is saved as the accountTOS field in the user table, which i suppose we should rename, but i'm ok with it as is for now until we have the new UI & flow implemented.

HexaField: Guests do not have any restrictions, the restriction is on media/chat, which is restricted to non-guest users who have checked 18+

packages/client-core/src/user/components/UserMenu/menus/ProfileMenu.tsx
HexaField: This is what I'm currently unsure on. If a user checks 18+ they get access to media/chat, but why would it ever be valid for them to uncheck it? In most flows, you either enter a birthday or do this upon sign up, in which the option is not available once checked. This is kind of how it is implemented as a quick implementation until the new UI & flow is in, just to ensure we have a minimum level of compliance.

HexaField: Oh, and the checkbox is disabled once the user checks it, so they can't uncheck it.

barankyle: Maybe I was thrown off by what the if statement bug was doing. Are users now only given media chat access when they log in and verify they're 18? If so, then this seems fine with the fix.

barankyle (approved):

PR #10734, opened 2024-07-26
fix assets panel refresh and back functionality

aditya-mitra:
## Summary

earlier the refresh button was unselecting the current category (due to which we were getting to the parent category). fixes it by not selecting the null category when refreshing

earlier the back button used to load infinitely due a hookstate error. fixes it by passing the no proxy object

## References

https://tsu.atlassian.net/browse/IR-3359


MichaelEstes (approved):
Looks good

PR #10735, opened 2024-07-26
fix reordering in hierarchy panel

aditya-mitra:
## Summary

- added nodes reparenting in addition to scenes.nodes for fixing reordering of nodes in EditorControls
- added types in HierarchyNode
- fixed the drop issue before/after node in hierarchy panel by giving the div a hieght

## References

https://tsu.atlassian.net/browse/IR-3300


MichaelEstes (commented):
The UI for this is good, but it looks like we don't currently have a way to reorder the nodes of the GLTF document, only reparent them. Edit: Oh I misunderstood the code, ignore this

packages/ui/src/components/editor/panels/Hierarchy/node/index.tsx
```diff
@@ -121,37 +118,23 @@ export const HierarchyTreeNode = (props: HierarchyTreeNodeProps) => {
...
-      else if (e.key === 'Enter') data.onRenameSubmit(entity, (e.target as any).value)
-    },
-    [data.onRenameSubmit, node]
-  )
-
-  const onClickNode = useCallback((e) => data.onClick(e, entity), [node, data.onClick])
-  const onMouseDownNode = useCallback((e) => data.onMouseDown && data.onMouseDown(e, entity), [node, data.onMouseDown])
-
-  const onChangeNodeName = useCallback((e) => data.onChangeName(entity, e.target.value), [node, data.onChangeName])
+  const onClickToggle = (e: React.MouseEvent) => {
```
SYBIOTE, line 121: use callback here for optimization

packages/ui/src/components/editor/panels/Hierarchy/node/index.tsx
```diff
@@ -208,9 +189,9 @@ export const HierarchyTreeNode = (props: HierarchyTreeNodeProps) => {
         console.warn('parent is not defined')
       }
 
-    return (item: any, monitor): void => {
+    return (item: FileDataType | DnDFileType | DragItemType, monitor: DropTargetMonitor): void => {
       if (parentNode) {
-        if (item.files) {
+        if ('files' in item) {
```
SYBIOTE, line 194: can we use something other than a static string here?

packages/ui/src/components/editor/panels/Hierarchy/node/index.tsx
```diff
@@ -224,41 +205,51 @@ export const HierarchyTreeNode = (props: HierarchyTreeNodeProps) => {
           return
         }
 
-        if (item.url) {
+        if ('url' in item) {
```
SYBIOTE, line 208: here as well, please convert to an Enum

SYBIOTE (changes requested):

packages/editor/src/functions/EditorControlFunctions.ts
```diff
@@ -569,8 +569,11 @@ const reparentObject = (entities: Entity[], before?: Entity | null, parent = get
             (n) => n.extensions?.[UUIDComponent.jsonID] === getComponent(before, UUIDComponent)
           )
           gltf.data.scenes![0].nodes.splice(beforeIndex, 0, nodeIndex)
+          gltf.data.nodes?.splice(beforeIndex, 0, gltf.data.nodes?.[nodeIndex])
```
HexaField, line 572: what are these extra lines doing exactly?

HexaField (changes requested):

packages/editor/src/functions/EditorControlFunctions.ts
aditya-mitra: When the node is reparented in scene root, the order previously changed only in `gltf.data.scenes`. However, the hierarchy panel is dependent on change in `gltf.data.nodes` for [hierarchy change detection](https://github.com/EtherealEngine/etherealengine/blob/0e1e55d8809128472211a42e5d4efc4a66df7324/packages/ui/src/components/editor/panels/Hierarchy/container/index.tsx/#L204-L207) (tried to do it in the hierarchy panel itself for detecting this change but it becomes too complex and requiring almost a full refactor of our change detection and tree walking algorithm). 
Since we are [changing the order in nodes' children when it is not the parent](https://github.com/EtherealEngine/etherealengine/blob/0e1e55d8809128472211a42e5d4efc4a66df7324/packages/editor/src/functions/EditorControlFunctions.ts#L585>), we should also change it in the gtlf nodes' array

packages/ui/src/components/editor/panels/Hierarchy/node/index.tsx
aditya-mitra: Here we are passing these functions directly to the event handlers. Had these functions been dependencies of some `useMemo` or `useEffect`, we could have wrapped them in `useCallback`s. In this case, it would rather been an overhead (due to redundant cache being stored) instead of optimization.

packages/ui/src/components/editor/panels/Hierarchy/node/index.tsx
aditya-mitra: This is also the key in the object. This helps determining the type of the object if those keys are present.
The other way around is to do `(item as any).files`which we probably want to avoid.

packages/ui/src/components/editor/panels/Hierarchy/node/index.tsx
aditya-mitra: same reason here.

packages/editor/src/functions/EditorControlFunctions.ts
HexaField: Gotcha. We definitely want to do what makes the most sense for the GLTF data, it is the fundamental data that should describe how the hierarchy panel displays objects. Ie, the hierarchy panel should show exactly what is in the GLTF, so this makes sense.

packages/ui/src/components/editor/panels/Hierarchy/node/index.tsx
SYBIOTE: Gotcha

packages/ui/src/components/editor/panels/Hierarchy/node/index.tsx
SYBIOTE: Maybe checking for the existence of the property in the object is better syntax wise

Using only a static string just doesn't feel right

aditya-mitra: Actually, [enums](https://www.typescriptlang.org/docs/handbook/enums.html) are used for an alternative to exhaustive constants in JS. Enums are best useful when you have [a finite set of choices](https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html).

[JS recommends to use string](https://arc.net/l/quote/smujllqd) to check within objects.

HexaField: Just confirming that strings are fine here.

HexaField (approved):

SYBIOTE (approved):

PR #10736, opened 2024-07-26
IR-3053: Add project-history service

CITIZENDOT:
## Summary
Add `project-history` service to track changes across project. Detailed technical info about this service is here: https://tsu.atlassian.net/wiki/x/F4AhjQ

## Subtasks Checklist
- [x] Add hooks to related services to log changes
- [x] User Interface for all the history types.

## References
closes https://tsu.atlassian.net/browse/IR-3053


packages/server-core/src/projects/project-history/project-history.hooks.ts
```diff
@@ -0,0 +1,180 @@
...
+import { BadRequest } from '@feathersjs/errors'
+import {
+  projectHistoryDataResolver,
+  projectHistoryExternalResolver,
+  projectHistoryQueryResolver,
+  projectHistoryResolver
+} from './project-history.resolvers'
+
+import { staticResourcePath, UserID, userPath } from '@etherealengine/common/src/schema.type.module'
+import { isSuperAdmin } from '@etherealengine/projects/projects/eepro-multitenancy/services/hooks/common/is-super-admin'
```
barankyle: Can't import from eepro-multitenancy.

barankyle (changes requested):
ProjectHistory needs to be imported in server-core/src/projects/service.ts and added to the export array from that file. Right now the service isn't being registered because it's not in any service.ts file.

If I comment out isSuperAdmin, when I run the API server locally, the startup runs of uploadLocalProjectToProvider fail because there's no actionIdentifier in the requests to project-history. Additionally, if I go to the studio and attempt to save a scene, there's no actionIdentifier there, either.

packages/server-core/src/projects/project-history/project-history.hooks.ts
CITIZENDOT: > - ProjectHistory needs to be imported in server-core/src/projects/service.ts and added to the export array from that file
> - Additionally, if I go to the studio and attempt to save a scene, there's no actionIdentifier there, either

Done

> Can't import from eepro-multitenancy

Can you elaborate this? Multitenancy doesn't need to import the service. It just needs to import `projectHistoryPath`.

HexaField: You're importing from a project within the core engine. The core engine can only import from the core engine packages.

CITIZENDOT: Oh! Got it.

CITIZENDOT: Fixed it.

packages/client-core/src/admin/components/project/ProjectHistory.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Ethereal Engine team.
+
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023 
+Ethereal Engine. All Rights Reserved.
+*/
+
+import { useHookstate } from '@etherealengine/hyperflux'
+import {
```
HexaField: client cannot import server files

packages/server-core/src/projects/project-history/hooks.ts
```diff
@@ -0,0 +1,245 @@
...
+    },
+    paginate: false
+  })
+
+  for (const projectPermission of projectPermissions) {
+    context['permissionTypeBeforeUpdate'][projectPermission.id] = projectPermission.type
+  }
+}
+
+staticResourceHooks.before.update.unshift(storeResourceKey)
```
HexaField: we need to move these into the respective hook definitions, rather than injecting them here

HexaField (changes requested):

packages/client-core/src/admin/components/project/ProjectHistory.tsx
CITIZENDOT: Moved schema definition & path to common package

packages/server-core/src/projects/project-history/hooks.ts
CITIZENDOT: Since we're discussing whether to follow hooks or trigger approach, I'll pause on this, until we confirm what we should finalise.

CITIZENDOT: Removed all hooks

barankyle (approved):

hanzlamateen (changes requested):
Fix user name field for removed use cases.

CITIZENDOT:
> Fix user name field for removed use cases.

Done

hanzlamateen (approved):

PR #10737, opened 2024-07-26
[IR-3324] studio: drag-n-drop folders

dtlehrer:
## Summary
[IR-3324] enables users to include folders in drag-n-drop selection, moves folders along with files, and fixes inconsistencies related to moving files into sub-folders

## References
closes #_insert number here_

## QA Steps

[IR-3324]: https://tsu.atlassian.net/browse/IR-3324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dinomut1 (dismissed):
Seeing strange behavior with dragging folders into other folders:
[Screencast from 07-30-2024 04:23:27 PM.webm](https://github.com/user-attachments/assets/e29b24b4-d9d0-42d3-879c-8649430704b8)

dtlehrer:
> Seeing strange behavior with dragging folders into other folders: [Screencast from 07-30-2024 04:23:27 PM.webm](https://github.com/user-attachments/assets/e29b24b4-d9d0-42d3-879c-8649430704b8)

this appears to stem from a deeper issue in `FileUtil.getIncrementalName(...)`

have noticed that it is possible to create a folder named 'texture' and then 'textures', but if you have a folder named 'textures' and then try to create one named 'texture', it will rename to 'texture(1)' even when no other folder named 'texture' exists .. the same holds true when testing other singular/plural directory names

HexaField:
Let's make sure we add comprehensive regression tests for this bug.

dtlehrer:
bugs in s3.storage `isDirectory` and `moveObject` methods resolved.. moving folders recursively now appears to be working as intended: 

https://github.com/user-attachments/assets/1684acc1-60ab-485a-b4c8-2a0437c99d67

next commit will include regression tests

packages/ui/src/components/editor/panels/Files/container/index.tsx
```diff
@@ -297,7 +297,12 @@ const FileBrowserContentPanel: React.FC<FileBrowserContentPanelProps> = (props)
   }
 
   const onSelect = (event, params: FileDataType) => {
-    if (params.type !== 'folder') {
+    if (params.isFolder) {
```
DanielBelmes: I believe this is breaking the new functionality to double click into folders. Which we want to keep.

packages/ui/src/components/editor/panels/Files/container/index.tsx
```diff
@@ -306,11 +311,6 @@ const FileBrowserContentPanel: React.FC<FileBrowserContentPanelProps> = (props)
       })
 
       ClickPlacementState.setSelectedAsset(params.url)
-    } else {
-      if (event.detail === 2) {
```
DanielBelmes, line 310: Since this isn't very clear what functionality we are using. Do you mind commenting what event.detail 2 is?

dtlehrer: event.detail === 2 was handling double-clicks

packages/ui/src/components/editor/panels/Files/container/index.tsx
dtlehrer: this PR should make it so users can navigate into folders with a single-click

also updated to make it so double-clicks always navigate into folders

DanielBelmes (approved):

PR #10744, opened 2024-07-28
feat(IR-2630): adding missing classes from designs, adding secondary …

kattsushi:
- feat(IR-2630): adding missing classes from desings, adding secondary button variant

## References
related Jira ticket https://tsu.atlassian.net/browse/IR-2630

### DIFF
--- a/packages/client-core/src/common/services/ThemeService.tsx
+++ b/packages/client-core/src/common/services/ThemeService.tsx
@@ -6,8 +6,8 @@ Version 1.0. (the "License"); you may not use this file except in compliance
 with the License. You may obtain a copy of the License at
 https://github.com/EtherealEngine/etherealengine/blob/dev/LICENSE.
 The License is based on the Mozilla Public License Version 1.1, but Sections 14
-and 15 have been added to cover use of software over a computer network and 
-provide for limited attribution for the Original Developer. In addition, 
+and 15 have been added to cover use of software over a computer network and
+provide for limited attribution for the Original Developer. In addition,
 Exhibit A has been modified to be consistent with Exhibit B.
 
 Software distributed under the License is distributed on an "AS IS" basis,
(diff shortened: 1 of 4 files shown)


SYBIOTE (approved):
Great, looks good

packages/client-core/src/common/services/ThemeService.tsx
```diff
@@ -56,9 +63,14 @@ const lightTheme: CSSClasses = {
   'bg-primary': '#F5F5F5',
   'bg-secondary': '#FFFFFF',
   'bg-highlight': '#D9D9D9',
+  'bg-accent': '#375DAF',
```
HexaField: There are so many duplicates here. Do we really need so many with the same colour?

HexaField (dismissed):

packages/ui/src/primitives/tailwind/Button/index.tsx
```diff
@@ -32,7 +32,16 @@ export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
   endIcon?: ReactNode
   children?: ReactNode
   size?: 'small' | 'medium' | 'large'
-  variant?: 'primary' | 'outline' | 'danger' | 'success' | 'pink' | 'transparent' | 'onboarding' | 'sidebar'
+  variant?:
```
HexaField: This is getting to a lot of variants, are they all actually used?

packages/client-core/src/common/services/ThemeService.tsx
kattsushi: good catch that i didn't realized, fixed removed the duplicates

packages/ui/src/primitives/tailwind/Button/index.tsx
kattsushi: Checking all variants are being used except pink, so I just removed it

packages/ui/src/primitives/tailwind/Button/index.tsx
```diff
@@ -32,7 +32,7 @@ export interface ButtonProps extends React.HTMLAttributes<HTMLButtonElement> {
   endIcon?: ReactNode
   children?: ReactNode
   size?: 'small' | 'medium' | 'large'
-  variant?: 'primary' | 'outline' | 'danger' | 'success' | 'pink' | 'transparent' | 'onboarding' | 'sidebar'
+  variant?: 'primary' | 'secondary' | 'outline' | 'danger' | 'success' | 'transparent' | 'onboarding' | 'sidebar'
```
aditya-mitra: we do not have onboarding variant anymore in the button

aditya-mitra (approved):
Looks good. Just need to resolve 1 comment

packages/client-core/src/common/services/ThemeService.tsx
```diff
@@ -62,9 +67,13 @@ const lightTheme: CSSClasses = {
   'bg-primary': '#F5F5F5',
   'bg-secondary': '#FFFFFF',
   'bg-highlight': '#D9D9D9',
+  'bg-surface-bg': '#FFFFFF',
   'bg-surface-main': '#FFFFFF',
```
aditya-mitra, line 71: can you also update surface main color to `#0E0F11`?

PR #10750, opened 2024-07-29
IR-3352 Scene-configured virtual spectator cameras

AidanCaruso:
## Summary
 - Spectate system now allows you to spectate any entity
 - Scenes can now be set to spectator-only by scene settings. Without assigning a camera the user is given fly camera controls, with an assigned entity the user's camera spectates the assigned entity.
 - The spectator system no longer looks for networked cameras owned by users, it spectates the entity at the uuid passed in through either the URL or the scene settings 
 - Avatar spawn system reactor now has an isClient check to prevent avatars being spawned by the server even when users are spectating

## References
https://tsu.atlassian.net/browse/ir-3352


packages/ui/src/components/editor/properties/scene/settings/index.tsx
```diff
@@ -164,12 +171,39 @@ export const SceneSettingsEditor: EditorComponentType = (props) => {
...
+      >
+        <BooleanInput value={useSpectatingEntity.value} onChange={useSpectatingEntity.set} />
+      </InputGroup>
+      {useSpectatingEntity.value ? (
+        <InputGroup
+          name="Entity UUID"
+          label={t('editor:properties.sceneSettings.lbl-uuid')}
+          info={t('editor:properties.sceneSettings.info-uuid')}
+        >
+          <SelectInput
```
dinomut1: Can we use a NodeInput here instead to let people drag and drop entities instead of selecting from drop down?

dinomut1 (changes requested):

packages/spatial/src/camera/components/VirtualCameraComponent.ts
```diff
@@ -0,0 +1,51 @@
...
+The Original Developer is the Initial Developer. The Initial Developer of the
+Original Code is the Ethereal Engine team.
+
+All portions of the code written by the Ethereal Engine team are Copyright © 2021-2023 
+Ethereal Engine. All Rights Reserved.
+*/
+
+import { defineComponent } from '@etherealengine/ecs'
+
+export const VirtualCameraComponent = defineComponent({
```
HexaField: we can re-use the camera component, we dont need a separate one. a PR was merged in recently that exposes the same values

HexaField (dismissed):

packages/spatial/src/camera/components/VirtualCameraComponent.ts
AidanCaruso: @HexaField Okay cool, I'll make the switch. I was going off of the ADR which requested a virtual camera component. The reason I didn't just stick with the camera component initially is because each uses an instance of the threejs camera object which seemed unnecessary for virtual cameras that only need to describe a few transform/frustum values for the viewer entity camera to lerp to.

dinomut1 (approved):
This looks good

dinomut1:
To clarify -- we do want a virtual camera concept that is separate from the camera. We can do that in a subsequent PR

PR #10752, opened 2024-07-29
Restrict file name length to 64 characters

SYBIOTE:
This pull request restricts the length of file names to 64 characters. Previously, there was no limit on the length of file names, which could lead to issues. This change ensures that file names are within the acceptable range.


packages/common/src/utils/validateFileName.ts
```diff
@@ -29,6 +29,7 @@ export function isValidFileName(fileName: string) {
   return (
     !INVALID_FILENAME_WHITESPACE_REGEX.test(fileName) &&
```
aditya-mitra, line 30: we can restrict the fileName length in the regex in itself. somewhat like `{1,64}`

aditya-mitra (dismissed):

packages/common/src/utils/validateFileName.ts
aditya-mitra: we can restrict the fileName length in the regex in itself. somewhat like `{1,64}`

aditya-mitra (changes requested):

packages/common/src/utils/validateFileName.ts
SYBIOTE: Is that necessary, I would not want to bundle the length restriction into the regex, 
We use regex across the codebase for mainly character search, baking it in for a one specific case doesn't feel right.

Take a look at https://tsu.atlassian.net/browse/IR-3169

We can take up advanced regex and make it a part of this PR, the effort will be more centralized

aditya-mitra: We can actually use length restrictions in regex. We have done in the scene name regex - https://github.com/EtherealEngine/etherealengine/blob/0e1e55d8809128472211a42e5d4efc4a66df7324/packages/common/src/regex/index.ts/#L33

SYBIOTE: Gotcha

SYBIOTE: implemented

packages/common/tests/regex.test.ts
```diff
@@ -61,14 +60,23 @@ describe('regex.test', () => {
...
+        'another\ncontrol',
+        'file name', // (a space and an underscore)
+        '< tag >', // (spaces and less-than < and greater-than > characters)
+        'key : value', // (a space and a colon :)
+        'quote " example', // (a space and a double quote ")
+        'path / to / file', // (spaces and forward slashes /)
+        'C:\\ path \\ to \\ file', // (spaces and backslashes \)
+        'pipe | character', // (a space and a pipe |)
+        'question ? mark', // (a space and a question mark ?)
+        'star * char' // (a space and an asterisk *)
```
aditya-mitra: your strings are good and self-explanatory. we should remove comments where they can be understood

aditya-mitra (dismissed):

packages/common/tests/regex.test.ts
SYBIOTE: does it matter? ill remove them nonetheless

SYBIOTE: done

CITIZENDOT (approved):

PR #10753, opened 2024-07-29
IR-2343: add basic cache to asset loader, use to cache prefab shelf

dinomut1:
Caches prefab shelf assets so that they're only fetched once, when the prefab shelf is initialized.

Closes https://tsu.atlassian.net/browse/IR-2343

### DIFF
--- a/packages/editor/src/components/prefabs/PrefabEditors.tsx
+++ b/packages/editor/src/components/prefabs/PrefabEditors.tsx
@@ -18,9 +18,9 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 import config from '@etherealengine/common/src/config'
-import { AssetLoader } from '@etherealengine/engine/src/assets/classes/AssetLoader'
+import { useGLTF } from '@etherealengine/engine/src/assets/functions/resourceLoaderHooks'
 import { defineState, getMutableState, useHookstate } from '@etherealengine/hyperflux'
-import React, { ReactNode, useEffect } from 'react'
+import React, { ReactNode } from 'react'
 import { FiHexagon } from 'react-icons/fi'
 
 export type PrefabShelfItem = {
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/prefabs/PrefabEditors.tsx
```diff
@@ -143,5 +144,20 @@ export const PrefabShelfState = defineState({
...
-    ] as PrefabShelfItem[]
+    ] as PrefabShelfItem[],
+  reactor: () => {
+    const shelfState = useHookstate(getMutableState(PrefabShelfState))
+    return shelfState.value.map((shelfItem) => <ShelfItemReactor key={shelfItem.url} url={shelfItem.url} />)
+  }
 })
+
+const ShelfItemReactor = (props: { key: string; url: string }): JSX.Element | null => {
+  useEffect(() => {
```
MichaelEstes: We have fairly robust caching already if you change this useEffect to useGLTF(props.url)

dinomut1: That's how the model component works normally and it's causing fetches every time on these prefabs

dinomut1: oh weird it works. Why isn't the useGLTFDocument in the model component already caching then?

MichaelEstes: I'd need some more context around what's happening, but my best guess is that the model components are being unmounted/remounted and if they're the only reference to the asset it will be removed from the cache and then re-added. I'm sure there's some tweaks we can do to the cache lifetimes if it's causing issues.

MichaelEstes (approved):

HexaField (approved):
nice!

packages/editor/src/components/prefabs/PrefabEditors.tsx
dinomut1: oh yes the dereference is unmounting them

PR #10757, opened 2024-07-30
Fix scaling rigidbodies and collider updates

HexaField:
## Summary

Turned out to be 3 compounding issues here, solved by:

- dirty transform child update iteration now happens after physics pre transform updates
- rigidbody to transform update now considers scale of ancestors
- transform to collider updates did not consider scale, now recreates collider

## References
closes #_insert number here_


dinomut1 (approved):
excellent

PR #10758, opened 2024-07-30
[IR-3420] Unit Tests: Coverage-check tooling

heysokam:
## Summary
Implements the tools/scripts necessary to manage test-coverage reports locally.

## References
https://tsu.atlassian.net/browse/IR-3420


package.json
```diff
@@ -92,6 +96,39 @@
     "add-license-headers": "cross-env ts-node --swc scripts/add-license-headers.ts",
     "storybook": "cd packages/ui && npm run storybook"
   },
+  "nyc": {
```
HexaField: is there a config file we can put this in instead of having it in the package.json ?

HexaField (changes requested):

package.json
heysokam: There is, yeah. I will move it to their specific .json config :writing_hand:

heysokam: @HexaField Implemented and pushed, but it doesn't pickup the options for the packages if the configuration is in its own file.

Should I revert the change and keep it into the package.json, or should I create a duplicate config file for each package instead?

heysokam: Ah, nvm. I just found the explicit `--nycrc-path` option, which seems to solve the problem when passed to the root command
Solved in c48ae10

HexaField (approved):

PR #10764, opened 2024-07-30
Removed components and put visual script behind feature flag

DanielBelmes:
## References
closes #_insert number here_


packages/editor/src/services/ComponentShelfCategoriesState.ts
```diff
@@ -71,13 +68,15 @@ import { InputComponent } from '@etherealengine/spatial/src/input/components/Inp
...
 import { RigidBodyComponent } from '@etherealengine/spatial/src/physics/components/RigidBodyComponent'
 import { TriggerComponent } from '@etherealengine/spatial/src/physics/components/TriggerComponent'
-import { GroupComponent } from '@etherealengine/spatial/src/renderer/components/GroupComponent'
 import { PostProcessingComponent } from '@etherealengine/spatial/src/renderer/components/PostProcessingComponent'
 import { LookAtComponent } from '@etherealengine/spatial/src/transform/components/LookAtComponent'
 
 export const ComponentShelfCategoriesState = defineState({
   name: 'ee.editor.ComponentShelfCategories',
   initial: () => {
+    const [visualScriptPanelEnabled] = useFeatureFlags([FeatureFlags.Studio.Panel.VisualScript])
```
HexaField: this is not a reactive context, you'll need to add a reactor to this state and handle mutations there

HexaField (changes requested):

packages/editor/src/services/ComponentShelfCategoriesState.ts
```diff
@@ -112,27 +103,23 @@ export const ComponentShelfCategoriesState = defineState({
...
-        TextComponent,
-        ScreenshareTargetComponent,
-        LookAtComponent
-      ]
+      Visual: [EnvMapBakeComponent, ScenePreviewCameraComponent, SkyboxComponent, TextComponent, LookAtComponent]
     } as Record<string, Component[]>
+  },
+  reactor: () => {
+    const [visualScriptPanelEnabled] = useFeatureFlags([FeatureFlags.Studio.Panel.VisualScript])
+    const cShelfState = getState(ComponentShelfCategoriesState)
```
dinomut1: Could we do this with a getMutableState just in case we want to react to this change elsewhere?

dinomut1 (changes requested):

packages/editor/src/services/ComponentShelfCategoriesState.ts
```diff
@@ -112,27 +103,22 @@ export const ComponentShelfCategoriesState = defineState({
...
-      ]
+      Visual: [EnvMapBakeComponent, ScenePreviewCameraComponent, SkyboxComponent, TextComponent, LookAtComponent]
     } as Record<string, Component[]>
+  },
+  reactor: () => {
+    const [visualScriptPanelEnabled] = useFeatureFlags([FeatureFlags.Studio.Panel.VisualScript])
+    const cShelfState = getMutableState(ComponentShelfCategoriesState)
+    useEffect(() => {
+      if (visualScriptPanelEnabled) {
+        cShelfState.Scripting.set([VisualScriptComponent])
```
HexaField: let's use `merge` instead of `set` to allow other reactors to mutate the array and not be overridden too

HexaField (dismissed):

dinomut1 (approved):

packages/editor/src/services/ComponentShelfCategoriesState.ts
```diff
@@ -105,27 +103,24 @@ export const ComponentShelfCategoriesState = defineState({
...
-        ShadowComponent,
-        ParticleSystemComponent,
-        EnvmapComponent,
-        SDFComponent,
-        PostProcessingComponent,
-        ReflectionProbeComponent
-      ],
-      Scripting: [VisualScriptComponent],
+      FX: [LoopAnimationComponent, ShadowComponent, ParticleSystemComponent, EnvmapComponent, PostProcessingComponent],
+      Scripting: [],
```
SYBIOTE, line 107: if its empty, why keep the scripting field at all? 
is the system component not relevant anymore?

SYBIOTE (approved):

packages/editor/src/services/ComponentShelfCategoriesState.ts
DanielBelmes: Keep it defined. It's not really needed but simplifies the reactor logic a tiny bit

PR #10770, opened 2024-07-30
Fixing the text color tailwind class names attributed to X/Y/Z property field names

Rezmason:
## References
closes https://tsu.atlassian.net/browse/IR-3409

## QA Steps

Make sure the X, Y and Z property labels are red, green, and blue like so in the Transform components in the property panel of selected entities:

image


HexaField (dismissed):
we can also remove the three variables from the [tailwind.conig.js](https://github.com/EtherealEngine/etherealengine/tree/dev/tailwind.config.js), they should never have been added in the first place.

SYBIOTE (approved):

aditya-mitra (approved):
LGTM

PR #10777, opened 2024-07-31
feat(IR-3436): make label prop of Checkbox component as string | Reac…

kattsushi:
…tNode

## Summary

Also updated the texts to match https://www.figma.com/design/a3qQCRxNWxP7K1titmso0e/Console-Design?node-id=5880-19887&t=8En1eQEUiLF2s3bi-0

## References
closes #_insert number here_


packages/ui/src/primitives/tailwind/Checkbox/index.tsx
```diff
@@ -32,7 +32,7 @@ import Label from '../Label'
 
 export interface CheckboxProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'value' | 'onChange'> {
   value: boolean
-  label?: string
+  label?: string | React.ReactNode
```
CITIZENDOT: If label is a ReactNode, it should be rendered as it is at Line 63 (not inside the Label component).

CITIZENDOT (changes requested):

packages/ui/src/primitives/tailwind/Checkbox/index.tsx
aditya-mitra: also type of label will be `label?: React.ReactNode` only since string is a sub type

aditya-mitra (changes requested):
as per the [component parity document](https://tsu.atlassian.net/wiki/x/AQC5j), label will always be a string [like shown here](https://arc.net/l/quote/lqyxrken)

after checking on IR-3436, it seems like the [required design according to figma](https://www.figma.com/design/a3qQCRxNWxP7K1titmso0e/Console-Design?node-id=5880-19887&t=n1SR7ip06vY9bboJ-0) is possible on the current checkbox

hanzlamateen:
> as per the [component parity document](https://tsu.atlassian.net/wiki/x/AQC5j), label will always be a string [like shown here](https://arc.net/l/quote/lqyxrken)
> 
> after checking on IR-3436, it seems like the [required design according to figma](https://www.figma.com/design/a3qQCRxNWxP7K1titmso0e/Console-Design?node-id=5880-19887&t=n1SR7ip06vY9bboJ-0) is possible on the current checkbox

@aditya-mitra we do need to add a link to term of service as done for the profile menu, Otherwise how can a user find/read terms of service. So we will need to make it ReactNode

image

packages/ui/src/primitives/tailwind/Checkbox/index.tsx
hanzlamateen: Yes, we should use `label?: React.ReactNode` instead of `label?: string | React.ReactNode`

hanzlamateen (approved):

PR #10779, opened 2024-07-31
Changes to use tailwind loading view

hanzlamateen:
## Summary
Change to make loading spinners similar across the application.

## References
closes #_insert number here_


SYBIOTE (approved):
great, safe to say we can remove loading circle now then?

hanzlamateen:
> great, safe to say we can remove loading circle now then?

It's still used in a couple of places in location UI which uses MUI.

PR #10781, opened 2024-07-31
create prefab remove original entity fix

JT00y:
## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
+++ b/packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
@@ -38,6 +38,7 @@ import {
 } from '@etherealengine/ecs'
 import { GLTFDocumentState } from '@etherealengine/engine/src/gltf/GLTFDocumentState'
 import { ModelComponent } from '@etherealengine/engine/src/scene/components/ModelComponent'
+import { SourceComponent } from '@etherealengine/engine/src/scene/components/SourceComponent'
 import { proxifyParentChildRelationships } from '@etherealengine/engine/src/scene/functions/loadGLTFModel'
 import { getMutableState, getState, startReactor, useHookstate } from '@etherealengine/hyperflux'
 import { TransformComponent } from '@etherealengine/spatial'
@@ -103,8 +104,10 @@ export default function CreatePrefabPanel({ entity }: { entity: Entity }) {
 
       removeEntity(prefabEntity)
       EditorControlFunctions.removeObject([entity])
+      const sceneID = getComponent(parentEntity, SourceComponent)
       const reactor = startReactor(() => {
         const documentState = useHookstate(getMutableState(GLTFDocumentState))
+        const nodes = documentState[sceneID].nodes
         useEffect(() => {
           if (!entityExists(entity)) {
             const { entityUUID } = EditorControlFunctions.createObjectFromSceneElement(
@@ -116,8 +119,10 @@ export default function CreatePrefabPanel({ entity }: { entity: Entity }) {
             )
             getMutableState(SelectionState).selectedEntities.set([entityUUID])
             reactor.stop()
+          } else {
+            console.log('Entity not removed')
           }
-        }, [documentState])
+        }, [nodes])
         return null
       })
       PopoverState.hidePopupover()
--- a/packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
+++ b/packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
@@ -102,7 +102,6 @@ export default function CreatePrefabPanel({ entity }: { entity: Entity }) {
       await Engine.instance.api.service(staticResourcePath).patch(resource.id, { tags: tags })
 
       removeEntity(prefabEntity)
-      await EditorControlFunctions.removeObject([entity])
       EditorControlFunctions.removeObject([entity])
       const reactor = startReactor(() => {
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
```diff
@@ -92,19 +90,21 @@ export default function CreatePrefabPanel({ entity }: { entity: Entity }) {
...
+
       removeEntity(prefabEntity)
-      const { entityUUID } = EditorControlFunctions.createObjectFromSceneElement(
+      await EditorControlFunctions.createPrefabObjectFromSceneElement(
         [
           { name: ModelComponent.jsonID, props: { src: fileURL } },
           { name: TransformComponent.jsonID, props: { position, rotation, scale } }
         ],
-        parentEntity
+        entity,
```
SYBIOTE: should this be parentEntity?

SYBIOTE (changes requested):

packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
JT00y: I made a new function for create prefab object from scene, here the entity means the original entity which will be unload in the scene and remove the node in snapshot.

SYBIOTE: so its intended, gotcha

SYBIOTE (approved):

dinomut1 (changes requested):
I don't think we want to be creating new editor control functions for something this specific. Here's a snippet of code you can add to `CreatePrefabPanelDialog.tsx` that will do the same thing, but using the existing control functions:
```
EditorControlFunctions.removeObject([entity])
const reactor = startReactor(() => {
  const documentState = useHookstate(getMutableState(GLTFDocumentState))
  useEffect(() => {
    if (!entityExists(entity)) {
      const { entityUUID } = EditorControlFunctions.createObjectFromSceneElement(
        [
          { name: ModelComponent.jsonID, props: { src: fileURL } },
          { name: TransformComponent.jsonID, props: { position, rotation, scale } }
        ],
        parentEntity
      )
      getMutableState(SelectionState).selectedEntities.set([entityUUID])
      reactor.stop()
    }
  }, [documentState])
  return null
})

```
So we use EditorControlFunctions.removeObject to delete the original entity, then set up a reactor which waits for that entity to be deleted, then we use the normal createObjectFromSceneElement to add the prefab reference.

JT00y:
> I don't think we want to be creating new editor control functions for something this specific. Here's a snippet of code you can add to `CreatePrefabPanelDialog.tsx` that will do the same thing, but using the existing control functions:
> 
> ```
> EditorControlFunctions.removeObject([entity])
> const reactor = startReactor(() => {
>   const documentState = useHookstate(getMutableState(GLTFDocumentState))
>   useEffect(() => {
>     if (!entityExists(entity)) {
>       const { entityUUID } = EditorControlFunctions.createObjectFromSceneElement(
>         [
>           { name: ModelComponent.jsonID, props: { src: fileURL } },
>           { name: TransformComponent.jsonID, props: { position, rotation, scale } }
>         ],
>         parentEntity
>       )
>       getMutableState(SelectionState).selectedEntities.set([entityUUID])
>       reactor.stop()
>     }
>   }, [documentState])
>   return null
> })
> ```
> 
> So we use EditorControlFunctions.removeObject to delete the original entity, then set up a reactor which waits for that entity to be deleted, then we use the normal createObjectFromSceneElement to add the prefab reference.

packages/editor/src/components/dialogs/CreatePrefabPanelDialog.tsx
```diff
@@ -92,19 +100,31 @@ export default function CreatePrefabPanel({ entity }: { entity: Entity }) {
       const resource = resources.data[0]
       const tags = [...prefabTag.value]
       await Engine.instance.api.service(staticResourcePath).patch(resource.id, { tags: tags })
+
+      removeEntity(prefabEntity)
+      await EditorControlFunctions.removeObject([entity])
```
dinomut1: calling removeObject twice. You don't need await, since the reactor handles the async logic

dinomut1 (changes requested):

JT00y:
> I don't think we want to be creating new editor control functions for something this specific. Here's a snippet of code you can add to `CreatePrefabPanelDialog.tsx` that will do the same thing, but using the existing control functions:
> 
> ```
> EditorControlFunctions.removeObject([entity])
> const reactor = startReactor(() => {
>   const documentState = useHookstate(getMutableState(GLTFDocumentState))
>   useEffect(() => {
>     if (!entityExists(entity)) {
>       const { entityUUID } = EditorControlFunctions.createObjectFromSceneElement(
>         [
>           { name: ModelComponent.jsonID, props: { src: fileURL } },
>           { name: TransformComponent.jsonID, props: { position, rotation, scale } }
>         ],
>         parentEntity
>       )
>       getMutableState(SelectionState).selectedEntities.set([entityUUID])
>       reactor.stop()
>     }
>   }, [documentState])
>   return null
> })
> ```
> 
> So we use EditorControlFunctions.removeObject to delete the original entity, then set up a reactor which waits for that entity to be deleted, then we use the normal createObjectFromSceneElement to add the prefab reference.

for entity without model component, the gltfdoument state not update, replace documentState to root entity nodes data

dinomut1 (approved):

PR #10782, opened 2024-07-31
IR-3142 - fixing a bug where if an audio were set to Single mode it would never set up the audio

SamMazerIR:
## Summary
fixing a bug where if an audio were set to Single mode it would never set up the audio


packages/engine/src/scene/components/MediaComponent.ts
```diff
@@ -360,10 +360,22 @@ export function MediaReactor() {
...
       if (media.resources.value.every((resource) => !resource)) return // if all resources are empty, we dont move to next track
 
+      const mediaElement = getOptionalComponent(entity, MediaElementComponent)
       const track = media.track.value
       let nextTrack = getNextTrack(track, media.resources.length, media.playMode.value)
-      if (nextTrack === -1) return
-      let path = media.resources[nextTrack].value
+
+      //check if we haven't set up for single play yet, or if our sources don't match the new resources
+      //TODO make this more robust in a refactor, feels very error prone with edge cases
```
DanielBelmes: /** @todo */ please

packages/engine/src/scene/components/MediaComponent.ts
```diff
@@ -360,10 +360,22 @@ export function MediaReactor() {
...
+      if (
+        nextTrack === -1 &&
+        mediaElement &&
+        mediaElement.element &&
+        media.resources.value.length > 0 &&
```
DanielBelmes: Can use ?. operator

AidanCaruso (approved):

PR #10691, opened 2024-07-22
studio: update scene item UI

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-3315

## References
closes #_insert number here_

PR #10790, opened 2024-07-31
Fix scene menu position

SYBIOTE:
This pull request fixes the position of the scene menu. Previously, the menu was positioned at the bottom of the element, but now it is positioned at the top. This change ensures that the menu is displayed correctly and improves the user experience.

in combination with #10691


hanzlamateen (changes requested):
> https://github.com/theinfinitereality/eepro-multitenancy/pull/352#discussion_r1686384397

@SYBIOTE I dont see it being cut off. Can you ellaboratre it further
image

Moreover, since this same sceneItem is being used in console, and in console we are opening all context menus to bottom right. Therefore for if in studio its required to be on top, then that can be handled through some prop?

SYBIOTE:
it cuts off on smaller screens, like this

hanzlamateen:
@SYBIOTE I can't access the image you referred. Maybe you can provide a property in scene item through which both these behaviors can be supported.

SYBIOTE:
this does not seem relevant anymore, closing the PR

PR #10793, opened 2024-08-01
[IR-3459] Unit Tests: Renderer Components coverage to 80%

heysokam:
## References
https://tsu.atlassian.net/browse/IR-3459


packages/spatial/src/renderer/components/MeshComponent.ts
```diff
@@ -40,7 +40,7 @@ import { useResource } from '../../resources/resourceHooks'
 import { addObjectToGroup, removeObjectFromGroup } from './GroupComponent'
 
 export const MeshComponent = defineComponent({
-  name: 'Mesh Component',
+  name: 'MeshComponent',
```
MichaelEstes, line 44: Are we sure these names aren't being used for the UI in the studio?

HexaField: They aren't. We have i18n that uses the json ID

PR #10796, opened 2024-08-01
IR-3495: Fixes duplicate row write to `static-resource` on SaveAs

CITIZENDOT:
## Summary
Currently, On **Save As** operation, two rows are written to `static-resource` table.
- One row has `type` column set to `file`
- Other has `scene`
Both rows has the same key (path), and same hash.

This PR removes the duplicate write that occurs during **Save** & **SaveAs** operations.

## References
closes #_insert number here_

### DIFF
--- a/packages/editor/src/functions/assetFunctions.ts
+++ b/packages/editor/src/functions/assetFunctions.ts
@@ -89,7 +89,7 @@ export const inputFileWithAddToScene = ({
     el.click()
   })
 
-export const uploadProjectFiles = (projectName: string, files: File[], paths: string[], onProgress?) => {
+export const uploadProjectFiles = (projectName: string, files: File[], paths: string[], args?: object) => {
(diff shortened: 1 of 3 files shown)


barankyle:
I'm not sure if this is within the scope of this issue, but can this fix also be done for New Scene? It seems like that is still making two copies.

CITIZENDOT:
> It seems like that is still making two copies

Where is this happening?

barankyle:
> > It seems like that is still making two copies
> 
> Where is this happening?

In the studio, if I do Hamburger menu -> New Scene

CITIZENDOT:
> In the studio, if I do Hamburger menu -> New Scene

Unable to reproduce this (probably fixed?). Also, same functionality of the changes in this PR are merged into dev. Closing this in favour of them.

barankyle:
Indeed, neither New Scene nor Save As are making duplicates for me.

PR #10800, opened 2024-08-01
[IR-3211] studio: show spinner over viewport when loading dropped items

dtlehrer:
## Summary
[IR-3211]

## References
closes #_insert number here_

## QA Steps

[IR-3211]: https://tsu.atlassian.net/browse/IR-3211?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/ui/src/components/editor/panels/Viewport/container/index.tsx
```diff
@@ -61,22 +62,48 @@ import TransformSpaceTool from '../tools/TransformSpaceTool'
 
 const ViewportDnD = ({ children }: { children: React.ReactNode }) => {
   const projectName = useMutableState(EditorState).projectName
+  const [isLoading, setIsLoading] = useState(false)
```
dinomut1: Can we use hookstate for this data instead of default react state?

dinomut1 (changes requested):

packages/ui/src/components/editor/panels/Viewport/overlay/LoadingOverlay.tsx
```diff
@@ -0,0 +1,33 @@
+import { useHookstate } from '@etherealengine/hyperflux'
+import React, { useEffect } from 'react'
+
+interface LoadingOverlayProps {
+  isLoading: boolean
+}
+
+const LoadingOverlay: React.FC<LoadingOverlayProps> = ({ isLoading }) => {
```
hanzlamateen: Why you need a new control and why not use `LoadingView` control? Is there a particular usecase that can not be handled by LoadingView?

hanzlamateen (changes requested):

packages/ui/src/components/editor/panels/Viewport/overlay/LoadingOverlay.tsx
dtlehrer: refactored to use `LoadingView`, thanks

packages/ui/src/primitives/tailwind/LoadingView/index.tsx
```diff
@@ -36,7 +36,8 @@ const LoadingView = ({
   fullSpace,
   containerClassname,
   titleClassname,
-  spinnerOnly
+  spinnerOnly,
+  delaySpinner = false
```
hanzlamateen, line 40: I dont think we need this delay functionality to be in this control. It can stay in viewport file as this delay is a rare functionality And shouldnt be part of loadingView control

packages/ui/src/components/editor/panels/Viewport/container/index.tsx
```diff
@@ -116,6 +158,12 @@ const ViewportDnD = ({ children }: { children: React.ReactNode }) => {
       className={twMerge('h-full w-full border border-white', isDragging ? 'border-4' : 'border-none')}
     >
       {children}
+      {isLoading.value && (
+        // immediately dim the viewport, then show LoadingView after delay
+        <div className="absolute inset-0 z-[80] flex items-center justify-center bg-black bg-opacity-50 transition-opacity duration-300">
+          <LoadingView fullSpace delaySpinner className="block h-12 w-12" />
+        </div>
```
hanzlamateen, line 165: Instead of using parent <div> here. You can use containerClass prop of LoadingView to manage container if required.

Also you can remove unused classes as I dont think we would require most of them i.e. flex, etc.

hanzlamateen (changes requested):

dtlehrer:
@hanzlamateen - the ticket asks to (1) immediately dim the viewport and then (2) show spinner 500ms later

I think that leaves us with three main options: either (1) utilize the containerClass prop of loadingView with the optional delaySpinner prop included.. (2) use a parent for the viewport dimming with optional delaySpinner prop on LoadingView ... or (3) control both the viewport dimming and spinner delay from the ViewportDnD component

I'd lean towards option 1 or 2 based on the requirements and the anticipation that there will be other times where the design team wants the spinner to be slightly delayed in the future, but open to your thoughts as well

hanzlamateen:
@HexaField your thoughts on this?

packages/ui/src/primitives/tailwind/LoadingView/index.tsx
HexaField: I agree.

HexaField:
I've left a question on the Jira ticket.

PR #10812, opened 2024-08-02
add name column in static-resources

aditya-mitra:
## Summary

- introduce name column in static-resources to find assets by their name

## References

https://tsu.atlassian.net/browse/IR-3155

## QA Steps

image

image

### DIFF
--- a/packages/client-core/src/admin/common/constants/resources.ts
+++ b/packages/client-core/src/admin/common/constants/resources.ts
@@ -27,7 +27,7 @@ import { t } from 'i18next'
 
 import { ITableHeadCell } from '../Table'
 
-type IdType = 'id' | 'key' | 'mimeType' | 'project' | 'action'
+type IdType = 'id' | 'name' | 'key' | 'mimeType' | 'project' | 'action'
 
 export type ResourceRowType = Record<IdType, string | JSX.Element | undefined>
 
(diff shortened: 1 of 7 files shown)


CITIZENDOT:
Instead of introducing another column, can we modify the search criteria to search only the last string?

Using regex in the query can work I think.

aditya-mitra:
> Instead of introducing another column, can we modify the search criteria to search only the last string?
> 
> Using regex in the query can work I think.

Regex would need to supported in our database - mysql. The other alternative is put such a regex in server or client on the database results. In this case, we would miss some search results because of pagination (for example, assume searching for _vase_ produces 3 results with _vase_ at the end of the string and 8 with _vase_ at the middle of the string. In a pagination set of 10, there is strong possibility we will only see 2 results instead of 3 because of last one being in another page)

CITIZENDOT:
MySQL & MariaDB both of them support `REGEXP`. It isn't supported by feathers query operator. So, some workaround has to be made with `action` and modifying query at the server.

For eg: In our database, below query searches through the file names, that has the word `sky` in them:

```sql
SELECT * FROM `static-resource` WHERE `key` REGEXP '\/[^\/]*(sky)[^\/]*$'
```

This will only search the string after the last slash, so only names are included in the search (project names, directories are not included).

image

Test link: https://regex101.com/r/M5BGAX/2

Edits: Escaped regex pattern, Added screenshot with some tests

HexaField:
Let's include some unit tests for this please

HexaField:
> MySQL & MariaDB both of them support `REGEXP`. It isn't supported by feathers query operator. So, some workaround has to be made with `action` and modifying query at the server.
> 
> For eg: In our database, below query searches through the file names, that has the word `sky` in them:
> 
> ```sql
> SELECT * FROM `static-resource` WHERE `key` REGEXP '/[^/]*(sky)[^/]*$'
> ```
> 
> This will only search the string after the last slash, so only names are included in the search (project names, directories are not included).

Does postgres support this?

CITIZENDOT:
It does: https://www.postgresql.org/docs/current/functions-matching.html#FUNCTIONS-POSIX-REGEXP. Syntax is a bit different.

PR #10854, opened 2024-08-04
Clamp audio slider between 0 - 2

SYBIOTE:
The audio slider in the MediaNodeEditor component was not properly clamped between the values of 0 and 2. This pull request fixes that issue by adding a clamp function to ensure the slider value stays within the desired range.


packages/ui/src/components/editor/properties/media/index.tsx
```diff
@@ -99,9 +100,9 @@ export const MediaNodeEditor: EditorComponentType = (props) => {
           min={0}
           max={100}
```
HexaField, line 101: wouldn't it be better to change the min/max/step ?

HexaField (changes requested):

packages/ui/src/components/editor/properties/media/index.tsx
SYBIOTE: the min max is for the slider UI, is we want the values to be between 0 - 100 on the UI
, we need to use a converter
to convert between 0-2 on the audio scale to 0 - 100 on the UI

SYBIOTE: we can keep the values between 0 - 2 on the UI as well, if needed, is that preffered?

PR #10873, opened 2024-08-05
Physics reactivity for collider component updates and order dependency resolution

AidanCaruso:
## Summary

note: fix these tests

## References
closes #_insert number here_

### DIFF
--- a/packages/spatial/src/physics/components/ColliderComponent.tsx
+++ b/packages/spatial/src/physics/components/ColliderComponent.tsx
@@ -83,13 +83,13 @@ export const ColliderComponent = defineComponent({
     const component = useComponent(entity, ColliderComponent)
     const transform = useComponent(entity, TransformComponent)
     const rigidbodyEntity = useAncestorWithComponent(entity, RigidBodyComponent)
-    const rigidbodyComponent = useOptionalComponent(rigidbodyEntity, RigidBodyComponent)
     const physicsWorld = Physics.useWorld(entity)
     const triggerComponent = useOptionalComponent(entity, TriggerComponent)
     const hasCollider = useState(false)
+    const physicsWorldRigidbody = Physics.useWorld(entity)?.Rigidbodies[entity]
 
     useEffect(() => {
-      if (!rigidbodyComponent || !physicsWorld) return
+      if (!rigidbodyEntity || !physicsWorld) return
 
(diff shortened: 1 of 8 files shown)


packages/spatial/src/physics/components/ColliderComponent.tsx
```diff
@@ -83,13 +83,13 @@ export const ColliderComponent = defineComponent({
...
     const rigidbodyEntity = useAncestorWithComponent(entity, RigidBodyComponent)
-    const rigidbodyComponent = useOptionalComponent(rigidbodyEntity, RigidBodyComponent)
     const physicsWorld = Physics.useWorld(entity)
     const triggerComponent = useOptionalComponent(entity, TriggerComponent)
     const hasCollider = useState(false)
+    const physicsWorldRigidbody = Physics.useWorld(entity)?.Rigidbodies[entity]
 
     useEffect(() => {
-      if (!rigidbodyComponent || !physicsWorld) return
+      if (!rigidbodyEntity || !physicsWorld) return
```
speigg, line 92: this needs to be `rigidbodyComponent.value`

packages/spatial/src/physics/components/ColliderComponent.tsx
```diff
@@ -101,7 +101,7 @@ export const ColliderComponent = defineComponent({
         Physics.removeCollider(physicsWorld, entity)
         hasCollider.set(false)
       }
-    }, [physicsWorld, component.shape, rigidbodyEntity, !!rigidbodyComponent, transform.scale])
+    }, [physicsWorld, component.shape, rigidbodyEntity, !!physicsWorldRigidbody, transform.scale])
```
the-reality-engineer, line 104: !!physicsWorldRigidbody?.value

the-reality-engineer (changes requested):

packages/spatial/src/physics/components/ColliderComponent.tsx
mrhegemon: This doesnt return the wrapped state.

the-reality-engineer: nvm, useWorld() doesn't return a state-wrapped physics world..

mrhegemon (approved):

PR #10878, opened 2024-08-05
feat(admin): crash report page [IR-3129]

mikeplascdev:
## Summary
- create a crash report page

## References
closes #_insert number here_

## QA Steps

- Go to [ir-admin](https://ir-engine-mt-dev.theinfinitereality.io/admin)
- The **Crash report** option should appear in the sidebar

## Evidence
report-page

## Jira
[IR-3129](https://tsu.atlassian.net/browse/IR-3129)

[IR-3129]: https://tsu.atlassian.net/browse/IR-3129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


barankyle (changes requested):
In order for the `VITE_*` environment variables, like the new VITE_METABASE_SITE_URL, to be included in the Docker build process, they need to be defined/passed in several places:

1. scripts/build_and_publish_package.sh (add a --build-arg for it to BOTH docker buildx build commands)
2. All of the Dockerfiles
- root Dockerfile
- dockerfiles/api/Dockerfile-api-client
- dockerfiles/client/Dockerfile-client
- dockerfiles/client/Dockerfile-client-serve-static
(These all need an ARG definition and then setting the same-named ENV from $ARG)
3. build_microk8s.sh (make an if/then/else block to set it from the ENV_VAR if present and add a --build-arg for it)
4. build-minikube.sh (same as build_microk8s)

If you search for VITE_ZENDESK_KEY and basically just copy/replace with VITE_METABASE_SITE_URL at all the places that shows up in those files, you should be good.

mikeplascdev:
Done @barankyle  013b32aba100029483d74e48b5ed3414c8223ef4

barankyle (approved):

packages/client-core/src/admin/DefaultAdminRoutes.tsx
```diff
@@ -150,5 +153,12 @@ export const DefaultAdminRoutes: Record<string, AdminRouteStateType> = {
     component: Channels,
     access: false,
     icon: <HiOutlineMegaphone />
+  },
+  crashReport: {
+    name: 'user:dashboard.crashReport',
+    scope: 'crashReport',
```
hanzlamateen: We dont have `'crashReport'` in our scope types. You can use `server` here

packages/server-core/src/integrations/metabase/metabase.hooks.ts
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+import appConfig from '@etherealengine/server-core/src/appconfig'
+import { Application, HookContext } from '@feathersjs/feathers/lib/declarations'
+import { disallow } from 'feathers-hooks-common'
+import { sign } from 'jsonwebtoken'
+
+const getMetabaseToken = (context: HookContext<Application>) => {
+  const payload = {
+    resource: { dashboard: 64 },
```
hanzlamateen: This dashboard ID, shouldnt be hardcoded.

Also, there is no security on hooks.

hanzlamateen (changes requested):
@mikeplascdev I think we should make `metabase` service to return dashboard urls instead of just serving the token.

I think we can merge the code from https://github.com/theinfinitereality/eepro-multitenancy/blob/78f4996494fd29491ccaed10888d0dfd824b30ac/services/account/account-metabase-url service and this into 1 service named metabase-url.

The service in core engine can serve the admin panel use case. Though for MT we can extend it using migration and admin panel extension to include project and account dashboard ids. Also hooks can be manipulated in MT project to check for appropriate permissions like https://github.com/theinfinitereality/eepro-multitenancy/blob/78f4996494fd29491ccaed10888d0dfd824b30ac/services/hooks/user/login.hooks.ts#L7

The implementation in MT also allows to easily change base metabase url (https://data-analytics-metabase-prd.theinfinitereality.io) which wouldnt be possible in this hardcoded approach. I think balazs mentioned that soon this URL will only be valid for prod and for other envs we will need to change it. Also the dashboard IDs are going to be different for each of the deployment environment.

packages/client-core/src/admin/DefaultAdminRoutes.tsx
mikeplascdev: done b16316005f9954108a279bc4781f33bd5357a99b

packages/server-core/src/integrations/metabase/metabase.hooks.ts
mikeplascdev: done 38dbf63ccce38429fa062d6c149d8b3a52f0a6d1

packages/server-core/tsconfig.json
```diff
@@ -40,6 +40,8 @@
     "./*.ts",
     "./src/**/*.ts",
     "./src/**/*.tsx",
-    "../common/src/schemas/setting/account-project.schema.ts"
+    "../common/src/schemas/setting/account-project.schema.ts",
```
HexaField: please fix these

HexaField (changes requested):

hanzlamateen (approved):

PR #10890, opened 2024-08-06
Add script to generate Apple SSO client Secret

muradkhateeb78:


barankyle:
Can you send me, via direct message, all of the files/keys I'd need to test this?

barankyle (approved):

PR #10903, opened 2024-08-06
IR-3645: Fix avatars falling through floor on scene load

dinomut1:
Closes https://tsu.atlassian.net/browse/IR-3645
Fixes this by making sure that every model has been fully initialized, not just that it's finished downloading the data

### DIFF
--- a/packages/engine/src/gltf/GLTFComponent.tsx
+++ b/packages/engine/src/gltf/GLTFComponent.tsx
@@ -30,18 +30,21 @@ import { parseStorageProviderURLs } from '@etherealengine/common/src/utils/parse
 import {
   defineComponent,
   Entity,
+  EntityUUID,
   getComponent,
   getMutableComponent,
   getOptionalComponent,
   hasComponent,
   useComponent,
   useEntityContext,
-  useQuery
+  useQuery,
+  UUIDComponent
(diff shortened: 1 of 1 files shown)


AidanCaruso (approved):

packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -83,7 +86,19 @@ const ResourceReactor = (props: { documentID: string; entity: Entity }) => {
...
-
+    const document = getState(GLTFDocumentState)[props.documentID]
+    const modelNodes = document.nodes?.filter((node) => !!node.extensions?.[ModelComponent.jsonID])
+    if (modelNodes) {
+      for (const node of modelNodes) {
+        //check if an entity exists for this node, and has a model component
+        const uuid = node.extensions![UUIDComponent.jsonID] as EntityUUID
+        if (!UUIDComponent.entitiesByUUIDState[uuid]) return
+        const entity = UUIDComponent.entitiesByUUIDState[uuid].value
+        const model = getOptionalComponent(entity, ModelComponent)
```
the-reality-engineer, line 97: It doesnt look like this effect is due to rerun if the model component is added; is that correct?

PR #10905, opened 2024-08-07
IR-3461 User login service to store user agent and ip address in database

hurairahmateen:
## Summary
- Created user-login service.
- Entry is created for user-login service when successful login using SSO or magic link.
- Removal of lastLogin.

## References
https://tsu.atlassian.net/browse/IR-3461


packages/common/src/schemas/user/user-login.schema.ts
```diff
@@ -0,0 +1,99 @@
...
+  {
+    id: Type.String({
+      format: 'uuid'
+    }),
+    userId: TypedString<UserID>({
+      format: 'uuid'
+    }),
+    userAgent: Type.String(),
+    ipAddress: Type.String(),
+    port: Type.Optional(Type.Integer()),
```
barankyle: It doesn't appear that the user's port is actually captured anywhere, nor `secure`. Are there plans to capture these? If not, do we need to have these columns at all?

barankyle (approved):

packages/common/src/schemas/user/user-login.schema.ts
hanzlamateen: For primus requests we do get this information in `params.forwarded`. Though we can remove it as its not much useful.

hanzlamateen (approved):

PR #10907, opened 2024-08-07
Unpinned version of @aws-sdk/client-s3

barankyle:
## Summary

Pinned version was now causing type errors because it was out of sync with other aws-sdk packages, and latest versions do not have the call hang issues that made us pin it to begin with.

## References
closes #_insert number here_

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,7 +150,7 @@
   },
   "dependencies": {
     "@aws-sdk/client-ecr": "^3.319.0",
-    "@aws-sdk/client-s3": "3.614.0",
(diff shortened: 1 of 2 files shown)


HexaField:
Hmm, maybe we should pin all versions together?

barankyle:
It seems like 99.9% of the time, having them be the latest version is fine. It might be worth pinning them all to something - I'm not sure what improvements they make in all these minor versions, but they sure do put out new minor versions frequently.

HexaField (approved):

hanzlamateen (approved):

PR #10908, opened 2024-08-07
studio: open context menu on file menu right click

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-3568
## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/components/editor/panels/Files/browserGrid/index.tsx
+++ b/packages/ui/src/components/editor/panels/Files/browserGrid/index.tsx
@@ -294,7 +294,6 @@ export function FileBrowserItem({
   }
 
   const pasteContent = async () => {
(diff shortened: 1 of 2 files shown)


aditya-mitra (changes requested):
almost fixed, just a small change - if we click on the someplace below the drop area, right click is non-functional. might need to change the div height in accordance with rc-dock

https://github.com/user-attachments/assets/1dd91f21-6edb-4a2d-84ce-4cf1513299d6

achen5671:
> almost fixed, just a small change - if we click on the someplace below the drop area, right click is non-functional. might need to change the div height in accordance with rc-dock
> 
>  2024-08-12.at.18.02.47.mp4

resolved

aditya-mitra (approved):
awesome

PR #10916, opened 2024-08-07
Component dependencies

MichaelEstes:
## Summary
I was getting a race condition on some scenes where the GLTF progress is never set to 100 because the models being checked in the useEffect are not reactive.
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/gltf/GLTFComponent.tsx
+++ b/packages/engine/src/gltf/GLTFComponent.tsx
@@ -29,7 +29,6 @@ import React, { useEffect } from 'react'
 import { parseStorageProviderURLs } from '@etherealengine/common/src/utils/parseSceneJSON'
 import {
   Component,
-  ComponentDependencyMap,
   ComponentJSONIDMap,
   ComponentType,
   defineComponent,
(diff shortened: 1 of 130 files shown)


packages/engine/src/gltf/GLTFComponent.tsx
```diff
@@ -40,17 +43,40 @@ import {
...
+} as Record<string, string[]>
+
+type ComponentDependencies = Record<ComponentType<typeof UUIDComponent>, Component[]>
+
+const buildComponentDependencies = (json: GLTF.IGLTF) => {
+  const dependencies = {} as ComponentDependencies
+  if (!json.nodes) return dependencies
+  for (const node of json.nodes) {
+    if (!node.extensions || !node.extensions[UUIDComponent.jsonID]) continue
+    const uuid = node.extensions[UUIDComponent.jsonID] as ComponentType<typeof UUIDComponent>
```
HexaField: `node.extensions[UUIDComponent.jsonID]` here is not of type `ComponentType<typeof UUIDComponent>` but just `EntityUUID`

HexaField (dismissed):

AidanCaruso (approved):

PR #10917, opened 2024-08-08
Fixed bugs in identity-provider service

barankyle:
## Summary

Users were able to create an identity-provider on another user. This would return a valid JWT for that other user that could authenticate requests as them. Enhanced identity-provider hooks for creation so that users can only create identity-providers for themselves, and unauthenticated requests cannot create identity-providers on existing users.

Also prevented the creation of guest identity-providers on existing users, because there's no reason for it and it seems like something that could introduce bugs.

Added tests to ensure that one cannot create an identity-provider on another user. Updated a number of tests to work around some issues this causes - oauth providers are seen as 'guest' providers, so tests that were creating github sign-ins were now failing.

Some identity-provider tests were not actually running to completion. assert.rejects needs to take in an async function whose contents are await'ed, and the assert.rejects itself needs to be await'ed, or else the test suite will mistakenly assume it succeeded when it's still running.

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/projects/project-branches/project-branches.test.ts
+++ b/packages/server-core/src/projects/project-branches/project-branches.test.ts
@@ -31,11 +31,12 @@ import { projectBranchesPath } from '@etherealengine/common/src/schemas/projects
 import { ScopeType } from '@etherealengine/common/src/schemas/scope/scope.schema'
 import { avatarPath } from '@etherealengine/common/src/schemas/user/avatar.schema'
 import { identityProviderPath } from '@etherealengine/common/src/schemas/user/identity-provider.schema'
-import { userApiKeyPath, UserApiKeyType } from '@etherealengine/common/src/schemas/user/user-api-key.schema'
+import { UserApiKeyType, userApiKeyPath } from '@etherealengine/common/src/schemas/user/user-api-key.schema'
 import { UserName, userPath } from '@etherealengine/common/src/schemas/user/user.schema'
 import { destroyEngine } from '@etherealengine/ecs/src/Engine'
+import { identityProviderDataResolver } from '../../user/identity-provider/identity-provider.resolvers'
 
-import { Application } from '../../../declarations'
+import { Application, HookContext } from '../../../declarations'
 import { createFeathersKoaApp } from '../../createApp'
 
 describe('project-branches.test', () => {
@@ -70,12 +71,15 @@ describe('project-branches.test', () => {
 
     testUserApiKey = await app.service(userApiKeyPath).create({ userId: testUser.id })
 
-    await app.service(identityProviderPath).create(
-      {
-        type: 'github',
-        token: `test-token-${Math.round(Math.random() * 1000)}`,
-        userId: testUser.id
-      },
+    await app.service(identityProviderPath)._create(
+      await identityProviderDataResolver.resolve(
+        {
+          type: 'github',
+          token: `test-token-${Math.round(Math.random() * 1000)}`,
+          userId: testUser.id
+        },
+        {} as HookContext
+      ),
       getParams()
     )
   })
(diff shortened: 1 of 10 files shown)


packages/server-core/src/projects/project-branches/project-branches.test.ts
```diff
@@ -70,12 +73,15 @@ describe('project-branches.test', () => {
 
     testUserApiKey = await app.service(userApiKeyPath).create({ userId: testUser.id })
 
-    await app.service(identityProviderPath).create(
+    await app.service(identityProviderPath)._create(
```
hanzlamateen, line 74:
```
await app.service(identityProviderPath)._create(identityProviderDataResolver.resolve(
      {
        type: 'github',
        token: `test-token-${Math.round(Math.random() * 1000)}`,
        userId: testUser.id
      }),
```
Instead of populating id, dates manually, you can call resolver using `identityProviderDataResolver.resolve` method to populate values.

packages/server-core/src/projects/project-builder-tags/project-builder-tags.test.ts
```diff
@@ -70,12 +73,15 @@ describe('project-builder-tags.test', () => {
 
     testUserApiKey = await app.service(userApiKeyPath).create({ userId: testUser.id })
 
-    await app.service(identityProviderPath).create(
+    await app.service(identityProviderPath)._create(
```
hanzlamateen, line 74: You can use `identityProviderDataResolver.resolve()` method here to populate default values.

packages/server-core/src/projects/project-check-source-destination-match/project-check-source-destination-match.test.ts
```diff
@@ -80,12 +83,15 @@ describe('project-check-source-destination-match.test', () => {
 
     testUserApiKey = await app.service(userApiKeyPath).create({ userId: testUser.id })
 
-    await app.service(identityProviderPath).create(
+    await app.service(identityProviderPath)._create(
```
hanzlamateen, line 84: You can use `identityProviderDataResolver.resolve()` method here to populate default values.

packages/server-core/src/projects/project-check-unfetched-commit/project-check-unfetched-commit.test.ts
```diff
@@ -71,12 +74,15 @@ describe('project-check-unfetched-commit.test', () => {
 
     testUserApiKey = await app.service(userApiKeyPath).create({ userId: testUser.id })
 
-    await app.service(identityProviderPath).create(
+    await app.service(identityProviderPath)._create(
```
hanzlamateen, line 75: You can use `identityProviderDataResolver.resolve()` method here to populate default values.

packages/server-core/src/projects/project-commits/project-commits.test.ts
```diff
@@ -69,12 +72,15 @@ describe('project-commits.test', () => {
 
     testUserApiKey = await app.service(userApiKeyPath).create({ userId: testUser.id })
 
-    await app.service(identityProviderPath).create(
+    await app.service(identityProviderPath)._create(
```
hanzlamateen, line 73: You can use `identityProviderDataResolver.resolve()` method here to populate default values. Same for other instances

packages/server-core/src/user/identity-provider/identity-provider.hooks.ts
```diff
@@ -108,27 +150,52 @@ async function checkOnlyIdentityProvider(context: HookContext<IdentityProviderSe
...
+  try {
+    if (userId) existingUser = await context.app.service(userPath).get(userId)
+  } catch (err) {
+    //
   }
 
-  if (!userId) {
-    if ((context.actualData as IdentityProviderData).type === 'guest') {
-      return
+  let isAuthenticated = false
```
hanzlamateen: Just wondering if the code is similar to line 84 of same file. Can it be moved to a common method so that both code blocks can call it? Just a suggestion.

hanzlamateen (approved):

PR #10919, opened 2024-08-08
Add animation to tooltips

CITIZENDOT:
## Summary

Adds a simple bouncy animation to tooltips.

## References
closes #_insert number here_

## QA Steps
https://github.com/user-attachments/assets/150341f4-ab25-43e8-992d-2bfbb5aeb64c


HexaField (dismissed):
i'd very much prefer this to not be a global class. can we add this to just the tooltip where it's used?

CITIZENDOT:
> i'd very much prefer this to not be a global class. can we add this to just the tooltip where it's used?

moved to component scope

aditya-mitra (approved):
Looks good ✅

PR #10920, opened 2024-08-08
fix(ir-engine): implemented scene name length validation

jonathan-casarrubias:
## Summary
Implemented scene name length specific validation error message

## References
closes #_insert number here_

### DIFF
--- a/packages/client-core/i18n/en/editor.json
+++ b/packages/client-core/i18n/en/editor.json
@@ -72,8 +72,7 @@
     "unknownStatus": "Unknown Status",
     "CORS": "Possibly a CORS error",
     "urlFetchError": "Failed to fetch \"{{url}}\"",
-    "invalidSceneName": "Invalid input: Only alphanumeric characters, hyphens, and underscores are allowed.",
-    "invalidSceneNameLength": "Invalid input: Name must be between 4 and 64 characters"
+    "invalidSceneName": "Scene name must be 4-64 characters long, using only alphanumeric characters, hyphens, and underscores."
   },
   "viewport": {
     "title": "Viewport",
--- a/packages/client-core/i18n/en/editor.json
+++ b/packages/client-core/i18n/en/editor.json
@@ -72,7 +72,8 @@
     "unknownStatus": "Unknown Status",
     "CORS": "Possibly a CORS error",
     "urlFetchError": "Failed to fetch \"{{url}}\"",
-    "invalidSceneName": "Invalid input: Only alphanumeric characters, hyphens, and underscores are allowed."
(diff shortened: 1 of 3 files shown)


mikeplascdev (approved):

kattsushi (approved):

packages/common/src/utils/validateSceneName.ts
```diff
@@ -28,3 +28,7 @@ import { VALID_SCENE_NAME_REGEX, WINDOWS_RESERVED_NAME_REGEX } from '@etherealen
 export default function isValidSceneName(sceneName: string) {
   return !WINDOWS_RESERVED_NAME_REGEX.test(sceneName) && VALID_SCENE_NAME_REGEX.test(sceneName)
 }
+
+export function isValidSceneNameLength(sceneName: string) {
+  return sceneName.length > 3 && sceneName.length < 65
+}
```
CITIZENDOT: This check is not necessary because `VALID_SCENE_NAME_REGEX` enforces min & max checks in the regex itself.

CITIZENDOT (changes requested):

packages/client-core/i18n/en/editor.json
```diff
@@ -72,7 +72,8 @@
     "unknownStatus": "Unknown Status",
     "CORS": "Possibly a CORS error",
     "urlFetchError": "Failed to fetch \"{{url}}\"",
-    "invalidSceneName": "Invalid input: Only alphanumeric characters, hyphens, and underscores are allowed."
+    "invalidSceneName": "Invalid input: Only alphanumeric characters, hyphens, and underscores are allowed.",
+    "invalidSceneNameLength": "Invalid input: Name must be between 4 and 64 characters"
```
CITIZENDOT: there is no need of a new message. Instead, we can rewrite the first message as: 
> Scene name must be 4-64 characters long, using only alphanumeric characters, hyphens, and underscores.

packages/editor/src/components/dialogs/SaveSceneDialog.tsx
```diff
@@ -98,6 +98,10 @@ export const SaveNewSceneDialog = (props: { onConfirm?: () => void; onCancel?: (
   const inputError = useHookstate('')
 
   const handleSubmit = async () => {
+    if (!isValidSceneNameLength(inputSceneName.value)) {
+      inputError.set(t('editor:errors.invalidSceneNameLength'))
+      return
+    }
```
CITIZENDOT: this is not necessary as well

CITIZENDOT (changes requested):

hanzlamateen (approved):

PR #10921, opened 2024-08-08
[IR-3547] studio: show unsaved changes dialog when switching between scenes

dtlehrer:
## Summary
[IR-3547]
- creates `confirmedSaveIfModified` function, which can now be used to check for unsaved changes and display the Unsaved Changes confirmation dialog across the app
- uses the `confirmedSaveIfModified` function when clicking on the `SceneItem` component to switch between scenes

https://github.com/user-attachments/assets/6bfb8dfc-9472-4bed-97e1-68839a093392

## References
closes #_insert number here_

## QA Steps

[IR-3547]: https://tsu.atlassian.net/browse/IR-3547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/client-core/i18n/en/editor.json
+++ b/packages/client-core/i18n/en/editor.json
@@ -1360,7 +1360,10 @@
       "lbl-thumbnail": "Generate thumbnail & envmap",
       "lbl-confirm": "Save Scene",
       "info-confirm": "Are you sure you want to save the scene?",
-      "info-question": "Do you want to save the current scene?"
+      "info-question": "Do you want to save the current scene?",
+      "unsavedChanges": {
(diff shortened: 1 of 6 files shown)


aditya-mitra (changes requested):
Finally this replaces the old alert

Can you also replace this alert on page close?
image

aditya-mitra (approved):
LGTM ✅

(I modified the beforeUnloadEvent to use browser's alert instead of our own alert because the former solution is cross-browser friendly and also gives the user a way to force close the window)

PR #10934, opened 2024-08-09
moved the rename / delete menu up to prevent the delete from being hidden

MbfloydIR:
moved the rename / delete menu up to prevent the delete from being hidden
https://tsu.atlassian.net/browse/IR-3754

## References
closes #_insert number here_


hanzlamateen (changes requested):
This somewhat breaks how we show context menu in Dashboard/Console. Because same SceneItem control is used there. 

Is it possible to have this behavior, if certains prop was passed to SceneItem control so that both behaviors can be supported.

image

hanzlamateen (approved):

PR #10936, opened 2024-08-09
clamped the scene name to be 20 characters with an ellipse and a tooltip

MbfloydIR:
## Summary
clamped the scene name to be 20 characters with an ellipse and a tooltip
https://tsu.atlassian.net/browse/IR-3667
## References
closes #_insert number here_


DanielBelmes (changes requested):
Hey Michael you should be able to simplify this using the CSS overflow property and https://www.w3schools.com/cssref/css3_pr_text-overflow.php

So we can set a fix css width on this HTML element and then set this other css property that will tell the browser to display elipsis when it is too large for that element.

PR #10940, opened 2024-08-12
fix(ir-engine): fixed missing email parameter in ZenDesk JWT

jonathan-casarrubias:
## Summary

Fixed missing email property from signed ZenDesk JWT 

## References
closes #_insert number here_

### DIFF
--- a/packages/server-core/src/integrations/zendesk/zendesk.hooks.ts
+++ b/packages/server-core/src/integrations/zendesk/zendesk.hooks.ts
@@ -29,11 +29,14 @@ import { disallow } from 'feathers-hooks-common'
 import { sign } from 'jsonwebtoken'
 
 const getZendeskToken = (context: HookContext<Application>) => {
+  const { email } = context.params.user.identityProviders.find((ip) => ip.email)
+
   context.result = sign(
     {
       scope: 'user',
       external_id: context.params.user.id,
-      name: context.params.user.name
+      name: context.params.user.name,
+      email
     },
     appConfig.zendesk.secret!,
     {


mikeplascdev (approved):

packages/server-core/src/integrations/zendesk/zendesk.hooks.ts
```diff
@@ -29,11 +29,14 @@ import { disallow } from 'feathers-hooks-common'
 import { sign } from 'jsonwebtoken'
 
 const getZendeskToken = (context: HookContext<Application>) => {
+  const { email } = context.params.user.identityProviders.find((ip) => ip.email)
```
HexaField, line 32: i don't think this will be populated for users who logged in with any other Oauth, or if the email field is populated in multiple identity providers, how do we decide which one to use?

HexaField (changes requested):

packages/server-core/src/integrations/zendesk/zendesk.hooks.ts
barankyle: Apple/Google/Github should all have an email after https://github.com/EtherealEngine/etherealengine/pull/10894. But it looks like Twitter/FB/Discord/LinkedIn were skipped since we're not using them on the platform right now, so that's still a concern. And deciding which one to use is a good point.

hanzlamateen: Regarding deciding, what if we prefer magiclink identity-provider over SSOs. And if its missing, then use any SSO.

Or maybe in SSO we can give preference to google > apple > github?

Or is there any other way?

jonathan-casarrubias: What's the suggestion here then?

jonathan-casarrubias: Hey so, I have locally created 2 different users, one through the magic link process and another using github oauth social login, I've verified that either way, the user's "email" property is being populated, see screenshot:

image

Based on the fact that email address is always available in the email property, I believe that it's safe to use the email property as single source of truth.

Regarding deciding which email to pick, I don't see any difference between picking the first against prioritizing selections, either we are making an assumption which by nature can be right or wrong regardless.

Unfortunately asking the user isn't an option since that's exactly what we are trying to avoid here, so if we can't ask the user and there's no difference nor guarantees between different assumptions, IMO the right way to solve this is by having  a "primary contact" kind of feature, like having a "Primary Shipping Address" in Amazon, "Primary Payment Card" in Uber and so on, that should be configured within a user's Profile and then using the primary user's selection when needed, like in this case.

That's the only way I see to fulfill the expectation, while maintaining the user's right to decide what's the personal preference.

hanzlamateen: @jonathan-casarrubias in your example you used magic-link and github which were associated with same email address. But what if, for magic-link you used jon@test.com while the github account that you used had jon@ir.com associated with it. In that case the email field would have different values and thats the inconsistency we are referring to. Hope it clears out the confusion.
> If you want I can post a video depicting this scenario.

I agree regarding your primary email address thing. For now we can go ahead with if there is any record with identity provider `type === 'email'` (this is the case when magic-link was used). If this identity provider is there, then use the email from it. Otherwise use email from the random one.

PR #10949, opened 2024-08-13
IR-3921: Copying resource should create a static-resource entry

CITIZENDOT:
## Summary
Currently, if we copy/paste a resource in studio's file browser, it does not create an entry in static-resource table. But it updates the existing entry with new key. This PR fixes it by identifying copy operations and creating a fresh entry for them.

## References
closes https://tsu.atlassian.net/browse/IR-3921


HexaField:
Nice, let's add a regression test for this please.

barankyle (approved):

hanzlamateen (approved):

PR #10951, opened 2024-08-13
studio: fix infinite scrolling

achen5671:
## Summary
https://tsu.atlassian.net/browse/IR-3743
## References
closes #_insert number here_

### DIFF
--- a/packages/ui/src/components/editor/panels/Assets/container/index.tsx
+++ b/packages/ui/src/components/editor/panels/Assets/container/index.tsx
@@ -450,8 +450,6 @@ const AssetPanel = () => {
   const assetsPreviewContext = useHookstate({ selectAssetURL: '' })
   const parentCategories = useHookstate<Category[]>([])
 
-  const requestCounter = useRef(0)
-
   const mapCategories = useCallback(() => {
     categories.set(mapCategoriesHelper(collapsedCategories.value))
   }, [categories, collapsedCategories])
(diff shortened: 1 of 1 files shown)


DanielBelmes (changes requested):
I'm still seeing the same bug. When I switch between Image to Model. The first load of images still appear and then it starts loading kits in. (Example is switching from Model to Image)
image

dinomut1 (changes requested):
I'm also still seeing the original bug. As the infinite scroll tries to load additional assets, if you click on another category it still keeps the previously loaded assets and loads the new ones in addition

packages/ui/src/components/editor/panels/Assets/container/index.tsx
```diff
@@ -502,6 +506,7 @@ const AssetPanel = () => {
         .service(staticResourcePath)
         .find({ query })
         .then((resources) => {
+          if (requestCounter.current !== currentRequestId) return
```
HexaField: please use an abort controller here instead of a counter

HexaField (changes requested):

achen5671:
@dinomut1  @DanielBelmes hm weird, It seems to be working for me. I changed it to use abort controller as per @HexaField request and it's also seems to work. Can you guys rereview when you have a chance. ty!

dinomut1 (changes requested):
[Screencast from 08-14-2024 10:09:49 AM.webm](https://github.com/user-attachments/assets/43166deb-c30a-4b0b-bc02-da60ca5b2764)
still seeing it. If you click on several categories one after the other it happens pretty consistently

PR #10960, opened 2024-08-14
IR-3588 adding a tag component - ThumbnailLightTagComponent for filtering render settings light query

SamMazerIR:
## Summary

adding a tag component - ThumbnailLightTagComponent for filtering the query for light options in render settings primary light dropdown options

## References

[https://tsu.atlassian.net/browse/IR-3588](https://tsu.atlassian.net/browse/IR-3588)

## QA Steps

Open a scene
Open the Files tab
Go to the settings component
Go to Render Settings → Primary Light
Click to open the drop-down
Observe the list of lights


HexaField (dismissed):
adding a new component is unnecessary. we can filter the useQuery with something like SourceComponent which will not exist on the thumbnail / non-scene entities.

SamMazerIR:
> adding a new component is unnecessary. we can filter the useQuery with something like SourceComponent which will not exist on the thumbnail / non-scene entities.

tested and implemented this change. cheers

MichaelEstes (approved):

PR #10961, opened 2024-08-14
Updated JWT authentication to handle RSA signatures

barankyle:
## Summary

New authentication-setting fields for JWT algorithm and JWT public key. Keys should be inserted as strings with '\n' replacing new lines.

Replaced the last of the API.instance.client uses with Engine.instance.api

Added a new service that will return the JWT public key.

Resolves IR-3827

## References
closes #_insert number here_


packages/client-core/src/user/services/AuthService.ts
```diff
@@ -170,16 +172,16 @@ export interface LinkedInLoginForm {
  */
 async function _resetToGuestToken(options = { reset: true }) {
   if (options.reset) {
-    await API.instance.client.authentication.reset()
+    await API.authentication.reset()
   }
-  const newProvider = await Engine.instance.api.service(identityProviderPath).create({
+  const newProvider = await API.service(identityProviderPath).create({
```
hanzlamateen: Why we replaced `Engine.instance.api` here?

As I see this change at various places in this file.

packages/client-core/src/API.ts
```diff
@@ -44,7 +44,7 @@ export type FeathersClient = FeathersApplication<ServiceTypes> &
 /**@deprecated - use 'Engine.instance.api' instead */
 export class API {
   /**@deprecated - use 'Engine.instance.api' instead */
-  static instance: API
+  static instance
```
hanzlamateen: Can we remove this `instance` property?

I think there are some last usages of it in

https://github.com/EtherealEngine/ee-development-test-suite/blob/e395b63b9cd7ec7b87a629da83367a1207eb5d56/src/old%20examples/matchmaking.tsx#L15

https://github.com/EtherealEngine/ee-development-test-suite/blob/e395b63b9cd7ec7b87a629da83367a1207eb5d56/src/old%20examples/matchmaking2.tsx#L26

packages/server-core/src/createApp.ts
```diff
@@ -201,6 +201,7 @@ export const createFeathersKoaApp = (
   app.set('port', appConfig.server.port)
 
   app.set('paginate', appConfig.server.paginate)
+  console.log('authentication')
```
hanzlamateen: I think we can remove this console log

packages/server-core/src/user/jwt-public-key/jwt-public-key.hooks.ts
```diff
@@ -0,0 +1,78 @@
...
+    const origin = new URL(context.params.query.redirectUrl).origin
+
+    // We only allow redirect to the same origin as of client
+    if (origin !== config.client.url) {
+      throw new BadRequest(
+        `Invalid redirect URL: ${context.params.query.redirectUrl}. Only redirect URLs from the same origin are allowed.`
+      )
+    }
+  }
+}
```
hanzlamateen: I think this is irrelevant

hanzlamateen (changes requested):
1. It seems like some github checks are failing
2. Getting following error when running locally.
image

packages/client-core/src/user/services/AuthService.ts
barankyle: I was having trouble with the typing. I tried casting the export `Engine.instance.api` as our custom FeathersClient type, but that was continually throwing errors elsewhere. It seemed simpler to cast `Engine.instance.api` as FeathersClient just where we need to use the custom aspects of it.

packages/client-core/src/API.ts
barankyle: https://github.com/EtherealEngine/ee-development-test-suite/pull/99

packages/client-core/src/user/services/AuthService.ts
barankyle: I redid some things to make Engine.instance.api a union type, though I still have to cast any client-side instance that's going to call authentication-client services as our custom FeathersClient type, or else TypeScript gets concerned that the base FeathersApplication type doesn't have those methods.

But trying to do that single API assignment at the top of the file was the reason things were breaking, as Engine.instance hasn't been defined when the file is first loaded.

PR #10979, opened 2024-08-15
Remove side effects from components

MichaelEstes:
## References
closes #_insert number here_

### DIFF
--- a/packages/engine/src/scene/components/SourceComponent.ts
+++ b/packages/engine/src/scene/components/SourceComponent.ts
@@ -23,9 +23,11 @@ All portions of the code written by the Ethereal Engine team are Copyright © 20
 Ethereal Engine. All Rights Reserved.
 */
 
-import { defineComponent } from '@etherealengine/ecs/src/ComponentFunctions'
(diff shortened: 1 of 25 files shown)


packages/engine/src/scene/components/SourceComponent.ts
```diff
@@ -49,15 +51,23 @@ export const SourceComponent = defineComponent({
...
+    const sourceComponent = useComponent(entity, SourceComponent)
 
-    const entities = SourceComponent.entitiesBySource[src].filter((currentEntity) => currentEntity !== entity)
-    if (entities.length === 0) {
-      SourceComponent.entitiesBySourceState[src].set(none)
-    } else {
-      SourceComponent.entitiesBySourceState[src].set(entities)
-    }
+    useLayoutEffect(() => {
+      const source = sourceComponent.value
```
HexaField, line 49: if the source changes after the component has been created, this will no longer have the same value. we should just move this line inside the returned function

MichaelEstes: Oh if the source can change this is setup incorrectly altogether and won't clean up properly, good catch

PR #10980, opened 2024-08-15
Updated scene name regex

barankyle:
## Summary

Underscores are not allowed at all, and hyphens can't be the end character.

Updated invalidSceneName text to reflect this.

Resolves IR-3952

## References
closes #_insert number here_


aditya-mitra (approved):
LGTM

